PR #115, opened 2026-03-12
feat: Add on_tool_not_found event (#107)

### LINKED ISSUE #107: Add on_tool_not_found event
**Priority:** Medium

**Summary:** Add event that fires when LLM requests a tool that doesn't exist, enabling better error messages and tool suggestions.

**Motivation:** Currently just returns "Tool 'x' not found". Could provide fuzzy matching suggestions, dynamic tool loading, better error messages.

**API:**
```python
@on_tool_not_found
def suggest_tool(agent) -> str:
    pending = agent.current_session['pending_tool']
    tool_name = pending['name']
    
    similar = fuzzy_match(tool_name, agent.tools._tools.keys())
    if similar:
        return f"Tool '{tool_name}' not found. Did you mean: {', '.join(similar)}?"
    
    return f"Tool '{tool_name}' not found. Available tools: {list(agent.tools._tools.keys())}"
```

**Implementation:** tool_executor.py line 136 in execute_single_tool
- Store pending tool info in session
- Fire event (can return custom error message)
- Requires _invoke_events_with_return() method

**Use Cases:** Fuzzy tool name matching, dynamic tool loading, better error messages, tool usage analytics

esreekarreddy:
## Summary

Add event that fires when the LLM requests a tool that doesn't exist, enabling better error messages and tool suggestions.

Closes #107

## Motivation

Currently when the LLM calls a non-existent tool, the agent returns a generic `"Tool 'x' not found"` message with no additional context. Plugins have no way to intercept this to provide fuzzy matching, dynamic tool loading, or better diagnostics.

## Changes

- **`connectonion/core/events.py`** — New `on_tool_not_found()` wrapper function (supports decorator + multi-arg syntax)
- **`connectonion/core/agent.py`** — Register `on_tool_not_found` in event registry; add `_invoke_events_with_return()` method that collects handler return values
- **`connectonion/core/tool_executor.py`** — Store `pending_tool` in session before event fires; invoke `on_tool_not_found` via `_invoke_events_with_return()`; use handler's return value as error message if provided
- **`connectonion/core/__init__.py`** — Export `on_tool_not_found`
- **`connectonion/__init__.py`** — Export `on_tool_not_found`
- **`CLAUDE.md`** — Updated events list
- **`docs/concepts/events.md`** — Added table row, lifecycle diagram entry, and usage example
- **`tests/unit/test_tool_not_found.py`** — 15 new tests
- **`tests/unit/test_events.py`** — Added `on_tool_not_found` to wrapper type assertion
- **`tests/unit/test_tool_executor.py`** — Added `_invoke_events_with_return` stub to FakeAgent
- **`tests/unit/test_tool_executor_errors.py`** — Added `_invoke_events_with_return` stub to mock agents

## API

```python
from connectonion import Agent, on_tool_not_found

@on_tool_not_found
def suggest_tool(agent) -> str:
    pending = agent.current_session['pending_tool']
    tool_name = pending['name']
    available = list(agent.tools._tools.keys())

    similar = [t for t in available if t.startswith(tool_name[:3])]
    if similar:
        return f"Tool '{tool_name}' not found. Did you mean: {', '.join(similar)}?"
    return f"Tool '{tool_name}' not found. Available: {available}"

agent = Agent('my_agent', tools=[...], on_events=[suggest_tool])
```

Handler returns a string to override the default error message. Return `None` (or nothing) to keep the default `"Tool 'x' not found"` behavior.

## Tests

All 83 relevant tests pass (15 new + 68 existing). Zero regressions.

```
tests/unit/test_tool_not_found.py: 15 passed
tests/unit/test_events.py: 39 passed
tests/unit/test_tool_executor.py: 10 passed
tests/unit/test_tool_executor_errors.py: 19 passed
```

Edge cases tested:
- Fires when tool is missing, does not fire for existing tools
- Handler return value overrides default message
- Handler returning None uses default message
- First non-None return wins across multiple handlers
- `pending_tool` set in session before handler fires, cleared after
- Trace status is `not_found`
- Decorator and wrapper syntax both work
- Multi-arg wrapper returns list
- Real Agent registers event correctly
- `_invoke_events_with_return` returns first non-None result

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,7 +16,7 @@ ConnectOnion is a Python framework for creating AI agents with automatic activit
 - **Tool Factory** (`connectonion/tool_factory.py`): Converts Python functions to OpenAI-compatible tool schemas automatically
 - **Logger** (`connectonion/logger.py`): Unified logging facade (terminal + plain text + YAML sessions) with `quiet` and `log` parameters
 - **Console** (`connectonion/console.py`): Low-level terminal output with Rich formatting (used internally by Logger)
-- **Events** (`connectonion/events.py:11`): Lifecycle hooks (after_user_input, before_llm, after_llm, before_each_tool, before_tools, after_each_tool, after_tools, on_error, on_complete, on_stop_signal)
+- **Events** (`connectonion/events.py:11`): Lifecycle hooks (after_user_input, before_llm, after_llm, before_each_tool, before_tools, after_each_tool, after_tools, on_error, on_complete, on_stop_signal, on_tool_not_found)
 - **Trust System** (`connectonion/network/trust/`): Three-level verification (open/careful/strict) with custom policy support
 - **XRay Debug** (`connectonion/xray.py:32`): Runtime context injection for interactive debugging with `@xray` decorator
 
--- a/connectonion/__init__.py
+++ b/connectonion/__init__.py
@@ -45,6 +45,7 @@ from .core import (
     on_error,
     on_complete,
     on_stop_signal,
+    on_tool_not_found,
 )
 from .logger import Logger
 from .llm_do import llm_do
@@ -132,4 +133,5 @@ __all__ = [
     "on_error",
     "on_complete",
     "on_stop_signal",
+    "on_tool_not_found",
 ]
--- a/connectonion/core/__init__.py
+++ b/connectonion/core/__init__.py
@@ -34,6 +34,7 @@ from .events import (
     on_error,
     on_complete,
     on_stop_signal,
+    on_tool_not_found,
 )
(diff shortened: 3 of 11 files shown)


wu-changxing:
Closing for now — thank you for the contribution.

This has been open since April and now conflicts with main, which has moved a long way in the event and tool-execution layer since. Rather than rebase it across that gap, we would rather revisit the design against current code when the feature comes up the roadmap.

#107 stays open and carries the idea, so this is a deferral rather than a rejection. The event system has changed shape since this was written, so whenever it is picked up it is worth re-deriving the hook point rather than porting this branch — through no fault of the original work.

PR #144, opened 2026-05-31
fix: strip <system-reminder> tags from chat messages

### LINKED ISSUE #138: session_to_chat_items leaks <system-reminder> tags into user-visible chat bubbles
## Problem

`<system-reminder>...</system-reminder>` blocks are injected into `current_session['messages']` by several plugins to give the LLM extra context (build hints, tool reminders, file-upload notices, approval-rejection follow-ups). These tags are intended for the LLM's eyes only — but `session_to_chat_items` (in `connectonion/network/host/session/ui.py`) walks `messages` and emits `user` / `agent` chat items verbatim. Result: the raw tag dump appears as a user bubble (or appended to a tool result) on the frontend.

## Where the tags get injected

Found 5 injection sites currently in tree:

| File | Line | Role injected as | Trigger |
|---|---|---|---|
| `cli/co_ai/plugins/system_reminder.py:228-231` | 230 | `user` | `@after_user_input` for build tasks |
| `core/agent.py:293` | 293 | `user` (appended to prompt) | when user uploads files |
| `useful_plugins/prefer_write_tool.py:106,134` | 106 / 134 | `tool` | bash tool used for file create |
| `useful_plugins/tool_approval/approval.py:595` | 595 | `tool` | reject_explain mode |
| `cli/co_ai/templates`, `useful_plugins/builtin_agents/explore/AGENT.md` | — | system prompts | init |

The most visible bug: the \`role: user\` injections show up as **fake user bubbles** in oo-chat with raw \`<system-reminder>\` text, which is confusing/leaky.

## Reproduce

1. Use co_ai agent with a build prompt that triggers \`detect_intent → is_build=True\`.
2. Watch oo-chat: a user bubble appears containing \`<system-reminder>\\n<build-context>\\n</system-reminder>\` immediately after the real prompt.
3. Same with file uploads (core/agent.py:293) and tool rejections (approval.py:595).

## Proposed fix

Strip \`<system-reminder>...</system-reminder>\` blocks (including any surrounding whitespace) when converting messages → chat_items. Two options:

**A. Strip in \`session_to_chat_items\`** (single chokepoint, all injection sites covered)
\`\`\`python
import re
SYSTEM_REMINDER_RE = re.compile(r'\\s*<system-reminder>.*?</system-reminder>\\s*', re.DOTALL)

def _strip_system_reminders(content: str) -> str:
    return SYSTEM_REMINDER_RE.sub('', content).strip() if isinstance(content, str) else content
\`\`\`
Apply to user/agent message content, and inside tool_result entries.

**B. Strip in chat-item-mapper on the client** (mirrors INTERJECTION_FRAME_PREFIX strip pattern)
Move the responsibility to the SDK / oo-chat. Cleaner separation but means every client implementation must do it.

Recommendation: **A**. The server already strips \`INTERJECTION_FRAME_PREFIX\` in this same function, so the precedent is set. Centralized + works for any client.

## Edge cases

- A user actually typing \`<system-reminder>foo</system-reminder>\` — extremely unlikely but harmless if stripped (they shouldn't be sending markup like this).
- Empty messages after stripping (entire message was just a reminder block) — should suppress the user item entirely, not emit an empty bubble.
- Nested or unclosed tags — \`re.DOTALL\` non-greedy handles common cases; malformed blocks left untouched (LLM bug, not UI bug).

## Tests to add

- \`test_host_session.py::test_strip_system_reminder_from_user_msg\` — message containing only reminder → no chat item emitted.
- \`test_host_session.py::test_strip_system_reminder_inline\` — message containing real text + reminder block → only real text in chat_item.
- \`test_host_session.py::test_strip_does_not_affect_assistant_msg\` — agent's response with text that happens to mention "system-reminder" stays intact.

## Impact

User-visible bug. Cosmetic but confusing — looks like the agent is talking to itself or that build context is being entered by the user.

## References

- \`connectonion/network/host/session/ui.py:session_to_chat_items\`
- Existing precedent: \`RUNTIME_INPUT_FRAME_PREFIX\` strip in same function
- Related: PR #134 (network refactor) which fixed the runtime_input prefix strip

Oxygen56:
## Summary
- Added _strip_system_reminders() regex function to remove <system-reminder>...</system-reminder> blocks
- Applied in session_to_chat_items() on user messages, assistant messages, and tool_result entries
- Empty messages after stripping are suppressed entirely (no empty bubble)
- Follows same pattern as RUNTIME_INPUT_FRAME_PREFIX stripping
- Added 4 test cases covering: pure-reminder suppression, inline reminder stripping, assistant text preservation, and tool_result stripping

Fixes #138

## Test Plan
- [x] pytest tests/ -x -q passes

### DIFF
--- a/connectonion/network/host/session/ui.py
+++ b/connectonion/network/host/session/ui.py
@@ -20,7 +20,27 @@ Streaming-only event types (llm_call, llm_result, compact, etc.) are not
 reconstructed — they're live-only feedback and don't survive reconnect by design.
 """
 
+import re
+
 from ....useful_plugins.runtime_input import RUNTIME_INPUT_FRAME_PREFIX
+
+# Regex to strip <system-reminder>...</system-reminder> blocks (internal Claude
+# API markers) that plugins inject into messages. These are intended only for
+# the LLM and must not appear in user-visible chat UI. Non-greedy match with
+# surrounding whitespace to avoid leaving bare newlines.
+_SYSTEM_REMINDER_RE = re.compile(r'\s*<system-reminder>.*?</system-reminder>\s*', re.DOTALL)
+
+
+def _strip_system_reminders(content: str) -> str:
+    """Remove <system-reminder>...</system-reminder> blocks from a string.
+
+    Returns the cleaned string with surrounding whitespace normalised.
+    Non-string values (e.g. list content blocks) are returned unchanged.
+    """
+    if not isinstance(content, str):
+        return content
+    stripped = _SYSTEM_REMINDER_RE.sub('', content)
+    return stripped.strip()
 
 
 def _trace_entry_to_item_ui(entry: dict, idx: int) -> dict | None:
@@ -84,7 +104,7 @@ def _trace_entry_to_item_ui(entry: dict, idx: int) -> dict | None:
             'name': name,
             'args': entry.get('args'),
             'status': ui_status,
-            'result': entry.get('result'),
+            'result': _strip_system_reminders(entry.get('result', '')),
             'timing_ms': entry.get('timing_ms'),
         }
 
@@ -124,6 +144,11 @@ def session_to_chat_items(session: dict) -> list[dict]:
             content = msg.get('content', '')
             if isinstance(content, str) and content.startswith(RUNTIME_INPUT_FRAME_PREFIX):
                 content = content[len(RUNTIME_INPUT_FRAME_PREFIX):]
+            content = _strip_system_reminders(content)
+            if not content:
+                # Entire message was just a <system-reminder> block — suppress the bubble.
+                user_count += 1
+                continue
             items_ui.append({'id': f"msg-{msg_idx}", 'type': 'user', 'content': content})
             user_count += 1
             if user_count < len(turn_entries):
@@ -132,7 +157,9 @@ def session_to_chat_items(session: dict) -> list[dict]:
                     if item_ui:
                         items_ui.append(item_ui)
         elif role == 'assistant' and msg.get('content'):
-            items_ui.append({'id': f"msg-{msg_idx}", 'type': 'agent', 'content': msg.get('content', '')})
+            content = _strip_system_reminders(msg.get('content', ''))
+            if content:
+                items_ui.append({'id': f"msg-{msg_idx}", 'type': 'agent', 'content': content})
 
     # Fallback: if no user_input markers found in trace, append all remaining trace
     # entries at the end so the data isn't silently dropped (older sessions).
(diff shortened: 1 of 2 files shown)


sanmaxdev (commented):
I checked the replay path and the empty-marker case still advances turn indexing, so trace items stay paired with the right user turn.

Verification: `python -m pytest tests/unit/test_host_session.py -q`

wu-changxing:
Reviewed and pushed a fix to this branch (478f687) — thanks for the contribution, the stripping itself is right and well-tested.

**The bug**

When a user message was *entirely* a `<system-reminder>` block, the new code did:

```python
if not content:
    user_count += 1
    continue
```

That `continue` also skipped the block right below it that renders `turn_entries[user_count]` — so **every tool call in that turn disappeared from the transcript.** Reproduced on the branch:

```
session: user=<system-reminder>injected</system-reminder>, assistant=done
        trace=[user_input, tool_result(bash)]
before fix → [('agent', 'done')]        # the bash tool_call is gone
after fix  → [('tool_call', 'bash'), ('agent', 'done')]
```

That trades a stray reminder bubble for silently deleted tool calls, which is the worse of the two.

**The fix**: emit the bubble conditionally, render the trace unconditionally — the message still opened a turn even when it has nothing to show.

Added `test_suppressed_user_bubble_still_renders_its_tool_calls`, which fails on the previous commit and passes now. `tests/unit/test_host_session.py` is 66 passed; the full unit suite is green against current main.

One note: CI has never run on this branch (fork PRs don't get repo secrets, and there's a pending fix for that in #184). I verified locally against current main instead.

wu-changxing:
Closing this one — not because it is wrong, but because we want to fix it upstream instead.

To be clear about what I verified, since the work was sound: the bug is real and I confirmed all five injection sites named in #138 exist in tree. Against those real shapes this PR behaves correctly — a reminder appended to a user prompt, a message that is nothing but a reminder, a tool result that is entirely a reminder, and a reminder appended to a tool result. I also found and fixed a bug in it along the way (the reminder-only branch was dropping that turn's tool calls from the transcript, which was worse than the leak it replaced). The unconditional `.strip()` turned out to be harmless — code-block indentation survives.

**Why close it anyway:** stripping at render time is a scrub on the way out, not a fix. The root cause is that `core/agent.py:293` concatenates the reminder into the *user's own prompt string*, and `cli/co_ai/plugins/system_reminder.py:230` appends a `role: user` message that is nothing but a reminder. By the time `session_to_chat_items` runs, system-injected text and user-typed text are indistinguishable — which is exactly why this PR silently rewrites a message from a user who happens to type the literal tag. The information needed to do it correctly was destroyed upstream.

Keeping the regex would also hide the fact that the message list is being polluted, so upstream never gets fixed.

Filed the root-cause issue and linked it from #138. Thanks for the work — the verification of the five sites and the reminder-only edge case both carry over to that fix.

PR #151, opened 2026-06-09
host: publish agent profile to the relay on first ANNOUNCE

yyy900:
## Problem
A `host()` agent connects to the relay and is reachable, but the online directory has nothing to show for it — no name, tools, or skills.

## Change
`host()` builds a publishable **profile** from agent metadata and sends it with the **first ANNOUNCE of each connection**:

```python
{alias, tools, model, skills: [{name, description}]}
```

The relay persists it through the existing durable profile path (oo-api), so heartbeats don't carry it — `create_announce_message` has supported `profile=` since March, so the wire format needs no changes and this diff is +68/−3.

## Privacy
Only **project-level** skills (`<project>/.co/skills/`) are published. Skill discovery also picks up `~/.co/skills` — the operator's personal toolbox; live testing showed a real machine would have published 62 skills (37KB), exposing which platforms the operator automates. Skill filesystem locations and the prompt summary stay private for the same reason.

## Why durable profile, not in-memory metadata
An earlier revision of this branch published a runtime `metadata` payload refreshed on every heartbeat and held in the relay's memory. Skills are publishable content — large, soon with SKILL.md bodies — and belong in the durable channel, sent once per connection. The in-memory registry keeps doing exactly its job: liveness + endpoints.

## Pairing
oo-api #16 (relay serves the profile with agent records) · oo-chat #14 / connectonion-ts #5 (read it).

## Tests
`test_host_relay.py`: project-only skill filtering (user-level skills, locations, and summary excluded), profile passed to the relay lifespan. Full unit suite: 1925 passed.

### DIFF
--- a/connectonion/network/host/server.py
+++ b/connectonion/network/host/server.py
@@ -119,24 +119,23 @@ def _extract_agent_metadata(create_agent: Callable) -> tuple[dict, object]:
     return metadata, sample
 
 
-def _build_agent_profile(agent_metadata: dict, project_dir: Path) -> dict:
+def _build_agent_profile(agent_metadata: dict) -> dict:
     """Build the publishable profile sent with the first ANNOUNCE of a connection.
 
-    Only project-level skills are published: skill discovery also picks up
-    ~/.co/skills (the operator's personal toolbox), which must not leak into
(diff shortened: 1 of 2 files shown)


yyy900:
Consolidated into #154.

PR #152, opened 2026-06-11
Dispatch browser calls to one thread; bound forwarder waits; persist screenshots into replay

yyy900:
Stacked on #151.

Three fixes from debugging the hosted LinkedIn agent:

1. **`browser: dispatch all public methods to one dedicated thread`** — Playwright objects are not thread-safe; calls coming in from different executor threads corrupted state. All public browser methods now dispatch to a single dedicated thread.
2. **`io: bound forwarder wait so idle sessions don't pin executor threads`** — idle sessions were holding executor threads forever in the forwarder wait.
3. **`session: persist screenshot images into replayed chat_items`** — screenshots survived only in the live stream; replays after reconnect lost them.

### DIFF
--- a/connectonion/network/host/session/ui.py
+++ b/connectonion/network/host/session/ui.py
@@ -91,6 +91,18 @@ def _trace_entry_to_item_ui(entry: dict, idx: int) -> dict | None:
     return None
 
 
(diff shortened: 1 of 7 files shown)


yyy900:
Consolidated into #154.

PR #153, opened 2026-06-11
deploy: named template deploys + per-skill --skills paths + nodejs in co-ai image

yyy900:
Stacked on #152.

Makes `co deploy --template co-ai --name <agent> --skills <skill>...` a full replacement for hand-built agent projects (the LinkedIn agent path):

1. **`--name`** — template deploys were hardcoded to `{template}-agent`, so two different skill combinations could not coexist; they overwrote each other's deployment. `--name` sets the project name (and URL); default unchanged.
2. **Per-skill `--skills` paths** — a `--skills` path that is itself a skill (contains `SKILL.md`) now lands at `.co/skills/<dirname>/` instead of being flattened into `.co/skills/` (where multiple skills' `SKILL.md` collided). Directory-of-skills paths keep the existing behavior, so both granularities work.
3. **nodejs in the co-ai image** — skills ship node scripts (`linkedin-article-submit` runs `parse-markdown.js` via bash). This was the only difference between the hand-written LinkedIn Dockerfile and the template one.

Verified end to end: `co deploy --template co-ai --name linkedin-skills --skills .../linkedin-login --skills .../linkedin-post-writer --skills .../linkedin-post-submit` deploys a working hosted agent with exactly those three skills under `.co/skills/`.

Tests: 22/22 in `tests/e2e/cli/test_cli_deploy.py` (two new: single-skill nesting, `--name` without `--template` errors).

### DIFF
--- a/connectonion/cli/commands/deploy_commands.py
+++ b/connectonion/cli/commands/deploy_commands.py
@@ -163,22 +163,26 @@ def _build_tarball(project_dir: Path, skills_paths: list[Path]) -> Path:
                     continue
                 tar.add(path, arcname=str(rel), recursive=False)
         for skills_path in skills_paths:
+            # A path is either one skill (has SKILL.md) or a directory of skills.
+            arc_prefix = Path(".co") / "skills"
+            if (skills_path / "SKILL.md").exists():
+                arc_prefix = arc_prefix / skills_path.name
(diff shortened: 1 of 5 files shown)


yyy900:
Consolidated into #154.

PR #156, opened 2026-06-18
fix: require word boundary for Bash(cmd *) permission wildcard

failsafesecurity:
## Problem
`matches_permission_pattern()` matches the `Bash(cmd *)` wildcard with `actual_cmd.startswith(prefix)` and no word boundary, so a permission scoped to one command family also authorizes unrelated commands sharing the prefix:

```
Bash(git *) -> matches 'gitleaks --dump-secrets'   (should not)
Bash(ls *)  -> matches 'lsof -i'                     (should not)
Bash(cat *) -> matches 'catnip /etc/passwd'          (should not)
```

This widens the auto-approved command set beyond operator intent (CWE-863). Severity is constrained by needing such a wildcard in config plus a useful prefix-sharing binary, so this is a hardening fix.

## Fix
Require an exact match or a trailing space after the prefix:
```python
if actual_cmd == prefix or actual_cmd.startswith(prefix + ' '):
    return True
```

## Test Plan
- [ ] Existing test suite passes (`pytest tests/unit/test_tool_approval.py`, Python >=3.10)
- [ ] `Bash(git *)` matches `git` and `git status`, but NOT `gitleaks --dump`
- [ ] `Bash(ls *)` does NOT match `lsof -i`

## Security Note
Low/hardening, CWE-863 (Incorrect Authorization). Reported by FailSafe Researcher (authorized security review).

### DIFF
--- a/connectonion/useful_plugins/tool_approval/approval.py
+++ b/connectonion/useful_plugins/tool_approval/approval.py
@@ -373,10 +373,12 @@ def matches_permission_pattern(tool_name: str, tool_args: dict, pattern: str) ->
         if cmd_pattern == actual_cmd:
             return True
 
-        # Wildcard match: "git diff *" matches "git diff --staged"
+        # Wildcard match: "git diff *" matches "git diff --staged" (and bare
+        # "git diff"), but NOT "git difftool". Require a word boundary after the
+        # prefix so "git *" cannot match an unrelated binary like "gitleaks".
         if cmd_pattern.endswith(' *'):
             prefix = cmd_pattern[:-2]  # Remove " *"
-            if actual_cmd.startswith(prefix):
+            if actual_cmd == prefix or actual_cmd.startswith(prefix + ' '):
                 return True
 
         # Wildcard match: "git *" matches "git status"
--- a/tests/unit/test_skills.py
+++ b/tests/unit/test_skills.py
@@ -186,6 +186,12 @@ class TestPatternMatching:
 
     def test_bash_command_prefix_wildcard(self):
         """Test bash command prefix wildcard (git *)."""
+        assert matches_permission_pattern(
+            'bash',
+            {'command': 'git'},
+            'Bash(git *)'
+        ) == True
+
         # "git *" should match any git command
         assert matches_permission_pattern(
             'bash',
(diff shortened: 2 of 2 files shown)


failsafesecurity:
Hi maintainers 👋 — friendly follow-up on this security hardening PR. It has been open for a while with no reviewer feedback yet. Happy to address any concerns or adjust the approach. Thanks!

failsafesecurity:
Thanks for reviewing this security fix. Just a friendly ping — this PR has been open for a while. Happy to adjust the approach if you have any concerns. No rush, just want to make sure it stays on your radar.

sanmaxdev (commented):
The behavior change is right, but this should add a regression test next to `test_bash_command_prefix_wildcard`. That would lock the allow case for `git status` and the reject case for `gitleaks detect`.

connectonion/useful_plugins/tool_approval/approval.py
```diff
@@ -373,10 +373,12 @@ def matches_permission_pattern(tool_name: str, tool_args: dict, pattern: str) ->
...
             return True
 
-        # Wildcard match: "git diff *" matches "git diff --staged"
+        # Wildcard match: "git diff *" matches "git diff --staged" (and bare
+        # "git diff"), but NOT "git difftool". Require a word boundary after the
+        # prefix so "git *" cannot match an unrelated binary like "gitleaks".
         if cmd_pattern.endswith(' *'):
             prefix = cmd_pattern[:-2]  # Remove " *"
-            if actual_cmd.startswith(prefix):
+            if actual_cmd == prefix or actual_cmd.startswith(prefix + ' '):
```
sanmaxdev, line 381: This boundary change looks right. One thing I would add before merging is a small regression beside `test_wildcard_git_diff_star` for `Bash(git *)`: accept `git status` and bare `git`, and reject `gitleaks detect` plus `git-diff`.

I verified the current implementation with a direct `matches_permission_pattern` smoke check, but test coverage would keep this path from slipping later.

wu-changxing:
Verified — the word-boundary fix is correct: `Bash(ls *)` no longer matches `lsof`, `Bash(git *)` no longer matches `gitleaks`, and the bare / `git status` / prefixed branches all still behave.

Before merge, **please add a regression test** in `tests/unit/test_skills.py` (which already exercises `matches_permission_pattern`) — e.g. assert `git *` matches `git` and `git status` but not `gitleaks`. With that assertion in place this is ready to merge.

wu-changxing:
Maintainer follow-up: the word-boundary fix is correct; regression coverage is the remaining blocker before merge.

Please add focused assertions alongside the existing matches_permission_pattern tests covering:

1. Bash(git *) matches the bare git command;
2. Bash(git *) matches git status and git diff --staged;
3. Bash(git *) does not match gitleaks detect;
4. Bash(ls *) does not match lsof -i;
5. a multi-word prefix such as Bash(git diff *) matches git diff and git diff --staged, but not git difftool.

Please run the relevant tool-approval and skills test suites and push the test commit so CI can verify the behavior. With those tests in place, this should be ready for merge.

failsafesecurity:
Added the requested regression coverage and pushed it to `fix/find-002-bash-prefix-boundary` in commit `316126f`.

Covered here:
- `Bash(git *)` matches `git`, `git status`, and `git diff --staged`;
- `Bash(git *)` does not match `gitleaks detect`;
- `Bash(ls *)` does not match `lsof -i`;
- `Bash(git diff *)` matches `git diff` and `git diff --staged`, but not `git difftool`.

Local verification:
- `uv run --python 3.11 pytest tests/unit/test_skills.py -k 'bash_command_prefix_wildcard or single_pattern_string'`
- `uv run --python 3.11 pytest tests/unit/test_config_permissions.py -k 'wildcard_git_diff_star'`

Please take another look when convenient.

PR #154, opened 2026-06-11
Hosted agents: relay profile, browser thread safety, template deploys with skills

yyy900:
Consolidates #151, #152, #153 into one PR (replaces all three).

**Relay profile (display metadata)**
- `host()` publishes a display profile (alias/tools/model + project-level skills only — user/builtin skills stay private) with the first ANNOUNCE of each connection; the relay persists it so heartbeats stay small. Pairs with openonion/oo-api#18 relay-side serving.
- Skill filter fix: `location` is a discovery category (project/user/builtin), not a path.

**Browser thread safety (hosted agents)**
- Playwright's sync API is thread-bound; `host()` runs turns on arbitrary pool threads. All public `BrowserAutomation` methods dispatch to the instance's dedicated worker thread (executor owned by `__init__`, wrapping applied by a class decorator).
- Bounded the idle-session forwarder wait (~1s) so idle sessions can't pin executor-pool threads and starve new connections.
- Screenshot data URLs stay on the trace (`entry['image']`) so session replay re-emits them; replays no longer lose screenshots.

**Onboard completes the gated CONNECT**
- The trust gate interrupts CONNECT before `conn` is populated; a successful ONBOARD_SUBMIT used to leave the connection in a dead end (`authenticate first (send CONNECT)` on the next INPUT, and a CONNECT replay can age past the 5-minute signature window during invite/payment entry). `handle_connect` now stashes the gated CONNECT; after a verified onboard the session loop runs `establish_connection()` with the onboard-verified identity, ending in CONNECTED so the client's pending input resumes. Pairs with the SDK-side retry removal in openonion/connectonion-ts#7.

**Template deploys with arbitrary skill combinations**
- `co deploy --template <t> --name <agent>` names the project (default `{template}-agent`), so different skill combos run side by side instead of overwriting each other.
- `--skills A B C` takes multiple paths after one flag (extra positionals are collected; misspellings fail loudly in path validation); repeated flags still work. A path that is itself a skill (contains `SKILL.md`) nests under `.co/skills/<dirname>/`.
- co-ai image gains `nodejs` (skills ship node scripts — the only difference vs the hand-written LinkedIn agent Dockerfile).
- Verified end to end: template deploy with three `--skills` paths → agent running with exactly those skills; the redeployed `linkedin-agent` runs this branch in production.

Every changed file's LLM-Note header updated. Tests: 2150 passed (unit + e2e/cli).

PR #161, opened 2026-06-26
Stabilize hosted WebSocket and relay connections

yyy900:
## Summary
- bound idle WebSocket forwarder waits so blocked sessions do not pin executor threads
- complete gated CONNECT server-side after onboard succeeds
- keep relay registration alive across transient supervisor failures and forward client PINGs through relay sessions
- document the connection lifecycle behavior

Conflict note: this branch intentionally does not include relay profile publication; ANNOUNCE shape stays on current main.

## Verification
- .venv/bin/python -m pytest tests/unit/test_io.py tests/unit/test_asgi_websocket_admin_onboard.py tests/unit/test_relay.py -q

### DIFF
--- a/connectonion/network/host/server.py
+++ b/connectonion/network/host/server.py
@@ -26,6 +26,7 @@ don't interfere between concurrent requests.
 """
 
 import asyncio
+import random
 from functools import partial
 from pathlib import Path
 from typing import Callable, Union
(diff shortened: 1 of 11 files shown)


wu-changxing:
Closing: conflicts with main beyond practical rebase, and the branch has not moved since June while the hosted WebSocket and relay layer has been reworked around it (#186, #225, #242 among others).

Being explicit about what is *not* being claimed: unlike #154, I could not verify that this branch's behaviour landed on main by another route. So if hosted WebSocket/relay connections are still dropping under the conditions this addressed, that is a live bug and closing this does not fix it — it should be re-filed as an issue against current main, where it can be reproduced and fixed against code that exists today rather than rebased across three months of divergence.

Recording what this claimed to fix so the thread is findable: stabilising hosted WebSocket and relay connections (+391/-26).

Thanks for the work.

PR #162, opened 2026-06-26
Preserve and bound session image replay

yyy900:
## Summary
- persist screenshot data URLs on trace entries so replayed sessions can rebuild image chat items
- keep only a small screenshot window in LLM context while preserving replay images
- sanitize stale or invalid image_url parts before provider calls

## Verification
- .venv/bin/python -m pytest tests/unit/test_host_session.py tests/unit/test_image_result_formatter.py -q

### DIFF
--- a/connectonion/network/host/session/ui.py
+++ b/connectonion/network/host/session/ui.py
@@ -91,6 +91,27 @@ def _trace_entry_to_item_ui(entry: dict, idx: int) -> dict | None:
     return None
 
 
+def _is_replayed_screenshot(message: dict, screenshot_urls: set) -> bool:
+    """True if this user message is a formatter-inserted screenshot — its image is one
+    the image_result_formatter recorded on a trace entry (trace_entry['image']).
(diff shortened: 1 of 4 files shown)


wu-changxing:
Closing: same reason as #161 — conflicts with main beyond practical rebase, untouched since June while the session and WebSocket layer moved on.

Also being explicit here: I checked main for this behaviour and could not find it. `ws_router/session.py` has no screenshot or data-URL handling, so **session replay may still lose images**. If so, closing this does not fix it.

Recording the claim so it stays searchable: preserve and bound session image replay — keep screenshot data URLs on the trace so replays re-emit them, with a bound so they cannot grow without limit (+329/-13). Related: the `image_result_formatter` plugin does the inbound half (turning a screenshot path into an image the model can see), but that is not the same as replaying one.

Worth re-filing as an issue against current main if replayed sessions are still coming back without their screenshots.

Thanks for the work.

PR #172, opened 2026-07-02
image_result_formatter: upload screenshots to oo-api (URL instead of base64)

yyy900:
## What
Managed (`co/`) agents upload screenshot bytes to oo-api's `/api/v1/images` (content-addressed GCS) and reference the image by a stable `/img` URL in the message, instead of embedding the base64 data URL.

## Why
The base64 payload sits in `current_session["messages"]`, which is replayed to the LLM every turn and persisted in session logs / stateless continuation. A multi-step browser agent (many screenshots) blows up context size, memory, and transfer. Uploading the bytes to GCS keeps history at a ~70-byte URL. Non-managed agents (no oo-api) keep the inline data URL, so their behavior is unchanged.

## Pairs with
oo-api images routes (`POST /api/v1/images`, `GET /img/{sha}`) + per-provider materialization (`forward_to_gemini` inlines the URL for Gemini, which can't fetch remote URLs). The oo-api side is already deployed.

## Notes
- Upload-only; the 3-screenshot sliding window (hosted-agents) is a separate concern, not included here.
- Fails loudly on upload error (managed agents already depend on oo-api for LLM calls).

## Test
Verified end-to-end locally: `co/gemini` agent → screenshot → upload → GCS → oo-api inline → real Gemini correctly read the image; the session held the `/img` URL, not base64.

## Follow-up (77baaef, review by Aaron)
- Hoisted the managed/non-managed branch into `_format_image_result` — the two paths are visible at the call site; helper renamed `_upload_to_oo_api` (it only uploads now; the old `_upload_image` name lied on the non-managed path)
- Removed defensive `getattr(agent, "llm", None)`; the test `FakeAgent` now has an `llm` like real Agents
- Added a unit test for the managed upload path (endpoint, Bearer auth, URL in message + io, no base64 left in history)
- Updated `docs/useful_plugins/image_result_formatter.md`: How It Works covers the oo-api upload; `send_image` documented as URL-or-data-URL

### DIFF
--- a/connectonion/useful_plugins/image_result_formatter.py
+++ b/connectonion/useful_plugins/image_result_formatter.py
@@ -3,14 +3,14 @@ LLM-Note: Image result formatter plugin.
 
 Purpose: Remove raw base64 image payloads from the tool message, add a
 multimodal image message so the LLM can see the image, and send image results to
-frontend IO when available. For managed (co/) agents the image bytes are uploaded
-to oo-api (content-addressed GCS) and referenced by URL, so screenshots never bloat
-the replayed message history; non-managed agents keep the inline data URL.
+frontend IO when available. Image bytes are uploaded to oo-api (content-addressed
+GCS) and referenced by URL, so screenshots never bloat the replayed message
+history. Requires OPENONION_API_KEY (set up by `co init`).
 
 Data flow: after_tools event -> scan successful tool_result trace entries ->
-detect image data URL or raw base64 -> upload to oo-api (managed) or keep data URL
--> replace the matching tool message with a short placeholder -> insert a user
-message containing image_url -> shorten the trace result.
+detect image data URL or raw base64 -> upload to oo-api -> replace the matching
+tool message with a short placeholder -> insert a user message containing
+image_url -> shorten the trace result.
 
 State/Effects: mutates agent.current_session["messages"] and trace entries;
 optionally calls agent.io.send_image(image_url). Non-image tool results are ignored.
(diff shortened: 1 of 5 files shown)


wu-changxing:
Pushed a follow-up commit (77baaef) addressing review feedback:

- **Naming**: `_upload_image` didn't upload on the non-managed path. The managed/non-managed branch is now hoisted into `_format_image_result` (visible at a glance), and the helper is renamed `_upload_to_oo_api` — it only uploads now.
- **No defensive getattr**: real Agents always have `.llm`; the test FakeAgent now has one instead.
- **Test gap**: added a unit test for the managed upload path (endpoint, Bearer auth, URL lands in message + io, no base64 left in history). All 1960 unit tests pass.
- **Docs**: `docs/useful_plugins/image_result_formatter.md` updated — How It Works covers the oo-api upload, and send_image is documented as URL-or-data-URL so frontends handle both.

Remaining open question for a separate issue: `GET /img/{sha}` is publicly readable (bearer-capability URL). Screenshots can contain secrets/PII and the URL flows into session logs and third-party providers — worth deciding on TTL or token-scoped access on the oo-api side.

wu-changxing:
Second follow-up (8da94c7): dropped the per-LLM branch entirely — every image now uploads to oo-api via `OPENONION_API_URL` + `OPENONION_API_KEY` from the environment, same pattern the gmail/calendar tools already use. No model sniffing, one code path. Also caught that the read_file image test was hitting the real network after this change — now stubbed. Full unit suite: 1960 passed.

wu-changxing (approved):

PR #184, opened 2026-07-06
feat: support async tool functions in tool executor

### LINKED ISSUE #125: Support async tool functions
## Problem

Tool executor (`tool_executor.py:195`) only supports sync functions:

```python
result = tool_func(**tool_args)
```

This blocks us from using async libraries as tools (browser-use CDP, async HTTP clients, async database drivers, etc). Currently the only workaround is wrapping every async call with `asyncio.run()` inside each tool method — ugly and error-prone.

## Proposed Change

One line change in `execute_single_tool()`:

```python
import asyncio, inspect

if inspect.iscoroutinefunction(tool_func):
    result = asyncio.run(tool_func(**tool_args))
else:
    result = tool_func(**tool_args)
```

No changes needed in agent.py, tool_factory.py, or tool_registry.py. The agent loop stays sync — we just bridge to async at the execution point.

## Why

Immediate use case: wrap [browser-use](https://github.com/browser-use/browser-use) BrowserSession as a ConnectOnion tool (`useful_tools/browser_use.py`). browser-use is fully async CDP-based. With async tool support, we can write clean tools like:

```python
class BrowserUse:
    async def navigate(self, url: str) -> str:
        event = self.session.event_bus.dispatch(NavigateToUrlEvent(url=url))
        await event
        return f"Navigated to {url}"
```

Instead of wrapping every method with `asyncio.run()`.

## Scope

- `connectonion/core/tool_executor.py` — detect coroutine, run with `asyncio.run()`
- `connectonion/core/tool_factory.py` — no change needed (signature inspection works the same for async functions)
- Tests: add one test with an async tool function

sanmaxdev:
## Summary
- Detect async tool functions with `inspect.iscoroutinefunction` and run them via `_run_async_tool`, which uses `asyncio.run` when no loop is running and falls back to a separate thread with its own loop when a loop is already active (avoids `RuntimeError: This event loop is already running`)
- Sync tools are unaffected: the existing `tool_func(**tool_args)` path runs only for non-coroutine functions
- Addresses the concern raised in the issue comments about `asyncio.run` failing inside an existing event loop

## Testing
- [x] `pytest tests/unit/test_tool_executor.py -v -m "not real_api and not network"` (13 passed)
- [x] `pytest tests/unit/test_tool_executor_errors.py tests/unit/test_tool_factory.py tests/unit/test_tool_registry.py -v -m "not real_api and not network"` (75 passed)

Closes #125

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,6 +34,8 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
           GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
-          OPENONION_API_KEY: ${{ secrets.OPENONION_API_KEY }}
+          # Fork PRs do not receive repository secrets. The offline test suite
+          # still constructs managed-model Agents, so give it a harmless key.
+          OPENONION_API_KEY: ${{ secrets.OPENONION_API_KEY || 'test-key' }}
         run: |
           pytest tests/ -v --tb=short -m "not slow and not real_api and not network"
--- a/connectonion/core/tool_executor.py
+++ b/connectonion/core/tool_executor.py
@@ -13,6 +13,7 @@ import time
 import json
 import asyncio
 import inspect
+import threading
 from typing import List, Dict, Any, Optional, Callable
 
 from ..debug.xray import (
@@ -22,25 +23,37 @@ from ..debug.xray import (
 )
 
 
+_async_loop: Optional[asyncio.AbstractEventLoop] = None
+_async_loop_lock = threading.Lock()
+
+
+def _get_async_tool_loop():
+    """Return the long-lived event loop used by asynchronous tools."""
+    global _async_loop
+
+    with _async_loop_lock:
+        if _async_loop is None or _async_loop.is_closed():
+            ready = threading.Event()
+
+            def run_loop():
+                global _async_loop
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
+                _async_loop = loop
+                ready.set()
+                loop.run_forever()
+
+            threading.Thread(target=run_loop, name="connectonion-async-tools", daemon=True).start()
+            ready.wait()
+
+    assert _async_loop is not None
+    return _async_loop
+
+
 def _run_async_tool(coro):
-    """Run a coroutine to completion, handling nested event loops.
-
-    When no event loop is running in the current thread, ``asyncio.run``
-    creates a fresh loop and drives the coroutine. When a loop is already
-    running (e.g. the agent is driven from an async context), ``asyncio.run``
-    raises ``RuntimeError``; in that case we run the coroutine in a separate
-    thread with its own loop so the calling loop is not disturbed.
-    """
-    try:
-        return asyncio.run(coro)
-    except RuntimeError:
-        # A loop is already running in this thread. Run the coroutine
-        # in a separate thread with its own event loop.
-        import concurrent.futures
-
-        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
-            future = pool.submit(asyncio.run, coro)
-            return future.result()
+    """Run a coroutine on the shared asynchronous-tool event loop."""
+    future = asyncio.run_coroutine_threadsafe(coro, _get_async_tool_loop())
+    return future.result()
 
 
 def execute_and_record_tools(
--- a/connectonion/core/tool_executor.py
+++ b/connectonion/core/tool_executor.py
@@ -11,6 +11,8 @@ LLM-Note:
 
 import time
 import json
+import asyncio
+import inspect
 from typing import List, Dict, Any, Optional, Callable
 
 from ..debug.xray import (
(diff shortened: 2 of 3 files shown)


wu-changxing:
are you a fu*king AI?

sanmaxdev:
> are you a fu*king AI?

no

wu-changxing:
Thanks for working on async tool support. The motivation is valid, but the current execution mechanism has a blocking correctness issue and is not ready to merge yet.

I reproduced this case against the PR head: when an async tool itself raises RuntimeError("original tool failure"), _run_async_tool catches that as though asyncio.run failed because of an already-running loop. It then submits the same already-consumed coroutine again, and the caller receives "cannot reuse already awaited coroutine" instead of the original tool error. This both hides the real failure and retries at the wrong abstraction boundary.

There is also the loop-affinity problem already discussed on #125: creating and destroying a loop per tool call breaks persistent async resources such as BrowserSession, async HTTP clients, and database connections that must remain on the loop where they were created.

Please revise the implementation so it chooses the execution strategy before running the coroutine rather than catching a broad RuntimeError. A long-lived background event loop, bridged with asyncio.run_coroutine_threadsafe, is the direction proposed on #125 and preserves resource affinity.

Please add regression coverage for:
1. an async tool called from a normal sync context;
2. an async tool called while the caller thread already has a running loop;
3. an async tool that raises RuntimeError, proving the original exception is preserved and the coroutine runs exactly once;
4. two calls sharing one loop-bound async resource.

Once those cases pass and CI is green, we can review this again.

sanmaxdev:
Updated the execution path to use one long-lived background event loop. This preserves loop-bound resources and propagates the original `RuntimeError` without retrying the coroutine.

Added regression coverage for a running caller loop, single execution on failure, and shared loop affinity across calls. Verification: `python -m pytest tests/unit/test_tool_executor.py -q` and the full non-network suite (`2307 passed`).

sanmaxdev:
Rebased onto the current main branch and kept the async loop-affinity fix intact.

Verified locally with:
`pytest tests/ -v --tb=short -m "not slow and not real_api and not network"`

All CI matrix jobs are running on the updated head now.

sanmaxdev:
Found the actual fork-PR CI blocker: GitHub injects an empty `OPENONION_API_KEY` when repository secrets are unavailable, which overrides the test fixture key.

Added a harmless fallback only for the offline test matrix. Reproduced the no-secret environment locally: 2410 passed, 5 skipped.

wu-changxing:
Reviewed and pushed one fix to this branch (f0532fc) — thanks for the contribution, the core approach is right.

**Review notes**

The design choice I wanted to confirm: this does *not* make the agent loop async. `_run_async_tool` blocks on `future.result()`, so `execute_single_tool` still returns a completed result and tool results land in the message list at exactly the same point as before. Sequential order, same context semantics. That's the right scope — actual concurrent tool execution is a separate change (it would need the `stop_signal` short-circuit, ordered logger output, trace ordering, and the `builtins.xray` global to be reworked). Tracking that separately.

Also verified there's only one `tool_func(**tool_args)` call site in the codebase, so no execution path was missed; `auto_debug.py` wraps `execute_single_tool` and inherits the behaviour.

**The bug I found**

The shared loop is single-threaded, so a *nested* tool execution deadlocks:

```
async tool  →  agent.input() (sub-agent)  →  sub-agent's own async tool
```

The inner `_run_async_tool` runs on the shared loop's own thread and submits back to that same loop — which is blocked waiting on the outer call. It hangs forever with no error. This is reachable today via `useful_plugins/subagents.py`.

Fix: detect that we're already on the loop thread and run the nested coroutine on its own throwaway loop. It loses resource affinity with the outer loop, but it completes instead of hanging. Added `test_nested_async_tool_does_not_deadlock` — it times out at 10s without the fix and passes with it.

**Not blocking, noted for later:** `future.result()` has no timeout, so a hanging async tool hangs the agent. Same as a hanging sync tool today, so not a regression — it's covered by the hard-stop work in #216.

wu-changxing:
Status update so this isn't sitting here silently: **we're deferring the async-tool feature to 1.8.0**, alongside #208 (converting the browser layer to a single async architecture). Both are about the codebase becoming async-aware, so they belong in the same release rather than shipping piecemeal. Nothing wrong with the code — the review stands, and #125 moved with it.

Your other PR, #293, is **merged** — thank you, that was a severe one.

**One thing worth splitting out, if you're up for it.** This PR bundles two unrelated changes:

1. the async tool support in `tool_executor.py` (the feature, now 1.8.0), and
2. one line in `.github/workflows/tests.yml` giving fork PRs a placeholder `OPENONION_API_KEY`.

Change 2 is the reason every fork PR currently shows red CI for reasons that have nothing to do with the PR — your #293 failed on `test_cli_auth_google.py` / `test_cli_auth_microsoft.py` purely because the secrets aren't available to forks. That fix is valuable *now* and shouldn't wait for a feature release.

If you open it as its own one-line PR, it can land independently. (Fair warning: it will still need a maintainer with `workflow` scope to merge, since GitHub blocks workflow-file changes otherwise — that's on our side, not yours.)

wu-changxing:
Merged as #512 — your three code commits cherry-picked onto current main with authorship intact, so this lands as your work.

This PR went `CONFLICTING`, but only on **a comment**: main had since made the identical `secrets.OPENONION_API_KEY || 'test-key'` change with different wording. Rather than ask you to rebase a fork branch across our workflow history — which needs the `workflow` scope a fork PR cannot get — I replayed the code commits and dropped the now-redundant CI one.

I drove it rather than reading it, and both of your claims hold:

```
same loop across separate agent runs: True      ← the point of the design
nested async tool → sub-agent → async tool:     NO DEADLOCK
```

And your deadlock commit earns its keep — with that branch removed the nested case times out at 124, with it, exit 0.

Thanks for sticking with this through the review. Sharing one long-lived loop instead of `asyncio.run` per call is what makes async tools actually usable with aiohttp sessions and playwright handles, and that was your call to make.

Closing in favour of #512.

PR #188, opened 2026-07-07
Graceful stop: honor a client INTERRUPT at the iteration boundary

yyy900:
## Why

A "stop" button in a chat frontend needs the agent to actually stop, not just the UI to hide a spinner. This is the agent side of that feature.

## What

**`poll_interrupt`** — a new `after_iteration` handler in the `tool_approval` plugin (`approval.py`). It drains `agent.io.receive_all('INTERRUPT')` and sets the existing `stop_signal`. The iteration loop already halts on `stop_signal` right after `after_iteration` (`core/agent.py`, unchanged) and returns its closing message.

**No core change.** This reuses the exact primitive and placement that `no_progress_guard` (`after_iteration` → sets `stop_signal`) and `reject_hard` already use. Registered in the `tool_approval` list, so every agent that loads it (co ai + the minimal template) is stoppable. It rides the same `agent.io` client-message channel as `poll_mode_changes`; the relay forwards the frame verbatim, so no server change is needed.

## Graceful by design

Running `after_iteration` means the current step (LLM call + tool batch) finishes before the stop lands — nothing is aborted mid-flight. It does **not** kill an in-flight LLM generation or a long-running tool; the stop lands at the next iteration boundary. Because `after_iteration` runs immediately before the loop's `stop_signal` check, the signal is consumed the same iteration it is set — no wasted LLM call and no cross-turn leak.

The closing message is the loop's existing `"What would you like me to do?"` — a proper agent turn, not a silent cutoff.

## Tests

`tests/unit/test_tool_approval.py::TestPollInterrupt` and `tests/unit/test_agent.py::TestGracefulInterrupt`. Full affected suites (`test_tool_approval`, `test_agent`, `test_ulw`, `test_no_progress_guard`) — 105 passed.

## Coordination

Pairs with the oo-chat stop button (openonion/oo-chat#24, draft), which sends `{type:'INTERRUPT'}` over the existing WS. The button only actually stops once a deployed agent runs a connectonion build that includes this change.

### DIFF
--- a/connectonion/useful_plugins/tool_approval/__init__.py
+++ b/connectonion/useful_plugins/tool_approval/__init__.py
@@ -1,10 +1,10 @@
 """
 Purpose: WebSocket-based tool approval plugin with mode system and unified permissions
 LLM-Note:
-  Dependencies: imports from [./approval.py (check_approval, load_config_permissions, poll_mode_changes, handle_mode_change, get_current_mode), ./constants.py (VALID_MODES, DEFAULT_MODE)] | imported by [../.__init__.py useful_plugins package, cli/co_ai/agent.py, cli/templates/minimal/agent.py, ulw.py, skills.py] | tested by [tests/unit/test_tool_approval.py, tests/integration/test_config_permissions.py, tests/integration/test_bash_chain_permissions.py]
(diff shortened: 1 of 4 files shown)


sanmaxdev (commented):
Checked the interrupt hook with loop-boundary behavior. The focused unit files pass: `python3 -m pytest tests/unit/test_tool_approval.py tests/unit/test_agent.py -q`.

wu-changxing (approved):

PR #192, opened 2026-07-11
Fix/ulw session restore

### LINKED ISSUE #191: [BUG] Agent.input(session=…) drops plugin session fields — ULW mode silently resets to Safe
*<D-c>*Describe the bug**
`Agent.input(session=…)` silently discards plugin-owned session fields (`mode`, `skip_tool_approval`, `ulw_turns`, `ulw_turns_used`) when restoring a session. This causes the `tool_approval` plugin to see `mode` as `None` and fall back to Safe mode, prompting for approval on every tool call. On the web [oochat](https://chat.openonion.ai/) switching to ULW visibly steps back to "Safe" after a turn or two.

The bug hits both HTTP and WebSocket clients:
- **HTTP (`POST /input`):** `agent.io` is `None` (see `http_router.py:286`), so `poll_mode_changes` returns immediately. `mode_change` messages are impossible via HTTP. The session dict is the only path to activate ULW — and it's broken. **ULW via HTTP is completely non-functional.**
- **WebSocket (`/ws`):** `mode_change` messages work for first-time activation because `poll_mode_changes` picks them up. But on subsequent turns or reconnection, the client relies on the session dict (returned from the previous turn with `mode: "ulw"`) to persist mode state. The fields are dropped, so mode silently resets to Safe.

**To Reproduce**
1. Install ConnectOnion from `main` branch (v1.1.0, commit `7473487` or current HEAD)
2. Create an agent with `plugins=[tool_approval, ulw]`
3. Call `agent.input("hello", session={'session_id': 'test', 'mode': 'ulw', 'skip_tool_approval': True, 'ulw_turns': 100, 'messages': [], 'trace': [], 'turn': 0})`
4. Check `agent.current_session.get('mode')` → returns `None`
5. Observe that any dangerous tool execution triggers approval prompts

**Expected behavior**
All session fields — including plugin-owned ones like `mode`, `skip_tool_approval`, `ulw_turns` — should survive `Agent.input(session=…)` so that:
- HTTP clients can use ULW mode at all (currently impossible)
- WS clients maintain ULW mode across reconnections without re-sending `mode_change` every turn
- Any plugin that stores state in `current_session` can rely on session persistence

**Code Example**
```python
from connectonion.useful_plugins import tool_approval, ulw

agent = Agent(
    "test",
    plugins=[tool_approval, ulw],
    model="co/gemini-2.5-pro",
    quiet=True,
)

# Client sends a session with ULW mode already active (e.g. reconnecting)
session = {
    "session_id": "test-123",
    "mode": "ulw",
    "skip_tool_approval": True,
    "ulw_turns": 100,
    "messages": [],
    "trace": [],
    "turn": 0,
}

agent.input("hello", session=session, max_iterations=1)

# mode should be 'ulw' — but is None
print(agent.current_session.get("mode"))           # → None ❌
print(agent.current_session.get("skip_tool_approval"))  # → None ❌
print(agent.current_session.get("ulw_turns"))      # → None ❌
```

**Error Output**
No exception is raised — the bug is silent data loss. The session fields are simply absent:

```
agent.current_session.get('mode')            # None (expected 'ulw')
agent.current_session.get('skip_tool_approval')  # None (expected True)
agent.current_session.get('ulw_turns')       # None (expected 100)
```

**Environment:**
 - OS: macOS 15
 - Python Version: 3.12
 - ConnectOnion Version: v1.1.0 (commit `7473487`)
 - API Key configured: yes (managed keys via `OPENONION_API_KEY`)

**Additional context**

**Root cause:** `connectonion/core/agent.py:248-254` rebuilds the session with an explicit whitelist of only 4 fields, discarding everything else:

```python
# agent.py:248-254 (current code)
if session is not None:
    self.current_session = {
        'session_id': session.get('session_id'),
        'messages': list(session.get('messages', [])),
        'trace': list(session.get('trace', [])),
        'turn': session.get('turn', 0)
    }
```

**Proposed fix (3 lines):**
```python
if session is not None:
 self.current_session = {
                'session_id': session.get('session_id'),
                'messages': list(session.get('messages', [])),
                'trace': list(session.get('trace', [])),
                'turn': session.get('turn', 0),
                'mode': session.get('mode'),
                'skip_tool_approval': session.get('skip_tool_approval'),
                'ulw_turns': session.get('ulw_turns'),
                'ulw_turns_used': session.get('ulw_turns_used'),
            }```

**Why `merge_sessions()` and other middleware are fine:**
- `http_router.py:65` passes the session dict as-is to `agent.input()`
- `merge_sessions()` in `merge.py:18` is a pure comparison function — returns the winning dict intact, no field filtering
- `agent_io.py:110` stores the raw dict from the client, no filtering
- `poll_mode_changes` in `approval.py:702` correctly reads `agent.io.receive_all('mode_change')` — but `agent.io` is `None` for HTTP, so this path is dead for stateless clients
- The `check_approval` hook at `approval.py:470` correctly checks `agent.current_session.get('mode') == 'ulw'` — it just never finds `'ulw'` because the field was already stripped

**Impact:**
- iOS app: ULW toggle in UI has no effect — agent keeps prompting for approval
- Web platform: ULW switch visibly reverts to Safe after a turn
- Any plugin that stores runtime state in `current_session`: silently broken on session restore
- Docs reference: https://docs.connectonion.com/useful-plugins/ulw describes `mode_change` activation but doesn't cover session dict persistence across reconnects

### PR DESCRIPTION
**Pull request by zbochuan**
## Related Issue
   Fixes #191

   ## Type of Change
   - [ ] Documentation update

   ## Changes Made
   - Added `mode`, `skip_tool_approval`, `ulw_turns`, `ulw_turns_used` to the session reconstruction whitelist in `Agent.input()`
   - Uses the same whitelist pattern rather than `dict(session)`, so transient keys (`permissions`, `stop_signal`, `pending_tool`) remain blocked
   - Added `test_input_preserves_plugin_session_fields` and `test_input_does_not_leak_transient_keys`

   ## Testing
   - [x] I have tested this code locally
   - [x] All existing tests pass

   ## Checklist
   - [ ] I have commented my code, particularly in hard-to-understand areas
   - [ ] I have made corresponding changes to the documentation

### DIFF
--- a/connectonion/core/agent.py
+++ b/connectonion/core/agent.py
@@ -250,7 +250,11 @@ class Agent:
                 'session_id': session.get('session_id'),
                 'messages': list(session.get('messages', [])),
                 'trace': list(session.get('trace', [])),
-                'turn': session.get('turn', 0)
+                'turn': session.get('turn', 0),
+                'mode': session.get('mode'),
+                'skip_tool_approval': session.get('skip_tool_approval'),
+                'ulw_turns': session.get('ulw_turns'),
+                'ulw_turns_used': session.get('ulw_turns_used'),
             }
             # Start YAML session logging with session_id for thread safety
             self.logger.start_session(self.system_prompt, session_id=session.get('session_id'))
--- a/tests/unit/test_agent.py
+++ b/tests/unit/test_agent.py
@@ -720,3 +720,47 @@ def test_agent_input_file_path_traversal(tmp_path):
     uploads_dir = tmp_path / ".co" / "uploads"
     assert len(list(uploads_dir.glob("*_passwd"))) == 1
     assert not (tmp_path / "etc").exists()
+
+
+def test_input_preserves_plugin_session_fields():
+    """Agent.input(session=…) should preserve plugin-owned fields (mode, ulw_turns, etc.)."""
+    agent = Agent(name="test", llm=MockLLM(), log=False)
+
+    session = {
+        "session_id": "test-123",
+        "mode": "ulw",
+        "skip_tool_approval": True,
+        "ulw_turns": 100,
+        "ulw_turns_used": 5,
+        "messages": [],
+        "trace": [],
+        "turn": 0,
+    }
+
+    agent.input("hello", session=session)
+
+    assert agent.current_session.get("mode") == "ulw"
+    assert agent.current_session.get("skip_tool_approval") is True
+    assert agent.current_session.get("ulw_turns") == 100
+    assert agent.current_session.get("ulw_turns_used") == 5
+
+
+def test_input_does_not_leak_transient_keys():
+    """Agent.input(session=…) should not blindly restore server-internal keys."""
+    agent = Agent(name="test", llm=MockLLM(), log=False)
+
+    session = {
+        "session_id": "test-456",
+        "messages": [],
+        "trace": [],
+        "turn": 0,
+        "permissions": {"admin": True},
+        "stop_signal": True,
+        "pending_tool": "bash",
+    }
+
+    agent.input("hello", session=session)
+
+    assert "permissions" not in agent.current_session
+    assert "stop_signal" not in agent.current_session
+    assert "pending_tool" not in agent.current_session

### CODE REVIEW

**Review by sanmaxdev — COMMENTED** [review]
The explicit session-field allowlist preserves ULW state without restoring unrelated request data. The focused ULW, approval, and session tests pass, and the new regression coverage checks both preservation and exclusion.

**Comment by wu-changxing** [comment]
Heads-up: after #188 (graceful stop) merged — which also touched the `tool_approval` / session-reconstruction path — this went `CONFLICTING`.

Could you **rebase onto `main`** and re-verify the ULW plugin's session-state keys (`mode` / `skip_tool_approval` / `ulw_turns` / `ulw_turns_used`) still line up with the post-#188 whitelist in `Agent.input()`? The fix itself is correct and well-targeted; it just needs the rebase and a CI re-run so we can confirm the plugin session-state survives cleanly alongside the new interrupt handling.

**Comment by wu-changxing** [comment]
and also check the plugin of ulw, should we fix it in the ulw plugin?

**Comment by wu-changxing** [comment]
Following up on the design — two things, one architectural, one security.

**1. Don't hardcode plugin keys into core.** Adding `mode`/`skip_tool_approval`/`ulw_turns`/`ulw_turns_used` to the whitelist in `Agent.input()` couples `core/agent.py` to the ULW plugin — every future stateful plugin would then have to edit core the same way. The whitelist exists for a good reason (the incoming `session` is untrusted client input, so `dict(session)` would let a client inject `permissions`/`stop_signal`), so the fix isn't to widen it per-plugin but to give plugins one generic, namespaced slot they own:

```python
# core agent.py — one generic key; core never mentions ULW
'plugin_state': dict(session.get('plugin_state', {})),
```

Plugins then read/write `session['plugin_state']['ulw']`, and core stays generic — any future stateful plugin works with zero core edits. I'm opening a small PR that adds exactly this `plugin_state` slot to core; please rebase onto it and move ULW's turn state into `plugin_state['ulw']` rather than touching the core whitelist.

**2. Security — `skip_tool_approval` must not round-trip through the client session.** This matters more than the coupling. `skip_tool_approval` is the capability flag `tool_approval` reads to bypass *all* approvals. On `main` it is deliberately **not** in the restore whitelist, so it resets on each stateless call. If this PR restores it from the client-provided session, a client can send `{"skip_tool_approval": true}` and bypass tool approval entirely. So don't blindly restore it — derive "skip" server-side from a server-validated `mode`, and keep only the benign counters (`ulw_turns`/`ulw_turns_used`) in `plugin_state`. Worth settling the trust model here before re-landing.

Happy to help wire ULW onto the `plugin_state` slot once the core PR is in.

wu-changxing:
AI slop, auto close

wu-changxing:
Thanks for the contribution and for identifying the stateless session-restore problem in #191.

We are closing this PR because #194 supersedes the core approach. The incoming session is untrusted client input, so mode and skip_tool_approval must not be restored as client-controlled top-level capability flags. Adding ULW-specific keys directly to core/agent.py would also couple core to one plugin and require the same core edit for every future stateful plugin.

#194 adds one generic, namespaced plugin_state slot, keeps capability flags out of the restore surface, includes regression tests, and currently has green CI across Python 3.10-3.13. The remaining ULW wiring will be handled on top of that design, so no further work is needed on this branch.

Thanks again for surfacing the underlying bug.

zbochuan:
Thank you.

PR #194, opened 2026-07-12
fix(session): persist owner-bound ULW state across reconnects

### LINKED ISSUE #191: [BUG] Agent.input(session=…) drops plugin session fields — ULW mode silently resets to Safe

wu-changxing:
## Problem

`Agent.input(session=...)` discarded plugin-owned state, so ULW silently fell back to Safe after HTTP continuation or WebSocket reconnect (#191).

Restoring `mode`, `skip_tool_approval`, or every client-provided session key directly would fix the symptom but create an approval-bypass vulnerability: the session snapshot is untrusted client input.

## Fix

- Add a deep-copied generic `plugin_state` slot for non-authoritative plugin/UI state.
- Keep approval capabilities in server-only state bound to the authenticated owner and session ID.
- Add explicit authenticated `mode` handling for HTTP and WebSocket INPUT.
- Persist ULW as a bounded, generation-scoped server lease; consume/checkpoint the lease before privileged work and fail closed on malformed or exhausted state.
- Serialize same-session execution, protect execution generations from stale thread completion, and reject replayed or concurrently raced requests.
- Bind modern CONNECT/INPUT signatures to action, recipient, session, request ID, prompt, mode, attachments, and session digest.
- Keep stored transcripts authoritative across reconnects, including an intentionally empty transcript.
- Reject INPUT received during an active turn with a retryable error *before* claiming its request ID, so approval/ask/checkpoint mailboxes cannot consume it.
- Preserve compatible local and legacy live controls without persisting unbound capabilities.
- Require one hosted worker while lease/replay state is process-local.
- Update protocol, hosting, persistence, plugin, ULW, and embedded documentation.

## Security properties

- Client session data cannot forge `permissions`, `skip_tool_approval`, pending tools, stop signals, mode, or ULW budget.
- Capability state cannot cross owners or session IDs.
- Invalid owner/session/mode/prompt/attachment data is rejected before replay claim.
- Concurrent sockets cannot run two agent generations for the same session.
- Stale completions and stale ULW checkpoints cannot overwrite newer generations.
- Malformed persisted or wire state fails closed with a clean error.

## Compatibility

- Local callers can pass `mode=` explicitly.
- Legacy unsigned INPUT is limited to an idle Safe turn and its capabilities are never restored.
- Modern persistent clients use fully bound signed INPUT.
- A running-session INPUT returns `retryable: true`; retry the same signed frame after the current OUTPUT.

## Verification

Focused host/session/ULW/approval matrix after merging current `main`:

```text
510 passed
```

Repository-default suite after merging current `main`:

```text
2599 passed, 5 skipped, 139 deselected
```

Also completed:

- [GitHub Actions passed on Python 3.10–3.13](https://github.com/openonion/connectonion/actions/runs/29506570558).
- Independent code/security review: APPROVE, no blocker/high/medium findings.
- `python -m compileall`
- `git diff --check`
- No new dependencies.

## Related

Fixes #191. Supersedes the unsafe capability-restoration approach discussed in #192.


wu-changxing:
@zbochuan this is ready for your review and specifically targets the HTTP + WebSocket/oochat behavior you reported in #191.

Could you please verify the original user-facing paths, especially:

- ULW stays active across a signed HTTP follow-up and a same-owner WebSocket reconnect.
- Tool approval is skipped only while the bounded ULW lease is active.
- A forged client session (`mode`, `skip_tool_approval`, or edited counters) cannot activate or extend ULW.
- Reaching the lease limit returns to Safe rather than hanging.
- INPUT sent during an active turn returns a retryable error and succeeds when retried after OUTPUT.

The branch is merged with current `main`; 510 focused tests and the 2,599-test repository-default suite pass. An independent security/code review also found no remaining blocker/high/medium findings. [GitHub Actions passed on Python 3.10–3.13](https://github.com/openonion/connectonion/actions/runs/29506570558).

Your oochat/iOS reproduction is the most valuable final check.

zbochuan:
Thanks, @wu-changxing. Sorry that my initial proposal was too naive about the trust boundary between client-restored session state and server-owned capabilities. I had focused on preserving the state without fully considering that client-provided session fields could be forged.

I tested head `ae915f9` with my original ooChat/iOS reproduction, and the user-facing flow now works properly.  Approval is skipped while the trusted ULW lease is active just like the "yolo mode". I confirm the original #191 behavior now works as expected on my endscreenshot

zbochuan:
@wu-changxing  
The owner-bound ULW lifecycle looks strong, and the functionality of `ulm` is fine. One cleanup remains: `skip_tool_approval` is no longer part of the runtime design—`check_approval()` now relies exclusively on `is_ulw_active()`—but several comments and docstrings still describe it as an active bypass flag.

Could we remove those stale operational references and the redundant `handle_mode_change()` cleanup? I would retain only the centralized legacy-state sanitization and the negative tests proving that a client-supplied `skip_tool_approval` value is ignored.

That would make the intended lifecycle explicit: approval bypass exists only while a valid, consumed, owner-bound ULW lease is active. `skip_tool_approval` would remain only as rejected legacy input, not as an apparent second source of authority.

wu-changxing:
**Deferred to 1.6.0.**

wu-changxing:
#191 was fixed on main by #581 while this was open, by a narrower route. Recording the overlap so nobody works on it twice, and leaving the decision on this PR to its author.

## What landed

`Agent.input(session=…)` keeps everything the caller passed instead of rebuilding from four keys, and `input_handler` strips a defined set of keys from whatever arrives and re-applies them from what the server stored:

```python
SERVER_OWNED_SESSION_KEYS = (
    'mode', 'ulw_turns', 'ulw_turns_used', 'skip_tool_approval',
    'permissions', 'approval', 'requester',
)
```

Verified against current main, both directions:

```
reconnect, client sends only the basics
  mode: ulw | ulw_turns_used: 2 | skip_tool_approval: True

client forges its own ULW
  mode: None | skip_tool_approval: None
```

So the reconnect case this PR is named for works, and the approval bypass this PR's description warns about — *"the session snapshot is untrusted client input"* — is closed. That warning was right, and it is the same reasoning that shaped the merged fix; I reached it independently and only found this PR afterwards.

## What this PR still has that main does not

- a generic `plugin_state` slot, rather than a fixed list of server-owned keys
- ULW as a **bounded, generation-scoped lease** rather than a value restored from storage

The second is the one worth a decision. Main's version restores `ulw_turns_used` from the server's own record, so a client cannot replay an old grant or reset its budget — but there is no generation, so an operator who grants ULW twice has no way to distinguish the two grants in the record. Whether that matters is a judgement I do not think is mine to make on someone else's PR.

If the lease is still wanted, this rebases onto a smaller diff now: the session-restoration half is done, and what remains is the ULW accounting.

wu-changxing:
Reviewed against `main` as it stands after this release's work, because the two overlap heavily and in one place disagree.

## It solves things that were open

This PR already contains what #649 calls option 3 — per-request signature binding — and a replay defence considerably stronger than the one I merged in #703:

```
+    # so a valid signature cannot be replayed against a different session.
+                raise ValueError("request replay cache is full")
+              Replayed request IDs are rejected.
+    # Reserve the execution generation before claiming replay state. The CAS
```

Mine caches CONNECT signatures and refuses a repeat. This binds each signed request to action, recipient, session, request id, prompt, mode and attachments, and claims request ids — which is the real answer, and is why I left #649 open rather than closing it.

## And it disagrees with one thing that shipped

```python
+    if workers != 1:
+        raise ValueError(
+            "host() requires workers=1 because session leases, replay claims, "
+            "and active WebSocket sessions are process-local"
+        )
```

Current `main` does the opposite: `workers: 2` prints "not honoured — running one worker" and keeps serving (#639), and #687 made the scheduler safe across processes with a per-tick lock precisely so forking is not a data hazard. This PR would turn a warning into a hard failure at startup for anyone with `workers > 1` in `host.yaml`.

Neither position is obviously wrong. Process-local state with one worker is simpler and is what this PR's leases and generations assume; cross-process locks are what #687 and #706 chose. But they are different architectures and only one can be true.

## Rebase cost

38 files, and **13 of them were changed in this release**, including every file the two designs both touch:

```
connectonion/network/host/auth.py
connectonion/network/host/ws_router/connect.py
connectonion/network/host/ws_router/session.py
connectonion/network/host/session/{__init__,active,storage}.py
connectonion/network/host/{server,http_router}.py
connectonion/network/connect.py
tests/e2e/test_host.py, tests/unit/test_asgi_http.py …
```

This is not a mechanical rebase. `auth.py` now holds a replay cache and signed-GET helpers; `ws_router/connect.py` now has session ownership and the replay check; `session/storage.py` now has a cross-process lock.

## What I am not doing

Not merging and not closing it. Merging would silently change what `host()` accepts and would land a second replay mechanism beside mine; closing would throw away the one piece of work that answers #649 properly.

This is Aaron's call: whether 1.6.0 ships the smaller hardening already merged, and this lands after, or whether this is the design and my #703 should be reverted in favour of it.

wu-changxing:
Release triage: converted back to draft. This still contains unique security work (owner-bound capability state, session serialization, and fully bound INPUT), so I am not closing it. It is currently conflicting and spans core agent, HTTP/WS protocol, storage, ULW, approval, docs, and thousands of test lines. Before review it needs to be replayed on current main and split into focused properties; several adjacent replay/trust fixes have landed since this branch, so resolving the conflict wholesale would be unsafe. It is not a 1.6 merge candidate in its current form.

wu-changxing:
Closing this superseded, conflicting implementation after preserving its remaining unique security property in #763. The original reconnect bug #191 was fixed on main by #581, and focused PR #750 now covers per-command signature binding/replay protection. What remains here is the generation-scoped ULW lease, but this branch implements it with process-local state and a hard single-worker requirement that conflicts with the current cross-process host design. Replaying a 38-file July branch over current auth/session/storage code would be unsafe for 1.6.0. The branch and discussion remain available; #763 defines the focused cross-worker acceptance criteria.

PR #195, opened 2026-07-12
co browser: multi-agent tab CLI (Patchright)

wu-changxing:
## co browser: multi-agent tab CLI (Patchright)

> Branch name is a legacy label — the code uses **Patchright**, there is no "cloak" anywhere.

Turns `co browser` into a CLI several AI agents can share on one persistent, logged-in browser without stepping on each other's pages.

### What's new
- **`-t/--tab` targeting** — one grammar across direct functions and `do`; bare commands use the shared `main` tab.
- **Tab lifecycle** — `tab open [NAME] [--who X] [--for "..."]` / `tab ls [--json]` / `tab close NAME`.
- **Contention guard** — a second agent mid-task on a tab gets a loud exit-4 teaching error naming the owner and the 3-step lifecycle; discovery via that error and `tab ls`.
- **Wire v1 = JSON envelope** `{v,caller,tab,line}` — caller/tab are structured, so any character in a name is quote-safe (can't break shlex or kill the daemon).
- **Exit codes** — `0 ok · 1 fail · 2 usage · 3 unknown tab · 4 tab busy`.

### Hardening (from two multi-agent code reviews)
- Per-tab claims on `_tab_meta` (releasing a tab releases its claim; takeover after 120s expiry).
- Guard covers every tab + `tab close`; anonymous callers can't stomp a named claim.
- `serve()` liveness uses a **shared-context** check, so a page-less tab no longer tears down every agent's browser; a bad request errors to its own client instead of unwinding the daemon.
- Client retries a transient `ECONNREFUSED` instead of deleting a busy daemon's socket.
- `do` instruction no longer quote-mangled; `-t` no longer eats a verb's own argument.

### Docs & tests
- New `docs/co-browser.md`; refreshed code-map headers on the CLI trio.
- +~30 regression tests (guard, exit codes, socket round-trip, reclaim-keeps-registration). Browser/daemon/CLI suites green.

### Known follow-up (not in this PR)
- `browser.newtab` overwrites the whole `_tab_meta[key]` record, wiping the claim — an NL `do` task that opens a tab can be hijacked. Fix pending.


wu-changxing:
### Code review

Found 1 issue:

1. The daemon path silently defeats `go_to()`'s "declare purpose/who" enforcement. `dispatch()` calls `_claim()` before every page-driving verb, and `_claim()` unconditionally `setdefault`s `_tab_meta[key]` with a truthy `who`/`purpose`. By the time `BrowserAutomation.go_to()` runs its occupancy check (`occupied = bool(existing.get("purpose") and existing.get("who"))`), the tab always looks occupied — so the documented ValueError ("The FIRST navigation on an unoccupied tab must say who it's for and why") can never fire through `co browser go_to` or `do`. Reproduced by dispatching `go_to https://example.com` on a fresh tab through `BrowserDaemon.dispatch()`: it succeeds instead of raising. Tests miss it because occupancy tests call `go_to()` directly and daemon tests use a `StubBrowser`. Only `newtab` still enforces it (special-cased before `_claim()`), leaving inconsistent behavior between verbs. This is not listed in the PR's "Known follow-up" section.

`_claim()` pre-populating the meta: https://github.com/openonion/connectonion/blob/66104069ddcead558ad6e1a57a2312aa26eeaefa/connectonion/cli/browser_agent/daemon.py#L198-L208

The occupancy check it defeats: https://github.com/openonion/connectonion/blob/66104069ddcead558ad6e1a57a2312aa26eeaefa/connectonion/useful_tools/browser_tools/browser.py#L624-L628

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

wu-changxing:
### Concurrency review (paper simulation) + fixes pushed in 5cf2027

Four independent reviewers hand-traced interleavings of the scenario "two agents firing a command every 0.1s" (no browser executed — state traced on paper, every transition cited to code). No protocol deadlock, no cross-thread deadlock, and no claim ping-pong livelock exist: the wire protocol is half-duplex with explicit half-close, the serve-loop→browser-thread wait graph is one-directional, and rejected exit-4 requests never refresh a claim. But four real problems were found; three are fixed in 5cf2027, plus the earlier occupancy-bypass finding resolved as documentation.

**1. [Critical — fixed] Backlog overflow forked the world into two daemons.**
The daemon is single-threaded with `listen(8)`. One slow command (a 30s `do`) plus 0.1s-interval clients fills the backlog in ~0.9s; the 9th client gets ECONNREFUSED, burns its 2s retry (client.py correctly knows busy ≠ dead), then spawns a new daemon — whose `_bind()` probe also gets ECONNREFUSED and treated it as a stale socket, **unlinking the live daemon's socket**. Result: split-brain tab registry, second Chrome failing on the profile SingletonLock in a doomed spawn loop, and the original daemon becoming an unreachable zombie holding a live Chrome until manual kill.
Fix: the daemon records its pid at bind; `_bind` now yields to a recorded owner that is still running and only replaces a truly dead owner's socket. `_cleanup` removes the pid file only when it is its own, so an exiting daemon cannot mark its successor dead.
https://github.com/openonion/connectonion/blob/5cf20276bb26d921fbacb0f07cdd752fb6f70228/connectonion/cli/browser_agent/daemon.py#L111-L126
https://github.com/openonion/connectonion/blob/5cf20276bb26d921fbacb0f07cdd752fb6f70228/connectonion/cli/browser_agent/daemon.py#L484-L511

**2. [Fixed] `newtab` overwrote the whole `_tab_meta` record (the PR's disclosed follow-up — blast radius was larger than described).**
A `do`-invoked `newtab` tool call wiped the caller's own `caller`/`claim_at`, collapsing the 120s guard window to zero: the rival's next request took over the tab and inherited the original owner's live page (half-filled forms, login state). Also, `_newtab`'s numeric id allocator didn't skip names registered via `tab open`, so a collision could stomp a live claim without waiting for expiry. Fixed: merge instead of replace; skip taken ids.
https://github.com/openonion/connectonion/blob/5cf20276bb26d921fbacb0f07cdd752fb6f70228/connectonion/useful_tools/browser_tools/browser.py#L666-L671
https://github.com/openonion/connectonion/blob/5cf20276bb26d921fbacb0f07cdd752fb6f70228/connectonion/cli/browser_agent/daemon.py#L391-L396

**3. [Fixed] Permanent `tab close` still saved `_page_url`.**
A future tab reusing the name auto-navigated to the previous owner's last URL before its own first navigation — under the shared logged-in profile, that hands agent A's sensitive page to agent B. Fixed: the URL is recorded only on reclaim (forget=False) and dropped on permanent close.
https://github.com/openonion/connectonion/blob/5cf20276bb26d921fbacb0f07cdd752fb6f70228/connectonion/useful_tools/browser_tools/browser.py#L455-L470

**4. [Resolved as docs] The earlier finding that `_claim()` neutralizes `go_to()`'s purpose/who enforcement is intentional layering** — docs/co-browser.md promises "solo use needs no ceremony" on main, and named tabs declare at `tab open`. The `go_to`/`_claim` docstrings now state the two declaration layers explicitly instead of contradicting each other.

**Known follow-up (not fixed here): no timeout on the browser thread.**
Every `submit(run).result()` is unbounded. A Playwright call that never returns (e.g. `page.evaluate` on a wedged renderer, a hung driver, `wait_for_manual_login`'s `input()` with an inherited tty) wedges the serve loop inside dispatch — a daemon-wide hang for all clients. With fix 1 in place clients no longer destroy the hung daemon's socket, but they can't get a clear "daemon hung" signal either. Worth a generous watchdog in a follow-up.

Regression tests added for each fix (busy-vs-stale `_bind`, claim-preserving `newtab` merge, id-collision skip, URL-drop on permanent close); browser/daemon/CLI suites green (114 passed).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

wu-changxing:
### Iterative review rounds 2-5 (paper-simulated concurrency + expert quality gates) — merge-ready

Four more rounds of fresh-agent review after the initial fixes; each round hand-traced two-agents-at-0.1s interleavings on paper and applied a Linus-style quality bar (readable names, simple direct functions). Fixed across commits afb663c, 5f12a3a, c9a70a9:

**Correctness**
- An unknown verb (typo) no longer acquires a 120s tab claim — rejected before claiming.
- The daemon's "Tab released" shortcut bypassed URL cleanup: a reclaimed tab closed by name could still resurrect the previous owner's URL on a same-name reopen. All close paths now funnel through `close_tab` → `_release_tab`, which drops page + registration + remembered URL together.
- `close_tab()` with no argument closed the shared **main** tab instead of the bound session's own — a hosted agent omitting the key destroyed someone else's tab and dodged the claim guard. It now resolves to the caller's tab, and accepts `"main"` (the name `tab_status` prints).
- Closing a registered-but-never-navigated tab before the browser launched reported success but left a phantom registration locking the name for 120s.
- A late-exiting zombie daemon could unlink its successor's live socket in `_cleanup`; socket and pid file are now removed only while the pid file still names the exiting process.
- Releasing a page-less tab replied exit 0 with "close page failed: 'NoneType'…" — a false failure on a happy path.

**Contract honesty**
- Usage errors now actually return exit 2 as the documented exit-code table promises.
- The client tells "daemon busy" apart from "daemon did not start" (pid-file check), fails fast on a dead owner's stale socket, and waits on a ~15s wall-clock deadline instead of a silently-multiplying attempt count.
- Legacy `newtab` rejects a `-t` target instead of silently driving a different tab than named; `-t x help` prints help locally.

**Readability**
- `_claim` split into `_register_tab` + `_stamp_claim`; `_close_tab(forget=)` split into `_reclaim_tab` / `_release_tab`; `_ensure_page`'s hidden reclaim extracted as `_reclaim_idle_tabs`; dead liveness loop removed; `_teardown` clears the tab registry (no ghost "reserved" rows); `_occupancy_help` coaches its actual audience (tool callers) in tool vocabulary; browser.py's stale code-map header rewritten.

Every fix carries a regression test (browser/daemon/CLI suites: 122 passed; full unit + e2e/cli suite green). Final round verdict from an independent reviewer: no confirmed bugs remaining; residual items are microsecond-scale bind races and pid-reuse theoreticals explicitly judged not worth more machinery, plus the known browser-thread-timeout follow-up.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #203, opened 2026-07-13
oauth: keep Microsoft refresh tokens server-side

wu-changxing:
## Summary

Keep Microsoft refresh tokens in oo-api and make the ConnectOnion client a small, expiry-aware consumer of short-lived access tokens.

Depends on https://github.com/openonion/oo-api/pull/25 and must be released with that API contract.

## Refresh behavior

- Microsoft controls access-token lifetime and reports it in `expires_in`: ordinarily 60–90 minutes, while CAE tokens can last 20–28 hours.
- The client caches the access token only in instance memory until five minutes before reported expiry.
- On demand, it asks oo-api for a usable token; oo-api owns refresh-token rotation.
- A Graph `401` sends only the rejected token's SHA-256 digest, obtains a usable token, and retries exactly once.
- There is no periodic refresh-token timer. Inactive clients do nothing.

For this confidential web flow, Microsoft documents a 90-day default refresh-token lifetime. It rotates on use and can be revoked earlier, so reconnect remains a normal recovery path.

Official references:

- https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens#token-lifetime
- https://learn.microsoft.com/en-us/entra/identity-platform/refresh-tokens
- https://learn.microsoft.com/en-us/entra/identity-platform/msal-acquire-cache-tokens

## What changed

- `co auth microsoft` binds an ephemeral `127.0.0.1` listener before initializing authorization.
- The hosted callback bounces the one-use code to that listener; the authenticated CLI submits it to oo-api for exchange. A copied authorization URL cannot attach another browser user's mailbox to the sender's account.
- Strictly validate the Microsoft authorization URL, URL-safe authorization ID, loopback path/host/query, callback metadata, and one-use result.
- Poll correlated status only when the completion HTTP result is genuinely ambiguous; definitive failures return immediately.
- Store only `MICROSOFT_CONNECTED`, scopes, and optional email. Successful reconnect removes legacy local Microsoft access/refresh tokens and expiry fields.
- Outlook and Microsoft Calendar fetch access tokens from oo-api, accept legitimate CAE lifetimes up to 48 hours, and never receive a refresh token.
- Keep existing connection metadata unchanged on timeout, denial, malformed callbacks, or failed replacement authorization.
- Isolate CLI auth tests from the developer's real home directory.

The browser must run on the same computer as the CLI. Remote/headless authorization needs an explicit device-flow design rather than weakening loopback validation.

## Validation

- Microsoft CLI/loopback/Outlook/Calendar focused suite: 77 passed.
- Full offline client suite: 2,437 passed, 5 skipped, 139 deselected.
- GitHub Actions matrix on Python 3.10, 3.11, 3.12, and 3.13: passed.
- Compile checks, `git diff --check`, and live-secret pattern scan: passed.
- Independent correctness, security, and simplicity reviews: no remaining implementation blocker.

The full suite still emits pre-existing browser-daemon thread shutdown tracebacks after pytest exits successfully; no tests fail.

## Deployment and compatibility

1. Complete the oo-api migration and deploy openonion/oo-api#25.
2. Release this SDK/CLI change in the same coordinated rollout.
3. Users with legacy local Microsoft token entries run `co auth microsoft` once to replace them with metadata-only entries.

## Merge gates — keep draft

- Credentials previously committed to oo-api history must be rotated/revoked externally.
- The existing long-lived, non-revocable OpenOnion API credential needs explicit risk acceptance or redesign because it can request delegated Microsoft access tokens.

No merge is requested until the API dependency and operational security gates are resolved.

### DIFF
--- a/connectonion/cli/commands/auth_commands.py
+++ b/connectonion/cli/commands/auth_commands.py
@@ -12,6 +12,7 @@ LLM-Note:
 import sys
 import time
 import requests
+from requests import RequestException
 import json
 import webbrowser
 import os
@@ -197,7 +198,8 @@ def handle_google_auth():
         console.print("  [bold]co auth[/bold]     Get your OpenOnion API key\n")
         return
 
-    api_url = "https://oo.openonion.ai/api/v1/oauth"
+    backend_url = os.getenv("OPENONION_API_URL", "https://oo.openonion.ai")
+    api_url = f"{backend_url.rstrip('/')}/api/v1/oauth"
     headers = {"Authorization": f"Bearer {api_key}"}
 
     # Clear any existing connection first - this ensures we wait for NEW OAuth to complete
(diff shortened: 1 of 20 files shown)


wu-changxing:
close this one, keep it remotely

PR #205, opened 2026-07-15
fix(permissions): preserve user approvals across skill cleanup

### LINKED ISSUE #204: [BUG] Session approvals granted during a skill are erased when skill cleanup restores permissions
When a skill is active, the skills plugin snapshots the entire session["permissions"] dictionary.

  If the user approves another tool with scope: "session" while that skill is running, the approval is correctly added to the live permissions. However, when the skill completes, cleanup_scope() replaces the entire permissions dictionary with the old snapshot.

  As a result, session approvals granted during skill execution are silently lost, represented  `trust for this session` cannot function well. Here is the most straightforward [video](https://drive.google.com/file/d/1QoAxSJkt4fPc7qejy6nfnwYMGFdkPnlV/view?usp=sharing). In this video, it is tested on oochat but the real problem is at upstream.

  ## To Reproduce

  1. Install ConnectOnion 1.2.0.
  2. Start a skill permission scope.
  3. Grant a session-level user approval while the skill is active.
  4. Run cleanup_scope().
  5. Observe that the newly granted session approval has disappeared.

  ## Expected behavior

  Skill cleanup should remove only permissions granted by the skill.

  User approvals granted during skill execution, including permissions with source: "user" and expires.type: "session_end", should remain valid after the skill completes.

  ## Code Example

  from types import SimpleNamespace

  from connectonion.useful_plugins.skills import (
      _grant_skill_permissions,
      cleanup_scope,
  )
  from connectonion.useful_plugins.tool_approval.approval import (
      _save_session_approval,
  )

  agent = Agent(
      "skill-approval-repro",
      llm=SimpleNamespace(model="fake"),
      log=False,
      quiet=True,
  )
  agent.current_session = {
      "turn": 1,
      "messages": [],
      "trace": [],
      "permissions": {},
  }

  # A skill starts and snapshots the current permissions.
  _grant_skill_permissions(
      agent,
      "commit",
      ["Bash(git status)"],
  )

  # During the skill, the user approves another tool for the session.
  _save_session_approval(
      agent.current_session,
      "write",
      {"file_path": "notes.txt"},
  )

  print("Before cleanup:", agent.current_session["permissions"])

  # Skill finishes.
  cleanup_scope(agent)

  print("After cleanup:", agent.current_session["permissions"])

  assert "write" in agent.current_session["permissions"], (
      "session approval was lost"
  )

  ## Error Output

  Before cleanup: {
      'Bash(git status)': {
          'allowed': True,
          'source': 'skill',
          'reason': 'commit skill (turn 1)',
          'when': {'command': 'git status'},
          'expires': {'type': 'turn_end'}
      },
      'write': {
          'allowed': True,
          'source': 'user',
          'reason': 'approved for session',
          'expires': {'type': 'session_end'}
      }
  }

  After cleanup: {}

  Traceback (most recent call last):
    File "<stdin>", line 42, in <module>
  AssertionError: session approval was lost

  ## Environment

  - OS: macOS, arm64
  - Python Version: 3.13.11
  - ConnectOnion Version: 1.2.0
  - OpenAI API Key configured: not required for this reproduction

  ## Additional context

  _grant_skill_permissions() currently stores a deep copy of all permissions in:

  session["_permission_snapshot"]

  _restore_permissions() later performs an unconditional replacement:

  session["permissions"] = session.pop("_permission_snapshot")

  This preserves approvals that existed before the skill started, but erases permissions created or changed while the skill was active.

  Existing tests cover pre-existing user approvals but do not cover user approvals granted during skill execution.

  A possible fix is to track the permissions introduced or overwritten by each skill and remove only those entries during cleanup. If a skill overwrote an existing permission, its previous value should be restored, while newer non-skill permissions must be preserved.

zbochuan:
## Description

Fix skill cleanup so temporary skill permissions never leak beyond the current turn, while preserving session-scoped approvals that the user grants before or during skill execution.

## Related Issue

Fixes #204.

## Type of Change

- [x] Bug fix (non-breaking)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

- Keep one pre-skill permission snapshot across repeated or nested `skill()` calls in the same turn.
- Restore overwritten pre-skill permissions when skill execution completes.
- Preserve newer user approvals granted while a skill is active.
- Remove temporary skill permissions even when a stale session has no in-memory snapshot.
- Stamp skill grants with the current turn and reject expired `turn_end` permissions during approval checks.
- Continue rejecting permissions, pending tools, and stop signals supplied by an untrusted client session.
- Add regression coverage for repeated, nested, disjoint, and overlapping skills; user approval overrides; stale grants; and forged client state.
- Let fork PRs run the offline CI matrix with a non-secret placeholder key; no repository secret is exposed.\n- Add no dependencies and make no changes to `connectonion/core/`.

## Security Boundary

This PR deliberately does **not** restore approvals from a generic hosted-session snapshot. Persisted/client-derived session data is not a trusted capability source unless it is separately authenticated and owner-bound. The final diff keeps permission restoration local to the skill scope and enforces expiration at use time.

## Testing

Relevant regression suite:

```text
python -m pytest tests/unit/test_skills.py tests/unit/test_tool_approval.py tests/unit/test_config_permissions.py -q
92 passed
```

Repository default suite:

```text
python -m pytest -q
2415 passed, 5 skipped, 139 deselected
```

GitHub Actions: [Python 3.10–3.13 matrix passed](https://github.com/openonion/connectonion/actions/runs/29499104810).\n\nAn additional run that explicitly included `network` tests produced 2422 passes and one relay multi-turn failure: the external model replied that it could not remember the test number. That test is excluded by the repository's default selection and reproduces independently of these permission changes.

## Checklist

- [x] Code follows the project style
- [x] Maintainer security review completed
- [x] Regression tests added
- [x] Relevant and repository-default tests pass locally
- [x] `git diff --check` passes
- [x] No new dependencies

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -34,6 +34,8 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
           GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
-          OPENONION_API_KEY: ${{ secrets.OPENONION_API_KEY }}
+          # Fork PRs cannot read repository secrets. The selected suite is fully
+          # offline, but some unit tests still construct the managed-key client.
+          OPENONION_API_KEY: ${{ secrets.OPENONION_API_KEY || 'test-key-for-offline-suite' }}
         run: |
           pytest tests/ -v --tb=short -m "not slow and not real_api and not network"
--- a/connectonion/network/host/http_router.py
+++ b/connectonion/network/host/http_router.py
@@ -2,7 +2,7 @@
 Purpose: HTTP route handlers for agent hosting endpoints (POST /input, session/admin/health/info)
 LLM-Note:
   Dependencies: imports from [network/host/session/ (Session, SessionStorage, merge_sessions, session_to_chat_items), network/asgi/http (read_body, send_json, send_html, send_text, CORS_HEADERS), network/trust/http_admin] | imported by [network/host/server.py, network/asgi/http.py] | tested by [tests/unit/test_host_routes.py]
-  Data flow: input_handler() receives prompt+session → calls create_agent() factory → merges client+server session via merge_sessions() if stored exists → calls agent.input(prompt, session) → records 'running' shell + final 'done' to SessionStorage → returns {session_id, status, result, duration_ms, session, chat_items, server_newer}
+  Data flow: input_handler() receives prompt+session → calls create_agent() factory → exposes an isolated trusted server-session snapshot to plugins → merges client+server session via merge_sessions() if stored exists → calls agent.input(prompt, session) → records 'running' shell + final 'done' to SessionStorage → returns {session_id, status, result, duration_ms, session, chat_items, server_newer}
   State/Effects: reads/writes SessionStorage via storage.save()/get() | factory creates fresh agent per request (prevents state bleeding) | session records carry TTL expiry
   Integration: exposes input_handler, session_handler, sessions_handler, health_handler, info_handler, admin_logs/sessions/trust/admins handlers | used by server.py and ASGI adapters
   Performance: factory creates fresh agent per request (thread-safe) | SessionStorage TTL auto-cleanup | session continuation via session_id provided by client
@@ -17,6 +17,7 @@ Session ID ownership:
 import json
 import time
 import uuid
+from copy import deepcopy
 from functools import partial
 from pathlib import Path
 from typing import Callable
@@ -46,11 +47,18 @@ def input_handler(create_agent: Callable, storage: SessionStorage, prompt: str, 
     server_newer = False
 
     stored = storage.get(session_id)
+    trusted_server_session = None
     if stored and stored.session:
+        # Expose an isolated server-owned snapshot to plugins. Plugins decide
+        # which state they own; core never needs plugin-specific restore logic.
+        trusted_server_session = deepcopy(stored.session)
+
         session, server_newer = merge_sessions(
             client_session=session,
             server_session=stored.session
         )
+
+    agent._trusted_server_session = trusted_server_session
 
     record = Session(
         session_id=session_id,
--- a/connectonion/network/host/http_router.py
+++ b/connectonion/network/host/http_router.py
@@ -2,7 +2,7 @@
 Purpose: HTTP route handlers for agent hosting endpoints (POST /input, session/admin/health/info)
 LLM-Note:
   Dependencies: imports from [network/host/session/ (Session, SessionStorage, merge_sessions, session_to_chat_items), network/asgi/http (read_body, send_json, send_html, send_text, CORS_HEADERS), network/trust/http_admin] | imported by [network/host/server.py, network/asgi/http.py] | tested by [tests/unit/test_host_routes.py]
-  Data flow: input_handler() receives prompt+session → calls create_agent() factory → exposes an isolated trusted server-session snapshot to plugins → merges client+server session via merge_sessions() if stored exists → calls agent.input(prompt, session) → records 'running' shell + final 'done' to SessionStorage → returns {session_id, status, result, duration_ms, session, chat_items, server_newer}
+  Data flow: input_handler() receives prompt+session → calls create_agent() factory → merges client+server session via merge_sessions() if stored exists → calls agent.input(prompt, session) → records 'running' shell + final 'done' to SessionStorage → returns {session_id, status, result, duration_ms, session, chat_items, server_newer}
   State/Effects: reads/writes SessionStorage via storage.save()/get() | factory creates fresh agent per request (prevents state bleeding) | session records carry TTL expiry
   Integration: exposes input_handler, session_handler, sessions_handler, health_handler, info_handler, admin_logs/sessions/trust/admins handlers | used by server.py and ASGI adapters
   Performance: factory creates fresh agent per request (thread-safe) | SessionStorage TTL auto-cleanup | session continuation via session_id provided by client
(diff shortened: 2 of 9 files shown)


wu-changxing:
@zbochuan Thanks for the clear reproduction and the original fix. I pushed a maintainer amendment in `1db0e4d` after a security review.

Could you please review the updated diff and confirm the original #204 flow still behaves as expected?

The final version now:
- preserves user approvals across skill cleanup;
- handles repeated/nested/overlapping skills without leaking temporary grants;
- enforces `turn_end` at permission-check time; and
- removes the generic hosted-session approval restoration, because persisted client-derived session data is not yet an owner-bound trusted capability source.

Relevant regression suite: **92 passed**. Since GitHub does not allow a PR author to be formally requested as a reviewer on their own PR, this comment is the review/confirmation request.

wu-changxing:
@zbochuan The amended head is now `4702920`, and the GitHub Actions matrix is green on Python 3.10–3.13. Please use this head for your #204 behavior confirmation/review. Thanks again for reporting and reproducing it.

wu-changxing (approved):
Maintainer review complete. The permission lifetime and cleanup paths are covered by focused regression tests, the repository-default suite is green, and I also verified the PR tree against the current main branch locally (92 focused tests and 2,415 default-suite tests passed).

zbochuan:
Thanks @wu-changxing . First I am sorry for my native design 🤣. I reviewed head `4702920` and reran the exact #204 skill-cleanup regression. It passes: a user `session_end` approval granted while a skill is active survives cleanup, while the temporary skill permission is removed.

So I confirm that the minimal reproduction in #204 behaves as expected.

However, the end-to-end hosted behavior that originally exposed the problem is still a separate concern: each hosted prompt creates a new `Agent`, and session-scoped tool approvals are not restored across requests. I understand that generic hosted-session permission restoration was intentionally removed for security reasons.

I will track the hosted cross-request approval behavior separately.

zbochuan:
PR #205 fully fixes the reproduction in #204: a user approval granted while a skill is active now survives skill cleanup, while temporary skill permissions are removed.

There is a separate hosted-session behavior worth tracking. Each hosted request creates a new `Agent`, and `Agent._input_once()` deliberately does not restore `permissions` from the client-provided session. As a result, an approval granted with `scope: "session"` is lost on the next hosted request, even when the authenticated owner and session ID are unchanged.

PR #194's owner-bound `server_state` appears to provide the right trust boundary for a follow-up. Rather than restoring the general `session["permissions"]` dictionary, the tool-approval plugin could maintain a canonical server-only ledger of user approvals with these properties:

- An approval is recorded only when the server processes an explicit user approval through the live approval flow.
- The ledger is bound to the authenticated owner and session ID.
- No approval is accepted from the client-provided session or `plugin_state`.
- Only user-granted, session-scoped approvals are persisted; config, safe, mode, and temporary skill permissions are excluded.
- Restore and merge behavior is owned by the tool-approval plugin, with expiry and revocation semantics enforced there.
- HTTP continuation and WebSocket reconnect paths have regression coverage, including forged client state and owner/session mismatches.

I will track this as a separate issue because it is outside #204's skill-cleanup scope. It should not block PR #194, but #194's trusted-state pipeline should make the follow-up possible without weakening the client-session trust boundary.

wu-changxing:
Reopening — this was closed on 2026-07-26 without a stated reason, and none of it landed.

**The bug it fixes is still live.** Re-reproduced today against `main` @ `4f289cf` (v1.5.2), both at unit level and through a full turn with the real `Agent` and the real `skills` + `tool_approval` plugins:

```
prompts during turn 1: ['write']
permissions after turn 1 cleanup: {}
prompts total after turn 2: ['write', 'write']
```

The user is still asked twice for a tool they trusted for the session. `skills.py:295-296` still does the unconditional whole-dict restore, and `git log -S"_permission_snapshot"` still returns exactly one commit — nothing superseded this.

**Everything here was already finished.** @zbochuan's original fix, the maintainer amendment in `1db0e4d8` after security review, green CI on Python 3.10–3.13, and @zbochuan's confirmation on head `4702920` that the #204 flow behaves correctly — a user `session_end` approval survives cleanup while the temporary skill grant is removed. The fork branch is intact at `47029209`, so the whole reviewed stack is recoverable as-is.

Two notes for whoever lands it:

**1. The last commit overlaps with #328.** `4702920` ("test(ci): run fork checks without repository secrets") makes the same one-line change as #328 — `secrets.OPENONION_API_KEY || <placeholder>` — differing only in the placeholder string. Worth noting that this exact fix has now been written three separate times: here on 2026-07-16, again in #184, and again in #328. Whichever lands first, the other should drop its copy. #328 is single-purpose and can land immediately; this stack needs a re-review pass first.

**2. The follow-up @zbochuan raised is still open and separate.** Each hosted request builds a new `Agent`, and `_input_once()` deliberately does not restore `permissions` from the client-supplied session — correctly so, since that dict is client-controlled and restoring it verbatim would make approvals forgeable. That's the same class of problem as #191 and shouldn't block this PR.

Also worth recording, found while re-verifying and not covered by this PR's tests: with the documented plugin order `[skills, tool_approval]`, the snapshot is taken before `load_config_permissions` runs, so cleanup wipes the entire config whitelist too — which is why the trace above shows `{}` rather than just the missing user grant.

Apologies for the churn, @zbochuan — this was a correct, reviewed fix and it should not have been dropped.

wu-changxing:
Retracting my previous comment — I reopened this without checking the diff properly, and I was wrong to characterise the close as unexplained or mistaken.

The close was a deliberate maintainer decision. Fixing this one bug pulled changes across three subsystems at once — `useful_plugins/skills.py`, `useful_plugins/tool_approval/approval.py`, and the session shape in `core/agent.py` — plus an unrelated CI change, with no single owner for the resulting permission model. That is a reasonable thing to decline regardless of whether each individual hunk is correct and the tests are green.

Closed again. The underlying #204 behaviour is still tracked in that issue; if it gets fixed, it should be as a scoped change with a settled ownership model, not this stack.

@zbochuan — my apology above was misdirected. The reporting and the reproduction were genuinely useful; the reason this did not land is architectural scope, not the quality of your diagnosis.

PR #206, opened 2026-07-16
Always persist Gmail/Outlook OAuth tokens to global ~/.co/keys.env

wu-changxing:
## Problem

`co auth google` and `co auth microsoft` only saved OAuth tokens to the global `~/.co/keys.env` **if the file already existed** — otherwise tokens landed only in the local project `.env`. On top of that, the token-refresh paths in `Gmail` and `Outlook` only rewrote lines that were already present in keys.env, so rotated tokens (Microsoft rotates the refresh token on every use) could be silently dropped.

Result: `~/.co/keys.env` was not guaranteed to hold the `co outlook` / `co gmail` tokens, and other projects/sessions lost access.

## Changes

- **`auth_commands.py`**: `handle_google_auth` and `handle_microsoft_auth` now always write credentials to `~/.co/keys.env` (creating `~/.co` if needed), in addition to the local `.env`.
- **`outlook.py` / `gmail.py`**: token refresh now uses `upsert_env` to persist refreshed/rotated tokens to global keys.env — missing keys are appended and the file is created if absent, instead of the old rewrite-only-existing-lines logic.

## Testing

- `pytest tests/unit/test_outlook.py tests/unit/test_gmail.py` — 63 passed
- Verified `upsert_env` creates a fresh keys.env, replaces existing keys, appends missing ones, and skips `None` (no rotated refresh token from older backends)

### DIFF
--- a/connectonion/cli/commands/auth_commands.py
+++ b/connectonion/cli/commands/auth_commands.py
@@ -257,11 +257,11 @@ def handle_google_auth():
     # Save credentials
     console.print("\n💾 Saving credentials...", style="cyan")
 
-    # Save to global ~/.co/keys.env
+    # Save to global ~/.co/keys.env (always, so every project can use the tokens)
     global_keys_env = Path.home() / ".co" / "keys.env"
-    if global_keys_env.exists():
-        _save_google_to_env(global_keys_env, credentials)
-        console.print(f"   ✓ Saved to {global_keys_env}", style="green")
+    global_keys_env.parent.mkdir(parents=True, exist_ok=True)
+    _save_google_to_env(global_keys_env, credentials)
+    console.print(f"   ✓ Saved to {global_keys_env}", style="green")
 
     # Save to local .env
     local_env = Path(".env")
@@ -350,11 +350,11 @@ def handle_microsoft_auth():
     # Save credentials
     console.print("\n💾 Saving credentials...", style="cyan")
 
-    # Save to global ~/.co/keys.env
+    # Save to global ~/.co/keys.env (always, so every project can use the tokens)
     global_keys_env = Path.home() / ".co" / "keys.env"
-    if global_keys_env.exists():
-        _save_microsoft_to_env(global_keys_env, credentials)
-        console.print(f"   ✓ Saved to {global_keys_env}", style="green")
+    global_keys_env.parent.mkdir(parents=True, exist_ok=True)
+    _save_microsoft_to_env(global_keys_env, credentials)
+    console.print(f"   ✓ Saved to {global_keys_env}", style="green")
 
     # Save to local .env
     local_env = Path(".env")
--- a/connectonion/useful_tools/gmail.py
+++ b/connectonion/useful_tools/gmail.py
@@ -53,6 +53,7 @@ Example:
 """
 
 import os
+from pathlib import Path
 import base64
 from google.oauth2.credentials import Credentials
 from googleapiclient.discovery import build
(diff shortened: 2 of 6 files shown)


connectonion/cli/commands/auth_commands.py
```diff
@@ -250,11 +250,11 @@ def handle_google_auth():
...
     console.print("\n💾 Saving credentials...", style="cyan")
 
-    # Save to global ~/.co/keys.env
+    # Save to global ~/.co/keys.env (always, so every project can use the tokens)
     global_keys_env = Path.home() / ".co" / "keys.env"
-    if global_keys_env.exists():
-        _save_google_to_env(global_keys_env, credentials)
-        console.print(f"   ✓ Saved to {global_keys_env}", style="green")
+    global_keys_env.parent.mkdir(parents=True, exist_ok=True)
+    _save_google_to_env(global_keys_env, credentials)
```
wu-changxing, line 263: 💡 Aaron says — **this creates two authoritative copies of the same rotating secret.**

This global write is followed by the existing local `.env` write, but Gmail/Outlook refresh only update the global file. ConnectOnion's package initialization prefers local `.env` and skips global `keys.env` whenever the local file exists. After a refresh succeeds, restarting from that project can therefore load the stale local token and undo the fix.

Pick one source of truth. The simpler contract is: OAuth secrets live in one global credential store; project files contain only non-secret metadata/reference state. If local copies must remain, both copies need one atomic persistence owner and explicit precedence—but duplicating rotating credentials is the root problem here.

connectonion/useful_tools/gmail.py
```diff
@@ -178,20 +179,14 @@ def _refresh_via_backend(self, refresh_token: str) -> str:
...
-            with open(env_file, 'w') as f:
-                for line in lines:
-                    if line.startswith("GOOGLE_ACCESS_TOKEN="):
-                        f.write(f"GOOGLE_ACCESS_TOKEN={new_access_token}\n")
-                    elif line.startswith("GOOGLE_TOKEN_EXPIRES_AT="):
-                        f.write(f"GOOGLE_TOKEN_EXPIRES_AT={expires_at}\n")
-                    else:
-                        f.write(line)
+        # Persist to global keys.env so the refreshed token survives this process
+        from ..cli.commands.project_cmd_lib import upsert_env
```
wu-changxing, line 180: 💡 Aaron says — **dependency inversion: a reusable tool should not import a CLI command module.**

`connectonion.useful_tools` is library/runtime code; `cli.commands.project_cmd_lib` is an application-layer module with unrelated CLI responsibilities. Hiding the import inside the method avoids an import cycle symptom, but it does not fix the architectural direction.

Move credential persistence into a feature-owned module (for example an OAuth credential store, not a generic `utils.py`) and have the auth CLI plus Gmail, GoogleCalendar, Outlook, and MicrosoftCalendar depend on that module. One owner also prevents the four refresh implementations from drifting again.

wu-changxing (commented):
## Architecture review — keep this Draft until token ownership is decided

Reusing one upsert implementation is directionally better than four hand-written rewrite loops, but the current patch still has no single OAuth credential contract.

### Blockers

1. **Two sources of truth remain.** Auth writes the same secrets to global `~/.co/keys.env` and local `.env`, while refresh updates only global state. ConnectOnion loads local `.env` first and does not load global keys when local exists, so a new process can resurrect the stale local access/refresh token.
2. **The dependency points the wrong way.** Reusable library tools now import a helper from `cli.commands.project_cmd_lib`. OAuth persistence should be a feature-owned module consumed by the CLI and every OAuth tool.
3. **The consumer set is incomplete.** `GoogleCalendar` and `MicrosoftCalendar` retain the old rewrite-only behavior; Microsoft Calendar also ignores a rotated refresh token. The bug still reproduces when Calendar refreshes first.
4. **This conflicts with the ownership model in #203.** #203 makes oo-api the Microsoft refresh-token owner; this PR expands local refresh-token persistence. Decide which contract wins before merging either path: https://github.com/openonion/connectonion/pull/203
5. **Broad prefix deletion crosses domains.** `_save_google_to_env(..., strip_prefix="GOOGLE_")` can delete `GOOGLE_API_KEY`, which belongs to the Gemini/LLM provider rather than Google OAuth. Replace only the explicit OAuth keys.

Please introduce one OAuth credential store/source of truth, route Gmail, GoogleCalendar, Outlook, MicrosoftCalendar, and the auth CLI through it, and add restart/precedence tests. There are currently no regression-test changes in this PR despite the persistence contract changing.

✨ The good part: replacing manual rewrite loops with an explicit upsert primitive is the right simplification; it just needs to live at the correct layer and own the complete consumer set.

**Assessment: keep Draft; architecture decision required before implementation is mergeable.**

wu-changxing:
Rebased onto main (after #243, #252, #253) and pushed two review fixes:

- **The fix was half a fix.** `load_dotenv` reads the project `.env` first and never overrides, so a stale `.env` shadows the rotated token in `keys.env` forever. Since Microsoft single-uses its refresh token, the connection still died. Refresh now updates a project `.env` too — but only one that already contains `MICROSOFT_ACCESS_TOKEN=`, so an unrelated `.env` never gets Microsoft keys scattered into it. Gmail deliberately left alone: Google refresh tokens don't rotate, so `keys.env` really is enough there.
- **`upsert_env` corrupted files with no trailing newline** — `"A=1"` + append → `A=1B=2`. Pre-existing, but this PR makes it reachable: the old code only rewrote existing lines, the new code appends.

Added `tests/unit/test_project_cmd_lib.py` (first direct coverage of `upsert_env`, which writes credentials) plus two `.env` rotation tests. Full unit suite: 2161 passed, 3 skipped.

PR #207, opened 2026-07-16
Add browser-workflow-skill-builder to useful_skills

wu-changxing:
## What

Ship `browser-workflow-skill-builder` as a copyable built-in skill, identical to the current `.codex/skills/` version (SKILL.md, 643 lines, plus the two script templates).

## Changes

- `connectonion/useful_skills/browser-workflow-skill-builder/` — SKILL.md + `scripts/extract-items-template.js` + `scripts/verify-item-template.js`, copied verbatim (`diff -r` clean against `.codex/skills/`)
- Registered in the `SKILLS` registry → `co copy browser-workflow-skill-builder` works
- Docs: `docs/useful_skills/browser-workflow-skill-builder.md`, rows in `docs/useful_skills/README.md` and `docs/cli/copy.md`
- Copy list table: `no_wrap` on the Name column so long skill names are not truncated to `…` (this broke the registry-driven list test), `overflow=fold` on Path

## Testing

- `tests/e2e/cli/test_cli_copy.py` — 19 passed (registry-driven tests pick up the new skill automatically)
- Manual: `co copy browser-workflow-skill-builder` in a clean dir copies all 3 files; `diff -r` against `.codex/skills/` confirms identical

https://claude.ai/code/session_016auL712vdMTjbBHvtLust5

### DIFF
--- a/connectonion/useful_skills/browser-workflow-skill-builder/SKILL.md
+++ b/connectonion/useful_skills/browser-workflow-skill-builder/SKILL.md
@@ -1,0 +1,643 @@
+---
+name: browser-workflow-skill-builder
+description: Create robust browser automation skills for sites like LinkedIn, X/Twitter, YouTube, Stripe, or other logged-in web apps by saving page context, analyzing HTML/CSS, writing skill-local JS extract/verify scripts, and using CSS selector actions with screenshot verification.
+tools:
+  - Bash
+  - Read
+  - Write
+  - Edit
+  - Glob
+---
+
+# Browser Workflow Skill Builder
+
+Create or improve browser automation skills for logged-in web apps. Use this when building skills that read a page, choose the correct item, generate or prepare content, click/type/upload, or submit/post/publish.
+
+The core pattern is: **save context -> analyze DOM/CSS -> write skill-local JS extraction -> verify same item by hash -> click/type with generic browser tools -> screenshot/context verify**.
+
+## Architecture Rule
+
+Do not add site-specific Python to the core browser tool.
+
+Use generic browser primitives:
+
+```text
+save_page_context(name)
+take_screenshot(path)
+run_page_script(script_path, args_json)
+run_frame_script(script_path, args_json, frame_url_contains, frame_name)
+click_element_by_selector(selector, index, text)
+type_text_by_selector(selector, text, index)
+click_element_near_selector(anchor_selector, target_selector, target_text, ...)
+upload_file_by_selector(selector, file_path, index, frame_url_contains, frame_name)
+upload_file_after_click_by_selector(click_selector, file_path, index, text, frame_url_contains, frame_name)
+```
+
+Put site-specific selectors and DOM logic in the target skill:
+
+```text
+.co/skills/<target-skill>/
+  SKILL.md
+  scripts/
+    extract-<items>.js
+    verify-<item>.js
+```
+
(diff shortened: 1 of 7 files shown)


connectonion/useful_skills/browser-workflow-skill-builder/SKILL.md
```diff
@@ -0,0 +1,617 @@
+---
+name: browser-workflow-skill-builder
+description: Create robust browser automation skills for sites like LinkedIn, X/Twitter, YouTube, Stripe, or other logged-in web apps by saving page context, analyzing HTML/CSS, writing skill-local JS extract/verify scripts, and using CSS selector actions with screenshot verification.
+tools:
+  - Bash
```
wu-changxing, line 5: 💡 Aaron says — **architecture blocker: this is a Claude Code permission contract copied into a ConnectOnion package.**

ConnectOnion exact-matches ordinary permission names; only the `Bash(...)` form receives special handling. The co-ai tools are named `bash`, `read_file`, `write`, `edit`, and `glob`, so `Bash` / `Read` / `Write` / `Edit` / `Glob` grant none of the intended scoped permissions.

"Copied verbatim" is not the right invariant when the source and destination runtimes have different tool contracts. Adapt the boundary with actual patterns such as `Bash(co browser *)`, `Bash(node *)`, `Bash(co ai *)`, `read_file`, `write`, `edit`, and `glob` (limited to what the workflow really needs), then test an actual invocation rather than only testing file copying.

connectonion/useful_skills/browser-workflow-skill-builder/scripts/verify-item-template.js
```diff
@@ -0,0 +1,118 @@
...
+      text,
+      text_hash: textHash,
+      action_index: action ? actionMatches.indexOf(action) : null,
+      has_action: Boolean(action)
+    };
+    scanned.push(item);
+
+    const authorOk = options.expectedAuthor ? compact(author) === compact(options.expectedAuthor) : true;
+    const textOk = options.expectedText ? compact(text) === compact(options.expectedText) : true;
+    const hashOk = options.expectedHash ? textHash === options.expectedHash : true;
```
wu-changxing, line 93: 💡 Aaron says — **safety blocker: a missing identity is treated as a successful match.**

When expected author/text/hash are omitted, all three predicates default to `true`; the first visible container with an action can therefore return `ok: true`. One malformed caller skips the exact-item gate this skill is designed to enforce.

Fail closed at the top: require `expected_text` and `expected_text_hash`, and require `expected_author` whenever an author selector is configured. Missing identity should return `ok: false` before scanning. Add negative tests for missing and mismatched fields.

wu-changxing (commented):
## Architecture review — execution contract is not validated

Keeping site-specific DOM logic in skill-local scripts instead of adding LinkedIn/X/etc. methods to browser core is exactly the right architecture.

The blockers are at the package boundary: this PR proves that files are copied byte-for-byte, but it does not prove that the copied skill is executable or safe in ConnectOnion. A skill's frontmatter and verification scripts are runtime contracts, not passive documentation.

Please fix the two inline findings and add integration coverage that:

- invokes the copied skill through the ConnectOnion skills + tool-approval plugins and verifies its real tool calls are permitted;
- calls the verifier without expected identity fields and proves it fails closed;
- calls it with wrong text/hash and proves no action can be authorized.

**Assessment: do not merge before the runtime contract is tested.**

Issue #208, opened 2026-07-16
RFC / Epic: migrate the browser layer to one async event-driven architecture


wu-changxing:
## Goal

Move the browser subsystem onto one async, event-driven core so unrelated sessions can make progress concurrently, while preserving the public synchronous `BrowserAutomation` API.

This is the architecture and acceptance epic. The implementation must land as independently reviewable child PRs.

## Decisions already made

- The internal driver core uses `patchright.async_api`.
- The public synchronous API remains supported through a compatibility facade; this is not an unannounced major-version break.
- Native Windows transport was delivered by #209 as a near-term seam. The async architecture may replace its transport internals, but must preserve its fail-closed authkey and liveness guarantees.
- Async tool functions are already supported by #125.

## Delivery breakdown

- [ ] #498 — port the browser driver core to `patchright.async_api`.
- [ ] #499 — replace the serial daemon loop with concurrent async sessions and safe tab claims.
- [ ] #500 — preserve the synchronous facade and migrate cross-platform IPC/tests.

## Required invariants

- One slow `co browser do` or tab must not block unrelated sessions.
- Tab claims/leases remain deterministic, owner-bound and auditable under interleaving operations.
- Cancellation, disconnect and timeout clean up browser/process/claim resources deterministically.
- Existing synchronous examples and public methods continue to work.
- POSIX and native Windows behavior are covered in CI.
- Patchright stealth checks and macOS keychain behavior are revalidated after the port.
- The async rewrite must not weaken the authentication, corrupt-authkey recovery or "no pipe = dead daemon" rules delivered by #209.

## Scope boundary

This epic is shared driver/runtime infrastructure for both the default Patchright engine and the licensed engine described by #511. It does not add anti-detection engine patches, licensing, session-capture product UX, or unrelated browser features.

Close this epic only after #498–#500 meet their definitions of done, or after an explicit architecture decision removes a child from scope. The comments below preserve the pre-#209 analysis and migration debt.

wu-changxing:
Status update: the near-term transport seam landed as PR #209 (native Windows named pipes + HMAC authkey, POSIX byte-identical, all CI green incl. `windows-latest`). What this RFC's async rewrite should **replace** when it lands — i.e., the deliberate debt the seam carries:

[...]

PR #209, opened 2026-07-16
Native Windows support for `co browser` (cross-platform daemon IPC)

wu-changxing:
## Description

Make `co browser` run **natively on Windows — no WSL** — by moving the daemon/client's POSIX-only IPC behind a thin cross-platform `transport` seam. The `co browser` daemon previously used `AF_UNIX` sockets + `fcntl.flock` + `os.kill(pid, 0)`, none of which exist on Windows. This adds a Windows transport (stdlib `multiprocessing.connection` **named pipes** with an HMAC `authkey`) alongside the existing POSIX path, which is left **byte-identical**.

Along the way, the new real-user Windows CI caught and this PR fixes three user-facing bugs (cp1252 emoji crash, scheme-carrying URL mangling, offline-first-run traceback), and first-run UX became zero-command: **`co browser` auto-installs its browser when none exists.**

## Related Issue

Related to #208 (async browser architecture RFC — the eventual unification; this PR is the near-term stepping stone). Narrows #213 (browser auto-install shipped here).

## Type of Change

## Changes Made

**Transport (native Windows IPC):**
- **`transport.py` (new)** — `default_address()` (POSIX unix socket unchanged / per-user named pipe), `pid_path`/`lock_path` sidecars (case-normalized on Windows), `pid_alive()` (EPERM = alive), `acquire_singleton_lock()` (flock / msvcrt), `spawn_detached()`, `load_or_create_authkey()` (fails closed, self-heals corrupt files atomically), `bounded_io()`/`accept_authenticated()` (every blocking pipe op under a hard deadline — mpc has none natively).
- **`daemon.py` / `client.py`** — Windows branches; POSIX wire byte-identical. "No pipe = daemon dead" on Windows (stale pid + recycled pid can never brick binding). Setup errors: one clean stderr line + exit 1, never a traceback (typer pretty-exceptions would print frame locals holding the HMAC secret).

**First-run experience:**
- **Browser auto-install:** the first page-driving `co browser` command with no system Chrome runs `python -m patchright install chrome` in the user's terminal, then continues. Pageless verbs (`status`/`tab`/`close`…) never provision; warm daemon pays zero. (`chrome_finder.py` shares detection between launcher and client.)
- **Offline first run is graceful:** `authenticate()` now has a 15s timeout and catches network errors — one friendly "continuing offline, run `co auth` later" line; `co init` still scaffolds + generates the keypair (it used to abort with a 4-frame traceback and NOTHING created).
- **cp1252 crash fixed:** CLI + daemon reconfigure their streams to UTF-8 on win32 — emoji output (e.g. `co doctor`'s 🔍) crashed every command on legacy codepages, including when piped by Claude Code/codex.
- **URL scheme fix:** `go_to` no longer rewrites `file://`/`about:`/`data:` URLs into `https://…` garbage (kept the bare-domain auto-prefix).

**CI — real Windows user simulation (`windows-e2e`) + expanded matrix:**
- Wheel install (what PyPI ships), CLI discovery from PowerShell **and cmd.exe**, `patchright install chrome`, package data in a path with **spaces + CJK**, real Chrome navigation (hermetic `file://`), **offline first run** (isolated USERPROFILE + dead proxy: exit 0, no traceback, scaffold + keypair created), **3 parallel agents** opening tabs concurrently, second cold start, browser.log dump on failure.
- `windows-browser` unit/daemon job matrix: 3.10 / 3.12 / 3.13.
- New tests: transport primitives + authkey healing + stalled-handshake anti-wedge (15), auth offline robustness (4), chrome auto-install logic (6), concurrency (8 simultaneous clients; cold-start singleton race, cross-platform).

No new runtime dependencies.

### Security note
Windows named pipes are ACL-scoped and HMAC-authenticated (`authkey`) — deliberately not loopback TCP, which any local process could connect to. POSIX keeps its `0700` socket-dir boundary; a failed chmod on it raises loudly.

## Example Usage

```powershell
# Fresh native Windows box (PowerShell or cmd), zero setup commands:
co browser go_to example.com     # auto-installs a browser on first run if needed
co browser do "log in and download my invoices"
co browser close
```

## Architecture assessment (honest)

**As a transition architecture this is, we believe, optimal under the constraints (no POSIX regression, fail-closed security, zero deps, shippable now). As an end-state it is deliberately NOT final** — that's #208's job. Known warts, owned explicitly:

1. **Dual wire framings** (POSIX stream+half-close vs mpc length-prefixed frames) — mirror-image maintenance risk across 4 sites; consistency traded for risk containment. The async rewrite picks ONE framing and deletes this.
2. **`bounded_io` is a thread-per-op shim** for mpc's missing timeouts. Correct and cheap at CLI rates; the root fix is asyncio Proactor pipes (#208).
3. **The daemon remains single-threaded serial** — concurrency is #208's headline.

Falsifiers to watch: mpc handshake flakiness on Windows CI; any protocol change applied to only one framing; mixed-Python-version daemon/client on one machine. Small follow-up: `co doctor` doesn't know the authkey sidecar exists (#213).

## Additional Notes

- POSIX intentionally byte-identical; the 67 pre-existing daemon tests pass unchanged.
- Reviewer's three blockers (fail-open authkey, unbounded handshake, POSIX authkey leak) fixed in 7cbc5e0 + 87c39aa with regression tests; see inline threads.


connectonion/cli/browser_agent/transport.py
```diff
@@ -0,0 +1,207 @@
...
+        except OSError:
+            break
+        try:
+            os.write(fd, binascii.hexlify(os.urandom(32)))
+        finally:
+            os.close(fd)
+    try:
+        return key_file.read_bytes().strip()
+    except OSError:
+        return b"co-browser-local"
```
wu-changxing: 💡 Aaron says — **security blocker: this credential primitive fails open.**

The loop validates an existing key, but after the retries the final read is returned without validating its length; if I/O fails, the code falls back to the public constant `b"co-browser-local"`. An empty/truncated file or an unwritable sidecar directory therefore turns the HMAC boundary into a predictable key while the daemon continues as if authentication were healthy.

The architecture contract here should be simple: a valid per-user secret exists, or browser IPC does not start. Please fail with the key path and context, or atomically replace an invalid file under synchronization. Never return a fixed/empty key. Add tests for empty, truncated, unreadable, and concurrent-create cases.

connectonion/cli/browser_agent/daemon.py
```diff
@@ -526,23 +529,76 @@ def _bind(self):
...
+        Path(transport.pid_path(self.sock_path)).write_text(str(os.getpid()))
+
+    # ---- transport seam: POSIX raw AF_UNIX socket vs Windows named-pipe Connection ----
+
+    def _accept(self):
+        """Accept one client. Returns a connection object, or None if the client failed
+        the Windows auth handshake (a bad handshake must not kill the single-threaded loop)."""
+        if transport.IS_WINDOWS:
+            try:
+                return self._srv.accept()  # mpc performs the HMAC handshake here
```
wu-changxing: 💡 Aaron says — **availability/architecture blocker: the 120-second timeout does not cover authentication.**

`multiprocessing.connection.Listener.accept()` performs both HMAC challenge exchanges before it returns. CPython does `deliver_challenge()` and `answer_challenge()` inside `accept()`, and those use blocking `recv_bytes()`: https://github.com/python/cpython/blob/v3.12.10/Lib/multiprocessing/connection.py#L479-L487

That means a client which connects and then stalls during the challenge wedges this single-threaded daemon indefinitely; execution never reaches `_read_request().poll(120)`. A crashed or hostile local client can make every later `co browser` command hang.

Please put accept/authentication behind a genuinely bounded boundary (for example, an isolated accept/auth worker that hands only authenticated connections to the single Playwright execution loop), and add a test where a raw pipe client connects but never answers the challenge.

connectonion/cli/browser_agent/daemon.py
```diff
@@ -508,12 +501,22 @@ def _bind(self):
...
-        self._bind_lock = open(self.sock_path + ".lock", "w")  # fd kept open for life
-        try:
-            fcntl.flock(self._bind_lock, fcntl.LOCK_EX | fcntl.LOCK_NB)
-        except OSError:
+        tells the two apart.
+
+        The lifetime lock, pid-file location, and (on Windows) the named-pipe wire live
+        behind `transport` so POSIX behavior is byte-identical while Windows gets native
+        named pipes with an HMAC-authenticated handshake."""
+        self._authkey = transport.load_or_create_authkey()  # used by the Windows pipe wire
```
wu-changxing: 💡 Aaron says — this leaks Windows-only state into the POSIX path.

`load_or_create_authkey()` is called unconditionally even though the key is only used by the Windows pipe transport. POSIX daemon startup now creates/reads an unused authkey file, so the claim that POSIX behavior is byte-identical is no longer true.

Keep the seam explicit: initialize the auth key only inside the Windows branch (and leave it unset on POSIX).

wu-changxing (commented):
## Architecture review — blockers before merge

The transport seam and native Windows CI are good, focused work. The remaining issues are at the security and concurrency boundaries, where a local browser-control daemon must fail closed and stay recoverable.

Please fix the three inline findings before merge:

1. Never degrade HMAC authentication to a fixed, empty, or truncated key.
2. Bound the authentication handshake itself; the current request timeout starts too late.
3. Keep Windows-only credential state out of the POSIX path.

After that, add regression tests for corrupt/unreadable auth keys and a client that connects but never completes authentication. The goal is a small transport boundary with explicit invariants, not hidden fallback states.

**Assessment: do not merge before these are fixed.**

connectonion/cli/browser_agent/transport.py
wu-changxing: Fixed across 7cbc5e0 + 87c39aa — the primitive now fails closed AND self-heals:

- The fixed-key fallback is gone entirely: `load_or_create_authkey()` returns a valid ≥16-byte key or raises with the key path ("could not create or read a valid browser IPC auth key at <path> — remove that file and retry").
- A file that stays invalid past the writers' microsecond create→write window (crashed writer / disk-full leftover) is **atomically replaced** (`O_EXCL` tmp + `os.replace`) instead of bricking every subsequent start.
- Concurrent cold-starts converge on one key (`O_EXCL` decides the writer; readers wait out the window).

Tests added in `tests/unit/test_transport.py`: empty file heals, truncated file heals, 8 concurrent callers converge on one key, plus the stability case. (An unreadable-file case degrades to the same invalid-key path; the atomic replace covers it where the OS allows, and the loud RuntimeError covers the rest.)

---

connectonion/cli/browser_agent/daemon.py
wu-changxing: Fixed in 87c39aa — you're right that the 120s bound started too late; the handshake (and two more unbounded pipe ops) are now all deadline-bounded:

- `win_listener()` now creates the pipe **without** an authkey, so `accept()` returns on raw connect. The mutual challenge runs in `transport.accept_authenticated()`: the public `deliver_challenge`/`answer_challenge` pair executes in a short-lived helper thread joined against a deadline (`HANDSHAKE_TIMEOUT`, 10s); on stall/failure the connection is closed (which errors the helper's blocked recv, so it exits) and the daemon keeps serving. The single Playwright loop only ever receives authenticated connections — the isolated-auth-boundary shape you suggested, without adding a persistent worker.
- The same `bounded_io()` primitive also covers the two other unbounded ops POSIX gets from `settimeout(120)`: a **partial frame** in `recv_bytes` (poll(120) only bounded time-to-first-byte) and a **reply into a full pipe** whose reader stopped consuming.
- A dead listener still raises out of accept(), so a dying daemon exits instead of spinning.

Test added (`test_wire_stalled_handshake_cannot_wedge_the_acceptor`): a raw client connects and never speaks the challenge → the acceptor drops it within the deadline and a well-behaved client succeeds immediately after. Runs on both CI legs (the helper is family-agnostic: AF_UNIX on POSIX, AF_PIPE on Windows).

---

connectonion/cli/browser_agent/daemon.py
wu-changxing: Fixed in 7cbc5e0 — `load_or_create_authkey()` moved inside the Windows branch of `_bind()`:

```python
if transport.IS_WINDOWS:
    self._authkey = transport.load_or_create_authkey()  # the pipe wire's HMAC secret
    self._bind_windows()
else:
    self._bind_posix()
```

POSIX daemons no longer create or read any authkey file; the client side is likewise Windows-gated (the loader is only called from `_connect_windows`). POSIX filesystem behavior is back to byte-identical.

---

PR #218, opened 2026-07-19
Add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- add an Atlas Cloud OpenAI-compatible LLM provider routed by `atlascloud/...` model prefixes
- wire `ATLASCLOUD_API_KEY` into CLI provider detection and generated `.env` config
- add focused tests for provider routing, endpoint setup, structured JSON mode, and CLI mapping

## Validation
- `uv run --extra dev pytest -q tests/unit/test_atlascloud_provider.py tests/unit/test_llm_errors.py::TestOpenAICompatibleProviders tests/unit/test_llm_errors.py::TestModelInference` (17 passed)
- `python3 -m compileall connectonion/core/llm.py connectonion/cli/commands/project_cmd_lib.py tests/unit/test_atlascloud_provider.py tests/unit/test_llm_errors.py`
- `git diff --check`
- live Atlas model catalog check confirmed `qwen/qwen3.5-flash` and `deepseek-ai/deepseek-v4-pro`

No README, sponsor, logo, credits, or partner promotion changes.

### DIFF
--- a/connectonion/cli/commands/project_cmd_lib.py
+++ b/connectonion/cli/commands/project_cmd_lib.py
@@ -512,6 +512,7 @@ def check_environment_for_api_keys() -> Optional[Tuple[str, str]]:
         ('GROQ_API_KEY', 'groq'),
         ('XAI_API_KEY', 'grok'),
         ('OPENROUTER_API_KEY', 'openrouter'),
+        ('ATLASCLOUD_API_KEY', 'atlascloud'),
     ]
 
     for env_var, provider in checks:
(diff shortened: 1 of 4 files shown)


wu-changxing:
Closing for now — thank you for the contribution, and this is about roadmap rather than the code.

Adding an LLM provider is a standing maintenance commitment: it has to be kept working as the provider's API moves, covered by the real-API test suite, and supported when it breaks for a user. We are not in a position to take that on for Atlas Cloud right now — the next few releases are committed to fixing existing behaviour rather than widening provider support.

Nothing here is wasted if that changes: the provider factory in `core/llm.py` is stable, so this rebases cleanly whenever we do revisit it.

If you want to keep using it in the meantime, a custom `LLM` subclass passed to `Agent(llm=...)` needs no fork.

Issue #237, opened 2026-07-23
Dashboard (DASHBOARD_SNAPSHOT) — Host support


wu-changxing:
## Dashboard (`DASHBOARD_SNAPSHOT`) — Host support (minimal)

Host slice of the OChat **Dashboard** feature. Design: **openonion/oo-chat#26**. The agent authors an ordinary `dashboard.html` in its working dir with normal Bash/Edit/Write tools; the Host sends it over the existing authenticated WebSocket. **oo-api/relay unchanged** (content-opaque).

### Wire contract (must match the SDK — openonion/connectonion-ts#16)
- `DASHBOARD_SNAPSHOT { html: string }`.

### Work — two emit points, both host-side, no plugin

1. **On connect** — ~4 lines in `establish_connection` (`network/host/ws_router/connect.py:96`), right after `await send_msg(connected_msg)`:
   ```python
   dash = Path("dashboard.html")          # cwd == project root (server.py:431)
   if dash.exists():
       await send_msg({"type": "DASHBOARD_SNAPSHOT", "html": dash.read_text()})
   ```
   (CONNECT-on-unknown-session is a verified read-only no-op, `connect.py:59-100`, so this is safe on the landing/draft connection.)
2. **After each run** — in the `OUTPUT` path of `forward_agent_msgs_to_client` (`network/host/ws_router/agent_io.py`), re-read and send the same frame (~3 lines). So a run that edited the dashboard shows the new version once it finishes.

3. **Starter template + Skill (the product):**
   - Ship a polished starter `dashboard.html` (empty-first, visual-over-textual — big numbers, whitespace, graphic-first, `prefers-color-scheme`), copied into the project root at `co` init when absent, filled with the agent name + skills.
   - Ship a built-in "update your dashboard" Skill at `cli/co_ai/skills/builtin/dashboard/SKILL.md` (discovery at `useful_plugins/skills.py:121`), teaching the agent to edit `dashboard.html` and keep the `data-ochat-skill` button contract.

### Explicitly NOT in v1 (this is where the old complexity was)
- ❌ No `after_each_tool` plugin / mid-run live streaming — updates land **after each run**. (Mid-run liveness is a clean v2 add: one opt-in plugin, no change to the above.)
- ❌ No `mtime` gate, no content hash, no `revision` — just send the small file. (An optional one-line `if html != last_sent: skip` can be added later; not needed now.)
[...]

PR #242, opened 2026-07-23
feat: dashboard.html delivery over the agent WebSocket (#237)

wu-changxing:
Implements openonion/connectonion#237 — the Host slice of the OChat Dashboard feature. Part of the three-repo rollout (master spec: openonion/oo-chat#26; sibling SDK: openonion/connectonion-ts#16 / PR openonion/connectonion-ts#17, **merged and published as 0.1.9**).

## What

The browser can't read a file inside the agent's container, so the Host reads `dashboard.html` from the agent's project directory and sends it over the existing authenticated WebSocket:

* **On connect** — in `establish_connection`, right after `CONNECTED`, so Home paints before any input. (CONNECT-on-unknown-session is a read-only no-op, so this is safe on the landing/draft connection.)
* **After each run** — in `forward_agent_msgs_to_client`, after `OUTPUT`, so a run that rewrote the dashboard shows the new version.

New `network/host/ws_router/dashboard.py` owns the feature:

* `read_dashboard_snapshot(session_id)` → the `DASHBOARD_SNAPSHOT { html }` frame (session_id stamped for relay routing), or `None` when there's no file — or when the file is larger than `MAX_DASHBOARD_BYTES` (**2MB**), unreadable, or not UTF-8. It's agent-authored input, so an unusable file degrades to "no Home" with a warning rather than raising out of the WS handler.
* `send_dashboard(send_msg, session_id, conn)` → the emit path. Reads off the event loop via `asyncio.to_thread`, and skips the send when the file's `(mtime, size)` matches what **this connection** last saw. Keyed on the per-socket `conn` dict, so a freshly connected client always paints while a run that didn't touch the dashboard costs no bandwidth. `resume_forwarding` threads `conn` through too, so the dedup also holds when a client reconnects to a still-running agent.
* `ensure_dashboard(agent_metadata, project_dir=None)` → writes a polished **empty-first** starter (agent name + up to 4 skills as `data-ochat-skill` buttons, `prefers-color-scheme` dark, no scripts) at host startup **only when `dashboard.html` is absent** — never clobbers the agent's file. It also anchors the directory every later read resolves against, so a mid-run `os.chdir` can't redirect which file gets served, and it warns instead of raising when the project dir isn't writable.
* Ships a built-in **Dashboard skill** (`cli/co_ai/skills/builtin/dashboard/SKILL.md`) teaching the agent to edit `dashboard.html` and keep the button contract.

**Why 2MB and not something tighter.** The client's CSP allows images only as `data:` URIs — external images are blocked — so inlining is the only way to put a chart or logo on a dashboard, and base64 adds ~33%. A single 400KB screenshot reaches ~530KB before any markup, so a smaller cap would reject pages the format actively pushes authors toward. 2MB still leaves ~8x headroom under the binding limit in the chain: the relay's uvicorn accepts WebSocket messages up to 16MB by default, and the snapshot crosses it as one frame.

**Starter buttons only feature published skills.** A client validates each button against the agent's published profile, which carries project-tree skills only — so a button naming a `user` or `builtin` skill would render and then silently refuse to run. `render_starter` and `_build_agent_profile` now draw from one shared allowlist, `PUBLISHED_SKILL_LOCATIONS` in `useful_plugins/skills.py`, and a test asserts the two lists agree.

## Deliberately minimal (v1)

No `after_each_tool` plugin / mid-run streaming, **no `core/` change**. Wire contract is just `DASHBOARD_SNAPSHOT { html }`. `oo-api`/relay untouched (content-opaque — the relay forwards any agent frame carrying `session_id`).

There is no `revision` on the wire and no hashing; suppressing an unchanged resend is a local `(mtime, size)` comparison per connection, which needs no protocol support.

## Docs

* `docs/network/dashboard.md` — new page: where the file lives, the starter written at host startup, editing it (by hand or via the built-in skill), the two sandbox constraints and the size cap, the button contract including the project-skills-only rule, when snapshots are sent, the wire format, and the module reference.
* `docs/network/websocket-protocol.md` — the `DASHBOARD_SNAPSHOT` server→client frame.
* `docs/network/README.md` — links the new page.
* `docs/features/skills.md` — names the four builtin skills, and explains that builtin and user-level skills aren't published to clients (the reason a button naming one never runs).

The public page is at `/dashboard` on the docs site (wu-changxing/connectonion-docs-site#19).

## Tests

`tests/unit/test_dashboard.py` — 26 tests: snapshot frame shape and session stamping, starter rendering (name escaping, ≤4 skill buttons, empty state, no invalid buttons, no scripting), no-clobber, the size-cap boundary at and over the limit, an unreadable path, per-connection dedup (skips unchanged, resends after a rewrite, independent across connections), skill-location filtering, starter/profile agreement, directory anchoring across a `chdir`, an unwritable project dir, and async integration tests proving the snapshot fires after `CONNECTED` and after `OUTPUT`.

Also fixed here: `test_docs_served()` called `create_app()` without chdir'ing into its `tmp_path`, so the starter `dashboard.html` landed in the repo root on every test run (and had been committed once). The test now chdirs, and `dashboard.html` is gitignored in this repo as a backstop.

All CI green (test 3.10–3.13, windows-browser 3.10–3.13, windows-e2e).


wu-changxing:
## Review

Reviewed as part of the three-repo rollout (sibling SDK openonion/connectonion-ts#17, UI openonion/oo-chat#34). Confirmed the "relay untouched / content-opaque" claim holds — oo-api's `relay/routes.py` forwards any agent frame carrying `session_id`, and the client-side handler stamps `session_id` on every client frame before the host sees it, so the on-connect snapshot routes correctly even for a brand-new session where the client sent no id. The two emit points are well chosen and the `create_app` addition is a good catch. Three issues.

**1. The starter dashboard can generate buttons a client will always refuse.** `render_starter` features `agent_metadata["skills"][:4]` unfiltered, but `_build_agent_profile` publishes only `location in ("project", "claude-project")` — and oo-chat validates every button against that published profile. So an agent whose first four skills are `user` or `builtin` gets a day-zero Home whose buttons render and then silently do nothing. It also diverges by transport, since a direct `/info` fetch isn't filtered the same way. Worth noting the new builtin `dashboard` skill is itself in that unpublished set.

Suggested fix: draw both lists from one shared allowlist so they can't drift, and have a test assert the starter's buttons agree with what the profile publishes. `SKILL.md` should state the constraint too — it currently tells the agent to avoid "internal/bootstrap skills" without explaining that location is what actually decides.

**2. Unbounded blocking read, on every connect and after every run.** `read_dashboard_snapshot` does `path.read_text()` with no size cap, synchronously, inside async WS handlers. `dashboard.html` is agent-authored, so an oversized one (or a file accidentally written there) blocks the host's event loop and ships the whole payload over the socket — and the relay forwards it verbatim. Worth a `stat().st_size` cap that degrades to "no Home" with a warning, plus `asyncio.to_thread` for the read.

Also `path.exists()` then `read_text()` isn't quite enough: `exists()` is true for a directory, and the read then raises `IsADirectoryError` out of the WS handler. Non-UTF-8 content raises `UnicodeDecodeError` the same way.

Separately, the post-run push re-ships the entire page after every turn even when the run never touched it. The per-socket `conn` dict is already threaded through both emit points, so caching the file's `(mtime, size)` there makes an unchanged Home free, while a freshly connected client (nothing recorded yet) still always paints.

**3. `ensure_dashboard` writes into the user's project dir unasked, and can raise at startup.** `Path("dashboard.html")` relative to cwd, from both `host()` and `create_app()` — so the first `co host` in a repo silently creates an untracked file with no log line, and an unwritable cwd raises straight out of host startup. A read-only project dir is a fine reason to have no dashboard; it isn't a reason to refuse to start.

Also, resolving `Path(DASHBOARD_FILE)` against the live cwd on every read means any later `os.chdir` (a tool, a plugin) redirects which file gets served. Better to capture the project dir once at startup.

---

All three are fixed on **`claude/dashboard-pr-review-qgtvsf`** (branched from this PR's head). `tests/unit/test_dashboard.py` goes to 26 passing, 15 of them new: size-cap boundary, unreadable path, per-connection dedup including resend-after-rewrite and independence across connections, location filtering, starter/profile skill agreement, cwd anchoring, and unwritable dir. The 35 failures elsewhere in `tests/unit` are pre-existing — identical set on this PR's untouched head.

---

wu-changxing:
Pushed the review fixes to this branch: `c99e386..e67ddac`.

**Three lines in the description are now out of date** — I didn't edit it directly because the API returns this body truncated (it cuts off mid-sentence in the Tests section, right where the `<script>` mention starts), so a rewrite would have silently dropped whatever follows. Worth a manual edit:

- **"Deliberately minimal (v1)"** says *"no `mtime`/hash/`revision`"*. There is now an `mtime`/size check: `send_dashboard()` compares the file's `(st_mtime_ns, st_size)` against what the current connection last saw, stored on the per-socket `conn` dict. A fresh client always paints (nothing recorded yet); a run that didn't touch the dashboard sends nothing. Still no `revision` on the wire — the contract is unchanged.
- **`read_dashboard_snapshot(session_id)`** is described as returning the frame *"or `None` when there's no file"*. It now also returns `None` — with a warning — when the file exceeds `MAX_DASHBOARD_BYTES` (512KB), is unreadable, or isn't valid UTF-8. The read moved off the event loop via `asyncio.to_thread`.
- **`ensure_dashboard(agent_metadata)`** now takes an optional `project_dir` and anchors the directory that every later read resolves against, so a subsequent `os.chdir` can't redirect which file gets served. It also logs the file it creates and warns instead of raising when the project dir isn't writable.

Also worth noting in the description: the starter dashboard now features only published (project-tree) skills, drawn from a shared `PUBLISHED_SKILL_LOCATIONS` allowlist in `useful_plugins/skills.py` that `_build_agent_profile` uses too — so it can no longer render a button a client will refuse to run.

Test count in the description (11) is now 26.

---

wu-changxing:
## Code review

**Overview:** Adds Host-side delivery of `dashboard.html` over the existing agent WebSocket (on connect + after each run, skipped when unchanged), a starter-file generator, a shared `PUBLISHED_SKILL_LOCATIONS` allowlist reused by both the announce profile and the starter dashboard's buttons, docs, and a 26-test unit suite.

### Issues

**1. Stray generated file committed at repo root — `dashboard.html`**
The diff adds a top-level `dashboard.html` (`<h1>docs-agent</h1>`) at the repository root. This looks like debris from locally exercising `ensure_dashboard()` against some "docs-agent" project that got `git add`-ed by mistake — the connectonion framework repo itself isn't a hosted-agent project and shouldn't ship a `dashboard.html`. Please remove it before merge.

**2. `send_dashboard()` only half-offloads its file I/O**
The docstring says the read runs in a worker thread "it's file I/O on the event loop otherwise," and `read_dashboard_snapshot` is correctly offloaded via `asyncio.to_thread`. But the leading `dashboard_path().stat()` call in `send_dashboard` still runs directly on the event loop before that offload, and `read_dashboard_snapshot` does its *own* `stat()` again inside the thread — a redundant double-stat plus one blocking stat left un-offloaded. Harmless on local disk, but inconsistent with the stated intent and could stall the loop on a slow/network filesystem.

**3. Minor race in the per-connection dedup**
`send_dashboard` reads `stat()` → compares to `conn["dashboard_stamp"]` → awaits the threaded read → *then* writes the new stamp. If it's ever invoked twice concurrently for the same `conn` (e.g. a reconnect racing the tail of a run), both calls can pass the "changed" check before either updates the stamp, causing a duplicate send. Low impact, but worth a comment or an early stamp-write if it's meant to be race-proof.

**4. Duplicated startup wiring**
`ensure_dashboard(agent_metadata)` (plus its import) is pasted identically into both `host()` and `create_app()` in `server.py`. Matches the file's existing duplication style (metadata extraction is duplicated the same way), so not a blocker, but a candidate for a shared helper if this file gets touched again.

**5. PR description understates test coverage**
The body says the test file has "11 tests" (and cuts off mid-sentence), but the actual file has ~26 tests covering snapshot reading, starter rendering, both emit points, size capping, per-connection dedup, and skill-location filtering. Worth updating the description for reviewers.

### Done well
- Size cap (512KB) checked via `stat()` before reading content into memory.
- HTML escaping in `render_starter` is correct in both the text and attribute contexts (covered by `test_render_starter_escapes_agent_name`).
- `PUBLISHED_SKILL_LOCATIONS` is shared between `_build_agent_profile` and `featured_skills`, closing a real gap where a starter-dashboard button could reference a skill the client-side validator would silently refuse to run — and `test_starter_skill_names_match_what_the_profile_publishes` asserts the two stay in sync.
- All failure modes (missing/oversized/unreadable/binary file, unwritable project dir at startup) degrade to "no dashboard" with a console warning instead of crashing the host.
- Docs (`dashboard.md`, `websocket-protocol.md`, `skills.md`) are consistent with the code (byte cap, allowlist, emit points all match).

### Security note (not a blocker)
The frame carries raw, unsanitized HTML — safety depends entirely on the receiving client sandboxing it (stripped `<script>`, no external URLs), which lives in the sibling `oo-chat`/`connectonion-ts` repos, not here. Consistent with the PR's stated "content-opaque" design, but worth confirming those sibling PRs actually enforce the CSP/sandbox before this ships.

---

Issue #232, opened 2026-07-22
31 unit tests fail in a clean checkout — they construct `Agent()` with the default managed model, which hard-requires `OPENONION_API_KEY`


wu-changxing:
## Summary

Running the unit suite in a clean checkout (no `OPENONION_API_KEY` in the environment) fails **31 tests across 4 files**. They aren't testing anything network-related — they just build an `Agent(...)` as a fixture, and `Agent.__init__` eagerly constructs the LLM for the default managed model `co/o4-mini`, which raises when no key is present. So the tests pass on a dev machine that happens to be authenticated and fail in CI / a fresh clone.

Discovered while running the full unit suite for the Windows GBK fix (#230). Confirmed pre-existing and unrelated to that branch.

## Failing tests (31)

```
tests/unit/test_prefer_write_tool.py      15
tests/unit/test_config_permissions.py      9
tests/unit/test_memory_integration.py      5
tests/unit/test_subagents.py               2
```

All fail with the same error at construction time:

```
ValueError: OPENONION_API_KEY not found in environment.
Run 'co init' to get started or set OPENONION_API_KEY in your .env file.
  connectonion/core/agent.py:143  self.llm = create_llm(model=model, api_key=api_key)
```

## Root cause

`Agent.__init__` (`connectonion/core/agent.py:143`) calls `create_llm(model=model, api_key=api_key)` immediately. With the default model `co/o4-mini`, `create_llm()` routes to `OpenOnionLLM`, whose constructor requires `OPENONION_API_KEY` (`connectonion/llm.py`). None of these tests mock `create_llm`/the LLM or pass a key, and none set a non-managed model — so simply *instantiating the fixture agent* raises before the test body runs. The tests never actually make an API call; the key is only needed to get past construction.

This is the test-isolation counterpart of the closed user-facing report #38 (same `ValueError` from `Agent(...)` with no key).

## Proof it's purely the credential coupling

Setting any non-empty key makes all 31 (plus the other agent-constructing tests in those files) pass — no other fix needed:

```bash
# clean checkout, no .env up the tree
python -m pytest tests/unit/test_prefer_write_tool.py tests/unit/test_config_permissions.py \
  tests/unit/test_memory_integration.py tests/unit/test_subagents.py -q
# → 31 failed

OPENONION_API_KEY=dummy_test_key python -m pytest tests/unit/test_prefer_write_tool.py \
  tests/unit/test_config_permissions.py tests/unit/test_memory_integration.py \
[...]

PR #244, opened 2026-07-24
fix: make .env file a fallback instead of precondition in send_email

### LINKED ISSUE #223: send_email hard-requires a physical .env file, ignoring env vars — 4 unit tests fail in clean checkouts
## Summary

`send_email()` refuses to run unless a physical `.env` file exists on disk, even when `OPENONION_API_KEY` / `AGENT_EMAIL` are already set in the environment. This makes 4 unit tests in `tests/unit/test_email_functions.py` **environment-dependent**: they pass on a dev machine that happens to have a `.env` up the tree, and fail in a clean checkout / CI.

Discovered while running the full unit suite. Confirmed pre-existing (fails on `main`, unrelated to any current feature branch).

## Failing tests

```
tests/unit/test_email_functions.py::test_send_email_success       assert False is True
tests/unit/test_email_functions.py::test_send_email_not_activated 'Email not activated' not in "No .env file found. ..."
tests/unit/test_email_functions.py::test_invalid_email_address    "No .env file..." == 'Invalid email address: not-an-email'
tests/unit/test_email_functions.py::test_send_email_rate_limit    "No .env file..." == 'Rate limit exceeded'
```

## Root cause

`connectonion/useful_tools/send_email.py:35-59` searches for a `.env` file (cwd → 5 parents → `~/.co/keys.env`) and **returns early** if none is found, *before* it ever reads the environment at line 62:

```python
# send_email.py:55-63
if not env_file:
    return {
        "success": False,
        "error": "No .env file found. Run 'co init' or 'co auth' first."
    }

# Get authentication token and agent email from environment
token = os.getenv("OPENONION_API_KEY")   # never reached without a .env file
from_email = os.getenv("AGENT_EMAIL")
```

The 4 tests inject creds via `@patch.dict('os.environ', {'OPENONION_API_KEY': ..., 'AGENT_EMAIL': ...})` but never create a `.env` **file**. So the function short-circuits at the file check and returns `"No .env file found"` before reaching the mocked env logic (success path, 403 not-activated, email-format validation, 429 rate-limit). On a machine with a stray `.env` up the tree the early check passes and the tests go green — which is why this isn't caught locally.

(Note: `test_missing_credentials` accepts *either* `"OPENONION_API_KEY"` or `"No .env file"` in the error, so it passes regardless — that's why only these 4 fail.)

## Two problems here

**1. Product behavior (the real bug).** Requiring a physical `.env` file is wrong when the credentials are already in `os.environ`. A container / CI / systemd deployment that injects `OPENONION_API_KEY` and `AGENT_EMAIL` directly as environment variables — no `.env` file — gets `"No .env file found"` even though the creds are present. `send_email` should work when the env vars are set, regardless of whether a `.env` file exists.

**2. Test isolation.** The tests mock `os.environ` but not the filesystem `.env` search, so they depend on ambient state.

## Suggested fix

Make the `.env` file a *fallback*, not a *precondition* — load it if present to populate the environment, then read from `os.environ` and only fail when the actual credentials are missing:

```python
# load .env if we can find one (populates os.environ), but don't require it
if env_file:
    load_dotenv(env_file)

token = os.getenv("OPENONION_API_KEY")
from_email = os.getenv("AGENT_EMAIL")
if not token:
    return {"success": False, "error": "OPENONION_API_KEY not set. Run 'co auth' to authenticate."}
if not from_email:
    return {"success": False, "error": "AGENT_EMAIL not set. Run 'co auth' to set up email."}
```

This fixes the product behavior *and* makes the 4 tests hermetic (they already set the env vars). `get_emails.py` may have the same pattern and is worth checking together.

## Repro

```bash
cd $(mktemp -d)          # anywhere with no .env up the tree
OPENONION_API_KEY=x AGENT_EMAIL=a@b.co \
  python -m pytest <repo>/tests/unit/test_email_functions.py -q
# → 4 failed
```

sanmaxdev:
## Description
`send_email()` treated a physical `.env` file as a precondition, returning `"No .env file found"` before ever checking environment variables. Deployments that inject `OPENONION_API_KEY` / `AGENT_EMAIL` directly as env vars (containers, CI, systemd) would fail even though credentials were present.

## Related Issue
Fixes #223

## Type of Change

## Changes Made
- Load `.env` via `load_dotenv()` if one is found, instead of gating on its existence
- `load_dotenv` does not override values already in `os.environ`, so externally-set vars take precedence
- Error messages updated to say "not set" instead of "not found in .env" since the source is now the environment, not just the file
- Updated the LLM-Note data-flow and errors comments to reflect the new behavior
- Made `test_send_email_no_project` hermetic by mocking `Path.exists` so it no longer depends on ambient `.env` files on disk

## Testing
- [x] `python -m pytest tests/unit/test_email_functions.py -v` (18 passed)

## Checklist

### DIFF
--- a/connectonion/useful_tools/send_email.py
+++ b/connectonion/useful_tools/send_email.py
@@ -2,11 +2,11 @@
 Purpose: Send emails via OpenOnion API using agent's authenticated email address
 LLM-Note:
   Dependencies: imports from [os, json, yaml, requests, pathlib, typing, dotenv] | imported by [__init__.py, useful_tools/__init__.py] | tested by [tests/test_email_functions.py, tests/test_real_email.py]
-  Data flow: Agent calls send_email(to, subject, message) → searches for .env file (cwd → parent dirs → ~/.co/keys.env) → loads OPENONION_API_KEY and AGENT_EMAIL → validates email format → detects HTML vs plain text → POST to oo.openonion.ai/api/email with auth token → returns {success, message_id, from, error}
+  Data flow: Agent calls send_email(to, subject, message) → searches for .env file (cwd → parent dirs → ~/.co/keys.env) → loads .env via dotenv if found (does not override existing env vars) → loads OPENONION_API_KEY and AGENT_EMAIL from environment → validates email format → detects HTML vs plain text → POST to oo.openonion.ai/api/email with auth token → returns {success, message_id, from, error}
   State/Effects: reads .env files from filesystem | loads environment variables via dotenv | makes HTTP POST request to OpenOnion API | no local state persistence
   Integration: exposes send_email(to, subject, message) → returns dict | used as agent tool function | requires prior 'co auth' to set OPENONION_API_KEY and AGENT_EMAIL | API endpoint: POST /api/email with Bearer token
   Performance: file search up to 5 parent dirs | one HTTP request per email | no caching | synchronous (blocks on network)
-  Errors: returns {success: False, error: str} for: missing .env, missing keys, invalid email format, API failures | HTTP errors caught and wrapped | validates @ and . in email | let-it-crash pattern (returns errors, doesn't raise)
+  Errors: returns {success: False, error: str} for: missing credentials, invalid email format, API failures | HTTP errors caught and wrapped | validates @ and . in email | let-it-crash pattern (returns errors, doesn't raise)
 """
 
 import os
@@ -15,6 +15,8 @@ import yaml
 import requests
 from pathlib import Path
 from typing import Dict, Optional
+
+from dotenv import load_dotenv
 
 
 def send_email(to: str, subject: str, message: str) -> Dict:
@@ -32,7 +34,10 @@ def send_email(to: str, subject: str, message: str) -> Dict:
             - from (str): Sender email address
             - error (str): Error message if failed
     """
-    # Find .env file by searching up the directory tree
+    # Load .env if one is available (populates os.environ, does not override
+    # values that are already set), then read credentials from the environment.
+    # A physical .env file is a convenience, not a requirement: deployments that
+    # inject OPENONION_API_KEY / AGENT_EMAIL directly as env vars should work.
     env_file = None
     current_dir = Path.cwd()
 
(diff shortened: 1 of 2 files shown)


wu-changxing:
## Overview

This PR changes `send_email()` so a `.env` file is a fallback rather than a precondition: it loads `.env` via `load_dotenv()` if found instead of gating on its existence, updates the error messages, and adjusts `test_send_email_no_project` to mock `Path.exists` so it doesn't depend on ambient `.env` files. The diff itself is sound in isolation — but see the blocking issue below.

## Issues

- **This PR is now stale/superseded.** `main` already contains commit `d48460b` ("Make unit tests hermetic and send_email work without a .env file", #278, merged 2026-07-27), which fixes the exact same issue (#223) with functionally the same approach (`load_dotenv()` as fallback, "not set" error wording). That's why `mergeable_state` is `dirty`. The author should close this PR or rebase onto current `main` to confirm nothing further is needed — as written, this diff can't land cleanly and duplicates already-merged work.
- **Precedence edge case worth a mention (in either version):** `load_dotenv()`'s default `override=False` means it never overwrites an existing env var — including an empty string. If a deployment accidentally injects `OPENONION_API_KEY=""` (blank), the "convenience fallback" `.env` value is never applied even though the intent is "env vars take precedence, `.env` fills the rest." Minor, but worth a code comment or a `.strip()`/falsy-check before treating a var as "already set."
- Malformed/empty `.env` files aren't a real risk here — `load_dotenv()` degrades gracefully (skips bad lines, no-ops on empty files), so no crash path there.
- On the hermetic-test point: this PR's `@patch('pathlib.Path.exists', return_value=False)` does make the test hermetic against ambient `.env` files, but the version that actually landed on `main` (mocking `load_dotenv` directly via `patch.object`) is arguably tighter/more targeted, since it doesn't blanket-stub `Path.exists` for the whole test.

## Done well

- Correctly identifies and explains the env-var-precedence guarantee of `load_dotenv()` (doesn't override already-set vars) — matches actual `python-dotenv` behavior.
- Error message and LLM-Note/docstring updates are consistent with the behavior change (no stale "not found in .env" wording left behind).
- Small, focused diff with a clear rationale tied to a real deployment scenario (containers/CI/systemd).

---

wu-changxing:
Already fixed on main by #278, which landed today: `send_email` no longer treats a `.env` file as a precondition. The header now reads *"A .env file is a convenience fallback"* — environment variables set any other way (exported, injected by CI, loaded by the `co` CLI) are equally valid, which is what this PR set out to do.

This branch is also CONFLICTING with 4 failing checks now, so rebasing it would be work for a change that is already in.

Thank you for the report and the fix — the problem was real, and #278 credits the same root cause (#223, #232).

PR #248, opened 2026-07-26
Show credential sources and optional values in co status

### LINKED ISSUE #247: feat(cli): show API-key availability, source, and optional reveal in `co status`
## Problem

When a provider API key exists in a project `.env` file but is not loaded into the current process, commands can report that the key is missing. Today, `co status` does not make this distinction visible, so users may waste time copying keys, adding duplicate secrets, or debugging the wrong layer.

A recent workflow exposed the confusing state clearly:

- a valid `GEMINI_API_KEY` existed in a project `.env`;
- the command running from another working directory could not see it;
- the resulting error looked like the key did not exist at all.

## Proposal

Show a redacted credential/configuration section in the default `co status` output. Because the default reports only variable names, availability, and privacy-safe source labels—not values—users and agents should not need an extra flag to discover why a provider is unavailable.

Example:

```text
Credentials
  GEMINI_API_KEY     configured · process environment
  OPENAI_API_KEY     discovered  · <project>/.env · not loaded
  ANTHROPIC_API_KEY  missing
```

Useful fields:

- provider/environment-variable name;
- status: `configured`, `discovered · not loaded`, `missing`, or `conflict`;
- effective source: process environment, `<project>/.env`, `~/.co/keys.env`, or another supported store;
- precedence/conflict warning when multiple sources exist;
- actionable hint, such as running from the correct project or loading the intended `.env`.

## Explicit reveal

Also support `co status --reveal` (short form `-r`), consistent with `co keys --reveal`, for deliberate local inspection of full provider values.

The reveal view should:

- require the explicit flag every time;
- show a prominent warning before the values;
- list each discovered source separately so conflicting values can be diagnosed;
- render secret text literally rather than interpreting terminal markup;
- leave all credential files and `os.environ` unchanged.

Default output remains redacted. Machine-readable secret output is out of scope until it has an equally explicit and safe contract.

## Security requirements

- Without `--reveal`, never print a raw secret, partial secret, prefix, suffix, length, hash, or fingerprint.
- Apply default redaction to human-readable output, logs, exceptions, telemetry, and debug mode.
- Treat `--reveal` as the only exception: print full values to the requested terminal output, warn the user, and do not persist or transmit them elsewhere.
- Do not copy, load, export, or otherwise mutate credentials as part of `co status`; it must remain read-only.
- Display source paths in a privacy-conscious form such as `<project>/.env` or `~/.co/keys.env`, not unnecessary absolute paths.
- Agents should use the default source metadata first and invoke `--reveal` only when the user deliberately requests full values.
- Tests must use fake credentials and prove both default redaction and explicit reveal behavior.

## Acceptance criteria

- [ ] Default `co status` distinguishes an effective process variable from a key merely discovered in an unloaded `.env`.
- [ ] Supported provider keys are listed without exposing any part of their values by default.
- [ ] Source precedence and conflicts are visible and actionable.
- [ ] `co status --reveal` and `co status -r` deliberately show full values with their privacy-safe source labels.
- [ ] Reveal output includes a prominent do-not-share/log warning.
- [ ] Running either mode does not load, export, copy, or mutate credentials.
- [ ] Source paths are project-relative or privacy-preserving.
- [ ] Regression tests prove values cannot leak through default output and appear only through explicit reveal.

This keeps credential debugging agent-friendly with a strict default boundary and an intentional escape hatch for local diagnosis.

wu-changxing:
## What changed

- Show a redacted credential inventory by default in `co status`.
- Report provider, environment-variable name, status, and privacy-safe source labels.
- Distinguish `configured`, `discovered · not loaded`, `missing`, and `conflict`.
- Add explicit `co status --reveal` / `-r` support for full provider values.
- In reveal mode, list each source separately so conflicting keys can be diagnosed.
- Cover all provider credential variables currently supported by the LLM layer.
- Remove the previous partial OpenOnion key preview from default status output.
- Add CLI documentation and regression tests.

## Why

A credential can exist in a dotenv file while the command that needs it cannot see it. Agents and users need a safe default view of which credentials exist and where they were discovered, plus an intentional local escape hatch when comparing the actual values is necessary.

## Security properties

- Default `co status` never returns a value, prefix, suffix, hash, or fingerprint.
- Full values appear only after an explicit `--reveal` or `-r` flag and a prominent do-not-share/log warning.
- Secret values are rendered as literal terminal text rather than Rich markup.
- Reads dotenv files without exporting their contents to `os.environ`.
- Displays only safe source labels: `process environment`, `<project>/.env`, and `~/.co/keys.env`.
- Does not create, move, or modify credential files.

## Validation

- `67 passed` across focused status/help/keys CLI tests.
- Regression coverage verifies default output contains no full or partial key material.
- Reveal coverage verifies full fake values and each source appear only when requested.
- Ruff checks passed for changed Python files.
- Python compile check passed.
- `git diff --check` passed.
- Earlier wider CLI run: `280 passed, 3 skipped`; 23 unrelated failures required browser sockets, network access, or writes outside the sandbox.

Closes #247

### DIFF
--- a/connectonion/cli/commands/status_commands.py
+++ b/connectonion/cli/commands/status_commands.py
@@ -1,17 +1,20 @@
 """
-Purpose: Display account status including balance, usage, and email configuration without re-authenticating
+Purpose: Display account status, deployments, and redacted credential-source diagnostics without re-authenticating
 LLM-Note:
-  Dependencies: imports from [os, yaml, requests, pathlib, rich.console, rich.panel, dotenv.load_dotenv, jwt, address] | imported by [cli/main.py via handle_status()] | calls backend at [https://oo.openonion.ai/api/v1/auth] | tested by [tests/e2e/cli/test_cli_status.py]
-  Data flow: receives no args → load_api_key() checks OPENONION_API_KEY from env/local .env/global ~/.co/keys.env → address.load() reads Ed25519 keypair → creates fresh auth message with timestamp → address.sign() creates signature → POST to /api/v1/auth to get current user data → displays balance, credits, total spent, email, agent ID → warns if balance <= 0
-  State/Effects: no state modifications | makes network GET request to oo.openonion.ai | reads from env vars, .env, ~/.co/keys.env | writes to stdout via rich.Console and rich.Panel | does NOT update any files
-  Integration: exposes handle_status() for CLI | similar to authenticate() but read-only | relies on address module for signature generation | uses requests for HTTP calls | displays Rich panel with account info | checks OPENONION_API_KEY in 3 locations (priority: env var > local .env > global ~/.co/keys.env)
+  Dependencies: imports from [os, requests, pathlib, dotenv.dotenv_values, rich.console, rich.panel, rich.table, address] | imported by [cli/main.py via handle_status()] | calls backend at [https://oo.openonion.ai/api/v1/auth] | tested by [tests/e2e/cli/test_cli_status.py]
+  Data flow: receives no args → inspects supported provider variable names in process env/local .env/global ~/.co/keys.env without loading values → displays redacted name/status/source table → load_api_key() resolves OPENONION_API_KEY → address.load() reads Ed25519 keypair → creates fresh auth message with timestamp → address.sign() creates signature → POST to /api/v1/auth → displays account and deployments
+  State/Effects: no state modifications | makes network requests to oo.openonion.ai after local diagnostics | reads env vars, .env, ~/.co/keys.env without exporting them | writes only credential names/status/source paths, never values or fingerprints | does NOT update any files
+  Integration: exposes handle_status() for CLI | credential discovery supports every provider in core/llm.py | OpenOnion auth still uses load_api_key() priority | source paths are privacy-safe (<project>/.env and ~/.co/keys.env)
   Performance: network call to backend (1-2s) | signature generation is fast (<10ms) | file I/O for .env files
-  Errors: fails gracefully if OPENONION_API_KEY not found (prints message to run 'co auth') | fails if keys missing in .co/keys/ | fails if backend unreachable (prints HTTP error) | handles response errors with status code display
+  Errors: credential parse/read failures are treated as no discovered keys | account status fails gracefully if OPENONION_API_KEY or identity keys are missing | backend errors do not expose credential values
 """
 
 import os
-import requests
 from pathlib import Path
+from typing import Mapping
+
+import requests
+from dotenv import dotenv_values
 from rich.console import Console
 from rich.panel import Panel
 from rich.table import Table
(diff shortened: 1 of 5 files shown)


wu-changxing:
## Overview

Reviewed the diff to `connectonion/cli/commands/status_commands.py`, `cli/main.py`, docs, and `tests/e2e/cli/test_cli_status.py` (539/-44, 5 files). This is a security-sensitive change (credential-source display + `--reveal` full-value printing), reviewed with that lens rather than a style pass. Traced every claim in the PR description against the actual code.

## Security analysis (claim-by-claim)

- **"Default output never shows value/prefix/suffix/hash/fingerprint"** — confirmed. `_credential_rows()` only returns `provider/credential/status/source`; the old `api_key[:20]...` preview line in the account-info panel is correctly removed.
- **"Full values only after `--reveal`/`-r`"** — confirmed. `_revealed_credential_rows()` is only invoked from inside the `if reveal:` branch of `_show_credentials`.
- **"Secrets rendered as literal terminal text, not Rich markup"** — confirmed for Rich's own markup parser: the Value column wraps the secret in `rich.text.Text(row["value"])`, which bypasses `Text.from_markup`, so a value like `[bold]x[/bold]` won't be styled/stripped by Rich. **However**, this only defeats Rich-markup injection — it does *not* sanitize raw ANSI/control-byte sequences. If a dotenv value contained literal escape bytes, they'd still reach the real terminal untouched (Rich's `Text` doesn't strip control characters). Given this is explicitly called out as a security property, worth either stripping non-printable characters before display or explicitly scoping the claim to "Rich markup injection" only.
- **"Dotenv read without exporting to os.environ"** — confirmed, `_read_credential_file` uses `dotenv_values(path, interpolate=False)` (interpolation disabled too, so a value like `${OTHER_SECRET}` won't be expanded/leaked). Good.
- **"Only safe source labels"** — confirmed, sources are hardcoded strings (`process environment`, `<project>/.env`, `~/.co/keys.env`), never resolved absolute paths, so no home-directory/username leakage even in the conflict-diagnosis path.
- **Env var coverage** — cross-checked `CREDENTIAL_ENV_VARS` against `core/llm.py`'s documented env vars: exact match (OpenAI/Anthropic/Gemini+alias/Groq/xAI/OpenRouter/Mistral/OpenOnion).

## Issues

1. **Bulk reveal increases blast radius.** `--reveal` dumps *every* configured/discovered secret across all providers in one table. A user who wants to check one Anthropic key exposes OpenAI/Gemini/Groq/etc. keys simultaneously in the same terminal dump — more exposure than necessary for shared terminals, screen shares, or recordings. Consider an optional `--reveal <VAR>` / provider filter.
2. **Untested for the one property that matters most.** No test feeds a fake secret containing Rich markup syntax (e.g. `"[red on white]x[/]"`) through the reveal path to confirm it prints literally rather than being interpreted/stripped. Given this is the headline security property in the description, it should have a dedicated regression test.
3. **Default-output regression test is narrow.** `test_default_status_never_prints_api_key_material` only exercises `OPENONION_API_KEY` via a plain env var through the full `handle_status()` render pipeline. The `.env`-sourced, multi-provider, and `conflict` scenarios are only checked at the lower-level `_credential_rows()`/`repr()` layer, not through the actual console-rendering path used by `handle_status()`/`_show_credentials()`. Worth adding at least one full-pipeline case with a `.env`-sourced secret and a conflicting-source secret.
4. **Minor:** `_is_configured`'s placeholder detection (`your-`, `sk-your`, `<...>`) is a fixed heuristic list — not a vulnerability, but a real secret that happens to start with one of these prefixes would be silently reported as "missing" rather than "configured".

## Done well

- Clean separation between the redacted default path and the explicit reveal path, with no shared code that could accidentally leak values into the default table.
- `interpolate=False` on `dotenv_values` is a nice, easy-to-miss detail that prevents `${VAR}` expansion from indirectly leaking unrelated secrets.
- Removing the old partial API-key preview from default `status` output is a genuine, if small, fix to a pre-existing info leak.
- Prominent do-not-share warning is printed before the revealed table, and it's a boolean CLI flag (no secret material ever appears in argv/shell history from invoking `--reveal` itself).

---

wu-changxing:
**Deferred to 1.6.0.**

PR #268, opened 2026-07-27
Split screenshot capture from image reading

wu-changxing:
## What changed

- make `BrowserAutomation.take_screenshot()` save exactly one screenshot and return its path instead of a base64 data URL
- add a public `read_image(path)` tool for PNG, JPEG, GIF, and WebP files, with actionable path/format errors
- route image handling in the copyable `read_file` tool through `read_image`
- expose `read_image` to browser agents and update browser prompts/documentation for the explicit two-step workflow
- add separate unit coverage for screenshot capture and image reading

## Why

Screenshot capture and visual inspection were coupled, so every screenshot injected base64 image data into agent context even when the caller only needed saved evidence. Separating the operations keeps capture lightweight and makes the vision/context cost explicit.

## Impact

Callers that need visual inspection now use:

```python
path = browser.take_screenshot()
read_image(path)
```

Direct CLI screenshot calls continue to print the saved path.

## Validation

- `pytest -q tests/unit/test_read_image.py tests/unit/test_read_file_tool.py tests/unit/test_browser_automation.py tests/unit/test_cli_templates.py` — 41 passed
- `python -m compileall -q connectonion`
- `git diff --check`
- browser daemon non-socket coverage passed; socket-binding cases cannot run in the Codex workspace sandbox (`PermissionError: Operation not permitted`)

### DIFF
--- a/connectonion/cli/browser_agent/agent.py
+++ b/connectonion/cli/browser_agent/agent.py
@@ -16,6 +16,7 @@ from connectonion import Agent
 from connectonion.useful_plugins import image_result_formatter, ui_stream
 from dotenv import load_dotenv
 from connectonion.useful_tools.browser_tools import BrowserAutomation
+from connectonion.useful_tools import read_image
 
 # Prompt path for browser agent
 PROMPT_PATH = Path(__file__).parent / "prompts" / "agent.md"
@@ -39,7 +40,7 @@ def build_browser_agent(browser, api_key: str) -> Agent:
         model="co/gemini-3.6-flash",
         api_key=api_key,
         system_prompt=PROMPT_PATH,
-        tools=[browser],
+        tools=[browser, read_image],
         plugins=[image_result_formatter, ui_stream],
         max_iterations=200,
     )
--- a/connectonion/cli/browser_agent/prompts/agent.md
+++ b/connectonion/cli/browser_agent/prompts/agent.md
@@ -77,7 +77,7 @@ When something fails:
 1. Open browser if not already open
 2. Navigate to the target site
 3. Wait for page to load completely
-4. **Take a screenshot after navigation**
+4. **Take a screenshot after navigation, then call `read_image` with the returned path when you need to inspect it visually**
 
 ### Finding Elements
 - Use natural descriptions first
@@ -88,7 +88,7 @@ When something fails:
 - Use `click_element_near_selector(...)` when a skill provides an anchor selector and a nearby target selector.
 - Fall back to text matching if needed
 - Never expose CSS selectors to users
-- **Take a screenshot when you find important elements**
+- **Take a screenshot when you find important elements, then call `read_image` with its path to inspect it**
 
 ### Saving Page Context
 When a user wants to analyze a site's HTML/CSS, make a workflow more accurate, or debug why a click matched the wrong element, use `save_page_context(name)`.
@@ -100,12 +100,12 @@ It saves under `~/.co/browser_context/`:
 
 ### Form Filling
 1. Find all form fields first
-2. **Take a screenshot of the empty form**
+2. **Take a screenshot of the empty form and call `read_image` with its path**
 3. Generate appropriate values using user context
 4. Fill fields in logical order
-5. **Take a screenshot after filling**
+5. **Take a screenshot after filling and call `read_image` with its path**
 6. Validate before submission
-7. **Take a screenshot after submission**
+7. **Take a screenshot after submission and call `read_image` with its path**
 
 ### Completing Tasks
 - **Take screenshots at each major step**
--- a/connectonion/useful_tools/__init__.py
+++ b/connectonion/useful_tools/__init__.py
@@ -27,6 +27,7 @@ from .terminal import yes_no, autocomplete
 from .todo_list import TodoList
 from .slash_command import SlashCommand
 from .ask_user import ask_user
+from .read_image import read_image
 
 # Claude Code-style file tools
 from .file_tools import (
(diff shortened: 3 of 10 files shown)


wu-changxing:
## Overview

The diff matches the PR description: `BrowserAutomation.take_screenshot()` now saves once and returns a bare path (the old base64-inflation guard/JPEG-re-encode hack is removed entirely); a new `useful_tools/read_image.py` handles PNG/JPEG/GIF/WebP with clear error strings; `read_file.py`'s `_read_image` is deleted in favor of delegating to `read_image()`; `cli/browser_agent/agent.py` and `cli/templates/browser/agent.py` both add `read_image` to their tool list and their prompts are updated for the explicit two-step workflow.

## Issues

- **Silent regression in the `hosted-browser` template.** `connectonion/cli/templates/hosted-browser/agent.py` builds its agent with `tools=[browser, ask_user]` + `image_result_formatter`, and its prompt (`prompts/agent.md`) leans entirely on "take a screenshot and verify" with no mention of `read_image`. This file is *not* touched by the PR. Before this change, `take_screenshot()` returned a base64 data URL, which `image_result_formatter._is_base64_image()` matched directly (via the `data:image/...` regex) and auto-attached as a vision message. After this change it returns a bare path, which matches neither the data-URL branch, the raw-base64-chars heuristic, nor `_PATH_PATTERN` (that one requires a `"saved to:"` prefix). So this shipped, deployable agent template silently stops "seeing" its own screenshots — no error, just a model that can no longer visually verify the pages it's driving, exactly the kind of breakage the review was asked to look for. Worth either adding `read_image` + updating this template's prompt too, or flagging it explicitly as a known follow-up.

- **CLI/daemon path is untested against the new contract.** `tests/e2e/cli/test_browser_daemon.py` (not modified by this PR) has `test_dispatch_image_payload_shows_path_not_base64`, which stubs `take_screenshot()` to return a base64 string and asserts the daemon rewrites it to `"Screenshot saved to: {last_screenshot_path}"`. That rewrite path only fires when the daemon detects base64 in the result — which real `take_screenshot()` will never produce again. It's unverified whether the daemon still prefixes a bare path with `"Screenshot saved to:"` when passing it through untouched, and that exact prefix is what `image_result_formatter._PATH_PATTERN` (and bash-driven agents like `co_ai`, whose prompt says "the image is attached from the path it prints") depend on. This is a more relevant gap than the socket-binding sandbox limitation the PR calls out — it's specific to this change and wasn't exercised by the listed validation run (`test_read_image.py`, `test_read_file_tool.py`, `test_browser_automation.py`, `test_cli_templates.py`). The noted socket-binding gap itself is real but unrelated: no daemon socket code was touched here.

- **Stale doc left behind:** `docs/templates/browser.md` still has `take_screenshot(path, full_page) | Capture screenshots, returns base64`, which is now wrong and contradicts the (correctly updated) `docs/cli/browser.md` and `docs/useful_tools/browser_tools.md`.

- **Minor:** `read_image()`'s format check is extension-only (no magic-byte sniffing), so a mislabeled file (e.g. a PNG saved as `.jpg`) is happily mislabeled in the returned data URL's MIME type. This mirrors the old `_read_image` behavior in `read_file.py`, so it's not a regression, just an existing limitation now living in a more prominent, public tool.

## Done well

- The removal of the "oversized PNG → re-encode as JPEG at quality 85 to fit the base64-under-1MB relay limit" logic in `take_screenshot()` is a clean simplification enabled by no longer shipping image bytes through that path at all — good riddance to a real hack.
- `read_image()`'s error handling is fail-fast and actionable (`does not exist`, `is not a file`, `unsupported format` with the supported-list spelled out) — matches the project's "no try-except-pass, descriptive errors" standard, and returns strings rather than raising, consistent with how other tools in this codebase signal errors back to the LLM for retry.
- `read_file.py` now delegates to `read_image()` instead of duplicating the base64/MIME logic — a good DRY cleanup, and it means image-format bugs only need fixing in one place.
- New unit tests (`test_read_image.py`) cover the happy path plus all three error branches (missing file, directory, unsupported extension), and `test_browser_automation.py`'s new test correctly asserts the in-process return value/`last_screenshot_path`/screenshot args.
- The prompt updates in `cli/browser_agent/prompts/agent.md` and the two-step `take_screenshot` → `read_image` guidance are threaded consistently through every workflow section (navigation, element-finding, form-filling).

---

wu-changxing:
**Deferred to 1.6.0.**

wu-changxing:
Rebased onto `main` as part of the 1.6.0 PR sweep. One conflict, and it resolved by deletion rather than by merging:

`connectonion/cli/templates/browser/agent.py` no longer exists — #279 and #294 removed it ("Templates drive the browser through `co browser`, not in-process"; "ship one template"). This PR's edit to that file was updating its prompt for the new tool, so it is moot rather than lost. Everything else applied unchanged.

```
27 passed   (screenshot / read_image / browser_agent)
```

## Not merging without a word, because it changes a public tool

`BrowserAutomation.take_screenshot()` returns **a path** now instead of a base64 data URL. Anything in user code that treated the return value as an inline image breaks — it is a better shape (a data URL for a full-page screenshot is enormous, and `read_image` makes reading it explicit), but it is a signature change to a tool people call directly, landing in a release meant to be sat on for a long time.

Two ways to read that:

- **Land it in 1.6.0**, because a long-term release is exactly when you want the better shape frozen, and the migration is `read_image(take_screenshot())`.
- **Hold it for 1.7**, and keep 1.6.0 to fixes.

Flagging rather than choosing. The rebase is done either way, so it is ready whenever.

wu-changxing:
Closing out of the 1.6.0 release queue. This is a clean implementation, but it deliberately changes the public return contract of BrowserAutomation.take_screenshot() from a data URL to a path. That is a feature/API migration, not a release blocker, and should not ride into the long-term release while we are stabilizing security and email. The rebased branch and tests remain recoverable for a focused 1.7 PR with migration notes.

Issue #234, opened 2026-07-22
All templates should drive the browser via the `co browser` CLI (prompt + shell tool), not the in-process BrowserAutomation tool


wu-changxing:
## Idea

Standardize every template on **one** way to use the browser: the `co browser`
CLI (backed by the persistent daemon from #157), driven by a **shell tool + a
browser section in the system prompt** — instead of wiring the in-process
`BrowserAutomation` object in as a set of tools.

So a template's agent does:

```
bash("co browser go_to https://example.com")
bash("co browser take_screenshot")
```

…and the system prompt has a short "Using the browser" section teaching the
`co browser` verbs. No `BrowserAutomation` in the tool list, no browser-specific
plugins (`bind_browser_session`, the idle reaper, etc.).

## Why this is better

- **One command surface.** The agent and a remote operator use the *same*
  `co browser` verbs. Remote control becomes `co call <addr> co browser take_screenshot`
  (via PR #225) — no separate code path, no separate whitelist.
- **Simpler templates.** `tools=[bash, ...]` + a prompt section, versus wiring
  `BrowserAutomation` + `bind_browser_session` + idle reaper + `image_result_formatter`.
  The browser integration lives in one place (the daemon), not re-plumbed per template.
- **No two-browsers-one-profile conflict.** The daemon is the sole owner of
  `~/.co/browser_profile` (see #157). Templates that also instantiate an
  in-process `BrowserAutomation` would collide; going CLI-only removes that whole
  class of problem.
- **Prompt-level, not code-level.** Adding browser ability to a new template is
  editing a prompt, not writing tool-wiring.

## Prerequisites found while reviewing PR #225

Two **code-level** constraints block a pure "prompt + shell tool" approach today.
They're small but real — the browser-via-CLI path does not work without them:

1. **`bash`'s `description` is required.** `def bash(command, description, ...)`
   → `bash(command="co browser take_screenshot")` raises `TypeError: missing
   'description'`. `remote.call("bash", command=...)` and `co call ... co browser`
   both hit this. **Fix:** make `description` optional (`description: str = ""`)
   in `useful_tools/bash.py`. One line, unblocks everything.

2. **The whitelist `Bash(...)` pattern only matches a tool literally named
   `bash`** (`tool_approval/approval.py:366` hardcodes `tool_name.lower() != 'bash'`).
   So the template's shell tool must be the **`bash`** tool (not `Shell.run`,
   which registers as `run`), or the whitelist needs a non-bash shell rule. Using
[...]

PR #273, opened 2026-07-27
co ai drives the browser through `co browser`, not 40 in-process tools

### LINKED ISSUE #270: co ai should drive the browser via `co browser`, not 40 in-process tool schemas
Sibling of #234, which does this for the `co create` templates. `co ai` has the same problem and is the more visible case.

## Today

`cli/co_ai/agent.py` wires the browser in as code:

```python
browser = BrowserAutomation(headless=False)
tools = [file_tools, browser, bash, ...]
plugins = [..., image_result_formatter, bind_browser_session]
```

That one object costs, measured on the current build:

- **40 tool schemas** sent to the model on **every** request
- **10,273 characters ≈ 2,570 tokens** of schema, before a single message
- a **60-line `bind_browser_session` plugin** that exists only because `co ai` is served via `host()` and every chat panel would otherwise navigate the one shared page
- `image_result_formatter` in the plugin list to get screenshots back

Meanwhile `co browser` already does all of this, from one persistent daemon that owns `~/.co/browser_profile`, with the same verbs a human types.

## Proposed

Drop `BrowserAutomation` from `co ai`'s tool list. Teach the browser in the **prompt** instead, and let the existing `bash` tool run it:

```
bash("co browser go_to https://example.com")
bash("co browser take_screenshot")
bash("co browser click_element_by_selector 'a' --text='Sign in'")
```

Add a "Using the browser" section to the `co ai` prompt (`cli/co_ai/prompts/`) covering the verbs, the `-t` tab targeting, and the exit-code contract.

## Why this is the right trade

**2,570 tokens back on every turn.** That's the part that compounds — it is paid on each request whether or not the task touches a browser, which for a coding agent is most of the time.

**One surface instead of two.** Right now `co ai` drives the browser through in-process Python while a human drives it through the CLI, so behavior can diverge and bugs must be fixed twice. After this there is one path, and remote control (`co call <addr> co browser take_screenshot`) is the same path again.

**Two plugins stop being necessary.** `bind_browser_session` exists purely to work around a shared in-process browser across panels; the daemon already handles tabs, so tab routing becomes `co browser -t <tab>` in the prompt. `image_result_formatter` is only needed because screenshots come back as in-process tool results.

**Adding browser ability becomes a prompt edit, not tool wiring.** Same argument as #234.

## Blockers — check these first

#234 found two code-level constraints. Both are now resolved, but confirm before starting:

1. `bash`'s `description` had to become optional, or `bash(command="co browser ...")` raises `TypeError`. Merged (#240-era work) — verify on current `main`.
2. The whitelist `Bash(...)` pattern only matches a tool literally named `bash` (`tool_approval/approval.py` hardcodes `tool_name.lower() != 'bash'`). `co ai` already passes the real `bash` tool, so this should be satisfied — confirm.

A third one specific to `co ai`: **`co ai` runs headful today** (`BrowserAutomation(headless=False)`). The daemon's headful/headless behaviour needs to match, or `co ai` users lose the visible browser they currently get.

## Scope

- `cli/co_ai/agent.py` — remove `BrowserAutomation`, `bind_browser_session`, and likely `image_result_formatter`
- `cli/co_ai/prompts/` — add the browser section
- keep `bash` in the tool list (it's already there)
- check `co ai`'s tool-approval whitelist covers `Bash(co browser *)` so the agent isn't prompted on every browser verb

## Don't forget

`co browser`'s screenshots return a **file path**, not an image. Whatever the prompt says has to make the agent read that file when it needs to look at the result — that is the one ergonomic thing `image_result_formatter` was buying, and losing it silently would make the agent blind to its own screenshots.

## Related

- #234 — same change for the `co create` templates
- #245 — `co browser` daemon socket bug; worth fixing before more callers depend on the daemon

wu-changxing:
Closes #270. Sibling of #234, which does the same for the `co create` templates.

## The cost that was being paid every turn

`BrowserAutomation` in `co ai`'s tool list meant **40 tool schemas — 10,273 characters, ~2,568 tokens** — in every request, whether or not the task went near a browser. For a coding agent that is most of the time.

The prompt section replacing it is **431 tokens**. Net: **~2,137 tokens back per request**.

## What changed

```python
-browser = BrowserAutomation(headless=False)
-tools = [file_tools, browser, bash, ...]
-plugins = [..., image_result_formatter, bind_browser_session]
+tools = [file_tools, bash, ...]
+plugins = [..., image_result_formatter]
```

The agent now runs `bash("co browser go_to https://example.com")` against the same daemon a human drives. One surface instead of two implementations that can drift, and remote control (`co call <addr> co browser ...`) is that same path again.

New `prompts/browser.md` teaches the verbs, `-t` tab targeting, `co browser status` before assuming a page is open, and the exit-code contract. It is loaded by name in the assembler — nothing in the tool list would pull it in now that the browser isn't a tool.

`prompts/main.md`'s **Remote Login Requests** section was still naming `open_browser` / `go_to` / `take_screenshot` as tools; rewritten to the CLI, keeping the `ask_user` handoff for QR codes, passwords and OTP exactly as it was.

## bind_browser_session is gone

That 60-line plugin existed for one reason: hosted `co ai` runs every chat panel's turns on one shared in-process browser, so without it concurrent panels navigated over each other's page. The daemon owns tabs itself (`-t <tab>`), so the workaround has nothing left to work around. The plugin still ships for anyone wiring `BrowserAutomation` directly.

## The part that made this more than a deletion

`co browser take_screenshot` deliberately returns a **path**, not base64 — the daemon strips the payload so CLI output stays readable, and there is a test pinning that (`test_dispatch_image_payload_shows_path_not_base64`).

`image_result_formatter` only recognised base64 and data URLs. Swapping the tools out without touching it would have meant **the agent could no longer see its own screenshots, and the user would see nothing** — silently, with the login flow ("once the QR code is visible, take a screenshot") quietly broken.

So the formatter now also reads the image a path points at:

```
Screenshot saved to: .tmp/screenshots/step_3.png   →  file read, uploaded, attached
```

Base64 detection runs **first**, so in-process tools behave exactly as before. A path to a file that doesn't exist, and ordinary text mentioning a `.md` or `.pdf`, are not images.

## Verified, not assumed

- `bash(command, description="")` — description is optional on `main`, so `bash("co browser ...")` doesn't `TypeError` (blocker 1 from #234)
- the whitelist's `Bash(...)` matching needs a tool literally named `bash`; `co ai` passes exactly that (blocker 2)
- `host.yaml` already whitelists `Bash(co *)`, so no approval prompt on every verb
- the daemon defaults to `headless=False`, matching the headful browser `co ai` users have today

## Tests

Three existing tests asserted the old contract and were updated to assert the new one rather than deleted — no in-process browser, `bash` present, `bind_browser_session` **not** registered, and the login prompt teaching `co browser`.

Five new tests on path detection, including that a nonexistent path isn't an image, that ordinary output is untouched, and that base64 still wins.

Full suite: **2,611 passed, 6 skipped**.

## Note on the numbers

While measuring for #270 I briefly got 609 tokens instead of 2,568 — that run stringified the function objects instead of calling `to_function_schema()`. The 2,568 figure in the issue is the correct one; both were re-derived from the real schemas for this PR.


wu-changxing:
## Self-review found a real defect — fixed in `1013c6d`

The first version scanned tool output for **any** existing image path. Ordinary shell output became a screenshot whenever it happened to name a real `.png`:

```
bash("ls")     →  logo.png\nREADME.md      →  treated as a screenshot
git status     →  modified:   logo.png      →  treated as a screenshot
```

Consequence: unrelated user files uploaded to the backend, and the tool result replaced with an image placeholder — quietly corrupting output the agent then reasons over. Verified by reproduction before fixing.

Detection is now anchored to the daemon's actual contract over the whole string:

```python
r'^\s*(?:screenshot\s+)?saved to:\s*([^\n]+?\.(?:png|jpe?g|gif|webp))\s*$'
```

Plus a 20 MB ceiling — a screenshot is far under it, and base64-ing anything larger would balloon the request. Three false-positive shapes (`ls`, `git status`, `grep`) and the size guard are pinned by tests.

## Considered and rejected: let `read_file` fetch the image instead

Raised during review, and it is the more principled shape — reading is an explicit action, no detection heuristics, no false-positive class at all. `useful_tools/read_file.py` already returns `data:image/...;base64` for images specifically so this plugin picks them up.

Two things rule it out here:

1. **`co ai` doesn't use that tool.** It passes `FileTools`, whose `read_file` does `read_text(encoding="utf-8", errors="replace")` — on a PNG that is mojibake. Getting the read route working means swapping `co ai`'s file-reading tool wholesale, which also changes PDF, audio and video handling. That deserves its own PR, not a rider on this one.

2. **It would break a contract users depend on.** `main.md` promises the QR-code screenshot is *"shown to the user automatically"*. Auto-attach serves the **user**; the read route only fires when the **agent** decides to look. Scan-to-login would become dependent on the model choosing to read a file.

Worth revisiting as a separate change — swapping `FileTools`' reader for the richer `read_file` has value beyond screenshots.

Full suite after the fix: **2,613 passed, 6 skipped**.

PR #274, opened 2026-07-27
Fix browser daemon busy errors under concurrent load

wu-changxing:
## What changed

- raise the POSIX Unix-socket listen backlog from 8 to the platform-supported maximum
- retry generic POSIX socket errors when the recorded daemon owner is still alive
- add regression coverage for 32 clients queued behind a long-running command and macOS-style generic `OSError` behavior

## Why

Browser commands execute serially. A long command could fill the eight-entry accept queue, causing concurrent callers to report `browser daemon is busy` even though the daemon was healthy. On macOS, a full queue may also surface as a generic `OSError`; the client previously treated that as a stale socket and unlinked the active daemon endpoint.

## Impact

Concurrent browser commands now queue behind the active operation instead of incorrectly treating a healthy daemon as unavailable or stale.

## Validation

- `python -m pytest tests/e2e/cli/test_browser_daemon.py -q` — 71 passed
- `python -m py_compile connectonion/cli/browser_agent/client.py connectonion/cli/browser_agent/daemon.py tests/e2e/cli/test_browser_daemon.py`
- `git diff --check`

### DIFF
--- a/connectonion/cli/browser_agent/client.py
+++ b/connectonion/cli/browser_agent/client.py
@@ -48,6 +48,13 @@ def _connect_posix(sock_path: str):
             time.sleep(0.1)
         except OSError:
             conn.close()
+            # macOS may report a full AF_UNIX accept queue as a generic OSError
+            # rather than ConnectionRefusedError. The socket belongs to a healthy
+            # daemon in that case; unlinking it would strand that daemon and let a
+            # second one start against the same browser profile.
+            if _owner_alive(sock_path):
+                time.sleep(0.1)
+                continue
             if os.path.exists(sock_path):
                 os.unlink(sock_path)  # stale socket: nothing is listening on it
             return None
--- a/connectonion/cli/browser_agent/daemon.py
+++ b/connectonion/cli/browser_agent/daemon.py
@@ -533,7 +533,11 @@ class BrowserDaemon:
         self._srv = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         self._srv.bind(self.sock_path)
         Path(transport.pid_path(self.sock_path)).write_text(str(os.getpid()), encoding="utf-8")
-        self._srv.listen(8)
+        # Commands execute serially, so a long browser action can leave many
+        # concurrent CLI callers waiting to be accepted. Use the platform's
+        # supported maximum queue instead of making the ninth caller look as if
+        # the healthy daemon disappeared.
+        self._srv.listen(socket.SOMAXCONN)
 
     def _bind_windows(self):
         """Named-pipe bind. A pipe vanishes WITH its owning process, so 'no pipe' is
--- a/tests/e2e/cli/test_browser_daemon.py
+++ b/tests/e2e/cli/test_browser_daemon.py
@@ -843,6 +843,34 @@ def test_bind_replaces_a_stale_socket_without_pidfile(short_sock):
     daemon._srv.close()
 
 
+@posix_only
+def test_connect_retries_generic_oserror_while_daemon_owner_is_alive(
+        short_sock, monkeypatch):
+    """macOS can surface a full accept queue as generic OSError. A live owner
+    makes that socket busy, not stale: retry it and never unlink its endpoint."""
+    Path(short_sock).touch()
+    attempts = []
+
+    class FakeSocket:
+        def connect(self, path):
+            attempts.append(path)
+            if len(attempts) == 1:
+                raise OSError("accept queue full")
+
+        def close(self):
+            pass
+
+    monkeypatch.setattr(c.socket, "socket", lambda *args: FakeSocket())
+    monkeypatch.setattr(c, "_owner_alive", lambda path: True)
+    monkeypatch.setattr(c.time, "sleep", lambda seconds: None)
+
+    conn = c._connect_posix(short_sock)
+
+    assert isinstance(conn, FakeSocket)
+    assert attempts == [short_sock, short_sock]
+    assert os.path.exists(short_sock)
+
+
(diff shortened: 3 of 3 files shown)


wu-changxing:
## Overview

Two small, focused changes: raise the POSIX listen backlog from `8` to `socket.SOMAXCONN`, and treat a generic `OSError` from `_connect_posix` the same way `ConnectionRefusedError` is already treated — retry while the recorded daemon owner is alive, fail fast (and unlink) when it isn't. Both changes are simple and well-commented; the core bug (macOS surfacing a full backlog as a generic `OSError`, causing the client to unlink a *healthy* daemon's socket) is real and this does fix it.

## Issues

**Retry bound — looks safe, but flag the asymmetry it inherits.** The new `OSError` branch shares the same `for attempt in range(20)` loop as `ConnectionRefusedError`, so it's bounded to ~2s total, not infinite, and it does fail fast (unlink + `return None`) when `_owner_alive()` is false — good, matches the fail-fast convention. However, note the pre-existing inconsistency this change highlights: on a *dead* owner, the `ConnectionRefusedError` branch returns `None` **without** unlinking the stale socket, while the `OSError` branch (new logic) **does** unlink it. Both branches reach the same conclusion ("stale, nobody home") but clean up differently. Worth a one-line comment explaining why, or unifying the two paths, so a future reader doesn't have to reverse-engineer the intent.

**Does the 32-client test actually exercise the backlog fix, or just the retry loop that already existed?** `test_32_clients_queue_behind_a_long_running_command` blocks the daemon for only ~0.25s (`time.sleep(0.25)` before `release.set()`) before releasing it. The client already had a ~2s ECONNREFUSED-retry budget *before* this PR (`_connect_posix`'s existing 20-attempt loop). Given AF_UNIX `connect()` refuses immediately when the backlog is full (not by blocking), the ~24 clients that would overflow an 8-entry backlog should very plausibly succeed on a later retry within that pre-existing 2s window even without raising `SOMAXCONN`, since the busy period here is far shorter than the retry budget. As written this test doesn't clearly prove causality — it would be worth confirming it actually fails against `main` (revert just the `listen(8)` → `listen(SOMAXCONN)` line) before relying on it as the regression guard, or lengthening the busy window well past the retry budget so backlog headroom is the only thing that can save it.

**Test coverage gap on the OSError branch.** `test_connect_retries_generic_oserror_while_daemon_owner_is_alive` only covers the "owner alive → retry" sub-path. There's no test for the "owner is dead → unlink + return `None`" sub-path of the *new* `OSError` branch specifically (as distinct from the existing `ConnectionRefusedError`-dead-owner test coverage, if any) — i.e., nothing pins down that a genuinely stale socket surfaced as a generic `OSError` still gets cleaned up and doesn't hang.

**Minor:** the bare `except OSError` also swallows things like `PermissionError` on the socket file, which would now retry for ~2s against a live-but-permission-denied owner before giving up with the same generic failure as before — harmless given the bound, but worth a mental note that "generic OSError" is being read as "backlog full" fairly broadly.

## Done well

- The fix directly targets the real failure mode described in the PR body (unlinking a healthy daemon's socket), and does so by extending an already-established pattern (`ConnectionRefusedError` handling) rather than inventing new logic — good for consistency and reviewability.
- `socket.SOMAXCONN` is the right general-purpose fix over a hardcoded bump (e.g. `128`) — no magic number, self-adjusts to the platform.
- Comments are clear about *why* (macOS quirk, serial command execution causing queue buildup), which will save the next reader real time.
- Functions stay small and single-purpose (`_connect_posix`, `_bind_posix` both comfortably under the 30-line guideline); no `try/except/pass`; failure paths still fail fast and loud per `CLAUDE.md`.

---

wu-changxing:
**Deferred to 1.6.0.**

wu-changxing:
Checked both halves of this against `main`. **One is already fixed; the other I could not reproduce.**

## The unlink hazard is fixed, by a different route

This PR's sharpest point was that a generic `OSError` on macOS made the client treat a healthy daemon as stale and unlink its socket. `client.py` on main now says exactly that, and does not:

> "nobody is listening". Any other OSError — EACCES from a process … is alive, and unlinking on those took a live, idle daemon permanently offline: it kept accept()ing on an unlinked inode

So only a truly stale socket is unlinked. That half no longer needs this PR.

## The backlog is still 8, and I could not make it bite

`daemon.py:679` is still `self._srv.listen(8)`.

Measured on macOS, with the daemon up and a page loaded, firing **16 concurrent** `co browser get_current_url`:

```
exit codes: 16 × 0
'busy' / 'stale' / 'refused' errors: none
```

All sixteen queued and completed. Either the kernel is more forgiving about Unix-domain backlog than the number suggests, or these commands finish fast enough that nine were never outstanding at once.

## What that means for this PR

I am not changing `listen(8)` on a symptom I cannot produce — a platform-dependent tuning change without a failing case is the kind of edit this repo has been bitten by before.

To justify it, the reproduction needs a command that holds the daemon long enough to stack the queue — a slow page load or a deliberate sleep on the daemon side — or a run on Linux, where the queue behaviour differs. If you have the original failing case from when this was filed, that would settle it faster than my guessing at one.

The regression tests here (32 clients behind a long-running command) look like exactly that reproduction, which is an argument for the PR rather than against it — but the branch conflicts with `main` now, and the half that mattered most has already landed.

wu-changxing:
Closing during the 1.6.0 cleanup. The live-daemon unlink hazard this addressed is already fixed on main by a narrower path; the remaining listen-backlog change is conflicting and still lacks a reproduction that fails on current main. The branch remains recoverable. Reopen as a fresh focused PR only if a current Linux/macOS failing case proves the backlog change is still needed.

PR #275, opened 2026-07-27
Add co syno: your Synology NAS from the terminal

### LINKED ISSUE #257: co syno: Synology File Station CLI
We have `co gdrive`, `co email`, and `co outlook`. A Synology NAS is the one storage surface still stuck behind the DSM web UI — proposing `co syno` to make it usable from the terminal.

## Scope (v1)

**File Station only.** Photos, Download Station, and Chat are out of scope until File Station is shipped and used.

```
co syno login              # QuickConnect ID + username + password prompt
co syno                    # list recent files
co syno ls /home/photos
co syno search <query>
co syno get <#>
co syno put <path>
co syno rm <#>
co syno share <#>
```

This deliberately mirrors `co gdrive` one-to-one — same verbs, same numbered-listing UX, so there is nothing new to learn.

## Design

**`connectonion/useful_tools/synology.py`** — a `Synology` class shaped like `GDrive`, so it also works as an agent tool via `Agent(tools=[Synology()])`:

- `_resolve()` — QuickConnect ID to an ordered list of endpoint candidates (LAN, then DDNS, then relay)
- `login()` and session `sid` reuse
- `list_files()`, `search_files()`, `download()`, `upload()`, `delete()`, `share()`

**`connectonion/cli/commands/synology_commands.py`** — CLI surface mirroring `gdrive_commands.py`, including the `~/.co/syno_last_list.json` numbering cache so `co syno get 3` refers to row 3 of the last listing.

## Two things that differ from gdrive

**1. No OAuth, so no `co auth synology`.** gdrive/gmail/outlook all broker tokens through oo-api. A NAS has nothing to broker — DSM auth is `SYNO.API.Auth` with username/password returning a session `sid`. Credentials stay local in `~/.co/keys.env` and never touch oo-api.

**2. QuickConnect resolution.** Most people never expose DSM directly; they reach it via `quickconnect.to/<ID>`. So there is a resolution step before any API call. Verified this endpoint is live and behaves as expected:

```
POST https://global.quickconnect.to/Serv.php
{"version":1,"command":"get_server_info","id":"dsm_portal_https","serverID":"<ID>", ...}
```

An unknown alias returns `{"errno":4,"errinfo":"get_server_info.go:92[Alias not found]"}`. Resolution is unauthenticated, and the resolved URL gets cached as `SYNOLOGY_URL` so it is paid once rather than per command.

## Decisions to confirm

- `rm` moves to `#recycle` rather than hard-deleting, matching how `co gdrive rm` trashes instead of destroying.
- Login prompts for an OTP only when DSM asks for one, rather than always.

## Blocked on

Needs a real QuickConnect ID and DSM account to test against before any code is written. The resolver response shape (which candidates come back, in what order) determines the connection fallback logic, and that cannot be guessed from docs.

## Prerequisite

`main.py` and `docs/cli/README.md` are currently mid-merge from the gdrive branch. That merge should land before a new command group is stacked onto `main.py`.

wu-changxing:
Closes #257.

`co gdrive`, `co gmail`, and `co outlook` cover the cloud accounts. A Synology NAS was the one storage surface still stuck behind the DSM web UI. This adds `co syno`.

```bash
co syno login              # QuickConnect ID, or --url
co syno                    # shared folders
co syno ls /home/photos
co syno search invoice --in /home
co syno get 3 --to ~/Downloads
co syno put report.pdf /home/docs
co syno share 3
```

Deliberately the same command shape and numbered-listing UX as `co gdrive`, so there's nothing new to learn.

## Why no new dependency

File Station is [officially documented](https://global.download.synology.com/download/Document/Software/DeveloperGuide/Package/FileStation/All/enu/Synology_File_Station_API_Guide.pdf) plain HTTP with JSON responses — no library needed, and `httpx` is already a dependency. The undocumented half is QuickConnect, and the best-known Python client ([synology-api](https://github.com/N4S4/synology-api)) resolves it **straight to the relay with no LAN fallback**, so depending on it would mean adding weight to every ConnectOnion install *and* still writing the connection ladder. Its value was protocol knowledge, not code.

## Three things worth reviewing

**QuickConnect is not one address.** The NAS may be on this LAN, at a DDNS name, or only reachable through Synology's relay — and Synology throttles the relay hard. `login` probes LAN first and caches whichever answered, so you get gigabit at home without losing access from a café. `--url` skips QuickConnect entirely, which also means the undocumented protocol can never become a hard dependency: if Synology changes the handshake, `co syno` degrades to "type your address" instead of dying.

**API paths and versions are discovered, not hardcoded.** `SYNO.API.Info` is queried once per session because DSM moves CGI paths and bumps max versions between releases. This is the detail that usually rots these integrations.

**There is no `co syno rm`.** `SYNO.FileStation.Delete` deletes permanently — the word "recycle" does not appear anywhere in the 112-page spec, because DSM's recycle bin is a share-level filesystem convention the web UI implements, not an API affordance. Unlike `co gdrive rm`, it could not be made recoverable, so rather than ship a delete that's quietly more destructive than its gdrive counterpart, it's left out. This reverses what the issue originally proposed.

## Other decisions

- Credentials live in `~/.co/keys.env` and never touch oo-api — a NAS has no OAuth to broker, hence no `co auth synology`.
- Passwords go in the POST body. Every example in the official guide passes `passwd=` in a GET query string, where it lands in proxy and server logs.
- Sessions expire after 7 days and die on duplicate login; DSM codes 106/107/119 trigger one silent re-auth so agents never see a session error.
- Upload sends the binary part last, per the RFC 1867 constraint the spec calls out, and always sets `overwrite` explicitly (DSM returns error 1805 rather than defaulting).

## Testing

21 new unit tests covering candidate ordering, the fallback ladder, credential handling, stale-session retry, response normalization, and the upload constraints.

```
tests/unit/test_synology.py .....................  21 passed
tests/unit (full)                                  2295 passed, 3 skipped
tests/e2e/cli                                      332 passed, 3 skipped
```

## ⚠️ Not yet run against real hardware

**This has not touched a physical NAS.** Every API shape is taken from the official spec and the tests are mocked. What I verified live is the QuickConnect resolution endpoint — it responds and returns `{"errno":4,"errinfo":"...[Alias not found]"}` for an unknown alias.

The piece most likely to need adjustment is `resolve_quickconnect()`: which candidate fields a real `get_server_info` returns, and in what shape, is the one thing that can't be settled from documentation. **Please don't merge before someone runs `co syno login` against an actual NAS.**

One more known tradeoff: TLS certificate verification is disabled for NAS connections, because DSM ships a self-signed certificate by default. Traffic is still HTTPS. Worth a second opinion on whether that should be opt-in instead.

### DIFF
--- a/connectonion/__init__.py
+++ b/connectonion/__init__.py
@@ -51,7 +51,7 @@ from .prompts import load_system_prompt
 from .debug import xray, auto_debug_exception, replay, xray_replay
 from .useful_tools import (
     send_email, get_emails, mark_read, mark_unread,
-    Memory, Gmail, GDrive, GoogleCalendar, Outlook, MicrosoftCalendar,
+    Memory, Gmail, GDrive, Synology, GoogleCalendar, Outlook, MicrosoftCalendar,
     WebFetch, Shell, bash, codex, DiffWriter, MODE_NORMAL, MODE_AUTO, MODE_PLAN,
     pick, yes_no, autocomplete, TodoList, SlashCommand,
     # Claude Code-style file tools
@@ -83,6 +83,7 @@ __all__ = [
     "Memory",
     "Gmail",
     "GDrive",
+    "Synology",
     "GoogleCalendar",
     "Outlook",
     "MicrosoftCalendar",
--- a/connectonion/cli/commands/synology_commands.py
+++ b/connectonion/cli/commands/synology_commands.py
@@ -1,0 +1,190 @@
+"""
+Purpose: CLI surface for the user's Synology NAS — log in, browse, search, download, upload, and share File Station files from the terminal
+LLM-Note:
+  Dependencies: imports from [json, os, pathlib, typer, questionary, dotenv, rich.console, rich.table, ...useful_tools.synology] | imported by [cli/main.py via handle_syno_*()] | hits the NAS through the Synology tool
+  Data flow: _syno() loads SYNOLOGY_* from .env / ~/.co/keys.env → Synology() instance | ls/search: list_files()/search_files() → numbered Rich table (tab-separated with full paths when piped) → saves {#: path} to ~/.co/syno_last_list.json | get/share: resolve short numbers via that cache → download()/share()
+  State/Effects: login writes SYNOLOGY_URL/ACCOUNT/PASSWORD/SID to ~/.co/keys.env | writes ~/.co/syno_last_list.json (last listing's # → path map) | get downloads local files | put uploads to the NAS
+  Integration: exposes handle_syno_login(), handle_syno_list(), handle_syno_search(), handle_syno_get(), handle_syno_put(), handle_syno_share() for cli/main.py | presentation mirrors gdrive_commands.py | NAS logic lives in useful_tools/synology.py
+  Errors: guarded failures print a hint and exit 1 (typer.Exit) — not logged in, unresolvable file #, missing upload path | ValueError from the Synology tool carries decoded DSM error text
+"""
+
+import json
+import os
+from pathlib import Path
+
+import typer
+from rich.console import Console
+from rich.table import Table
+
+console = Console()
+
+LIST_CACHE = Path.home() / ".co" / "syno_last_list.json"
+
+
+def _syno():
+    """Load SYNOLOGY_* credentials from .env files and return a Synology instance. Exits 1 with a hint if not connected."""
+    from dotenv import load_dotenv
+
+    for env_path in [Path(".env"), Path.home() / ".co" / "keys.env"]:
+        if env_path.exists():
+            load_dotenv(env_path)
+
+    if not os.getenv("SYNOLOGY_URL"):
+        console.print("\n❌ [bold red]Synology NAS not connected[/bold red]")
+        console.print("\n[cyan]Connect your NAS first:[/cyan]")
+        console.print("  [bold]co syno login[/bold]     Log in with your QuickConnect ID\n")
+        raise typer.Exit(1)
+
+    from ...useful_tools.synology import Synology
+    return Synology()
+
+
+def _size(count: int) -> str:
+    """Render a byte count as B/KB/MB/GB; '-' for folders."""
+    if not count:
+        return "-"
+    size = float(count)
+    for unit in ("B", "KB", "MB", "GB"):
+        if size < 1024 or unit == "GB":
+            return f"{size:.0f}{unit}" if unit == "B" else f"{size:.1f}{unit}"
+        size /= 1024
+
+
+def _when(timestamp: int) -> str:
+    """Render a File Station mtime (unix seconds) as 'Jul 26 14:30' in local time."""
+    from datetime import datetime
+
+    if not timestamp:
+        return ""
+    return datetime.fromtimestamp(timestamp).strftime("%b %d %H:%M")
+
+
+def _print_listing(files: list, title: str):
+    """Render files as a numbered table (or tab-separated rows with full paths when piped) and cache the numbering."""
+    LIST_CACHE.parent.mkdir(parents=True, exist_ok=True)
+    LIST_CACHE.write_text(json.dumps({str(i): f["path"] for i, f in enumerate(files, 1)}), encoding="utf-8")
+
+    if not console.is_terminal:
+        # Scripts and agents get full paths, never a truncated column. Plain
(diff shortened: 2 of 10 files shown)


wu-changxing:
## Overview

Reviewed the diff directly (`connectonion/useful_tools/synology.py`, `synology_commands.py`, `cli/main.py` wiring, `tests/unit/test_synology.py`). The shape is solid: discovered API paths via `SYNO.API.Info` instead of hardcoding `entry.cgi`, a LAN→DDNS→external→relay candidate ladder for QuickConnect, POST-body credentials, and silent re-auth on stale-session codes (106/107/119). 21 mocked unit tests back the claims in the description reasonably well.

## Issues

- **`download()` builds JSON by hand, not `json.dumps`**: `path = f'["{path}"]'` (synology.py). If a NAS path ever contains a `"`, this breaks the intended single-element array and can inject extra array elements into the `SYNO.FileStation.Download` request. Every other call site lets httpx encode params properly — this one spot doesn't. Trivial fix: `json.dumps([path])`. Low real-world severity (scoped to the authenticated user's own File Station access) but it's the one place raw string interpolation leaks into a structured payload, worth closing given `path` can originate from agent/LLM-driven tool calls, not just the human typing a CLI arg.
- **No visible feedback that cert verification is off.** `verify=False` is set silently in `_client()`/`pick_reachable()`; nothing prints a one-time notice to the user. Given the security tradeoff, at least surface it once at `co syno login` time.
- **Credentials are a raw DSM password in `~/.co/keys.env`**, not a token — weaker than the OAuth-based gdrive/gmail/outlook integrations, though understandable given File Station has no OAuth story. Worth a code comment acknowledging the asymmetry so it isn't mistaken for an oversight later.
- **`_login()` failures during `co syno login` aren't wrapped** — a wrong password raises `ValueError` straight out of `handle_syno_login`, while every other guarded failure in this file uses `typer.Exit(1)` with a formatted hint. If `main.py` doesn't have a blanket ValueError→friendly-message handler elsewhere, this one path will print a raw traceback instead of matching the polish of the rest of the command.
- Nothing to flag on the deliberate omission of `co syno rm` — the reasoning (no real recycle bin in the FileStation API, unlike gdrive) checks out against the spec description and isn't contradicted anywhere in the diff.

## TLS verification opinion

Disabling verification is reasonable for the **bare-IP candidates** (LAN interface IP, external IP, relay IP) — DSM's self-signed cert can't validate against an IP SAN anyway, so "verify" would just fail every time regardless of MITM risk. But `verify=False` is applied uniformly to *every* candidate in `_client()`, including the **hostname-based** ones: the DDNS name (`*.synology.me`) and the relay fallback (`{id}.quickconnect.to`). Synology issues real, CA-signed certs for both of those hostnames via its DDNS/QuickConnect service — turning off verification there throws away free, working security for zero benefit, and the relay path is exactly the leg that leaves the LAN and crosses the open internet where MITM actually matters.

My recommendation: don't make it a blanket flag. Attempt standard verified TLS for the two hostname candidates first; only fall back to `verify=False` for the numeric-IP candidates, and pin the leaf cert's fingerprint on first successful IP connection (store it next to `SYNOLOGY_URL` in `keys.env`, compare on reconnect). That gets you the zero-friction "it just works on my LAN" experience the PR wants, without silently disabling verification on the one connection path (relay) that's actually exposed to the internet. If TOFU is too much scope for v1, opt-in (default verify on, `--insecure` to disable) is still preferable to a silent global downgrade — but plain opt-in re-introduces the "self-signed cert breaks day one" problem the author is trying to avoid, so TOFU is the better tradeoff here, not just the safer one.

## Done well

- Candidate ordering test (`test_resolve_quickconnect_puts_lan_before_relay`) actually asserts LAN-first, not just "some list came back."
- Stale-session retry is tested for all three DSM codes and confirms exactly one re-login + one retry, not a loop.
- Upload test asserts `overwrite` is always explicit in the payload (never relying on DSM's default), matching the 1805-error rationale in the description.
- Path-per-API version/lookup (`_path_for`) is genuinely the right fix for the "DSM moves CGI paths" rot the description calls out — most integrations in the wild hardcode this and break silently on DSM updates.

Agreed with the author: this should not merge until `co syno login` has actually been run against a real NAS — `resolve_quickconnect()`'s field-shape assumptions (`server.interface[].ip`, `service.relay_ip` vs `relay_ip`) are the part documentation can't fully settle.

---

Issue #272, opened 2026-07-27
co ai: expose ULW as `--yolo` for autonomous skill runs


wu-changxing:
## Problem

ConnectOnion already has the `ulw` plugin for approval-free, multi-turn work, but `co ai` can only activate it through a frontend `mode_change` message. A terminal or scripted skill run has no supported way to opt into that behavior:

```bash
co ai "/deploy-oo-chat"
```

The name `ULW` is also internal vocabulary. For a CLI switch that deliberately skips approval prompts, `yolo` is clearer and consistent with other coding agents.

The current `connectonion:deploy-oo-chat` skill is Claude-oriented in its invocation and deployment assumptions, and does not document a `co ai` entry point.

## Proposed change

- Keep the existing `ulw` import, mode message, session keys, and checkpoint events as compatibility aliases.
- Add a public YOLO activation path backed by the existing ULW engine.
- Add `co ai --yolo` plus a bounded turn option for both one-shot prompts and hosted `co ai`.
- Make `co ai --yolo "/deploy-oo-chat"` discover and invoke the project skill from `.claude/skills/deploy-oo-chat/SKILL.md`.
- Update tests and CLI/plugin documentation.
- Update the deploy skill in `openonion/connectonion-claude-plugin` with portable frontmatter, both Claude Code and `co ai` invocation, and fail-closed deployment checks.
- Do not perform a production publish or deployment as part of automated tests.

## Acceptance criteria

- [x] `co ai --yolo "task"` enters approval-free autonomous mode before the first tool call.
- [x] `co ai --yolo "/deploy-oo-chat"` loads the project skill and runs with the configured bounded turn budget.
- [x] `--yolo-turns` rejects non-positive values.
- [x] Plain `co ai` behavior is unchanged.
- [x] Existing `ulw` imports and frontend wire messages continue to work.
- [x] Unit tests cover activation, compatibility, turn bounds, and CLI argument forwarding.
- [x] The linked deploy-skill PR passes skill validation and a no-deploy forward test.

## Canonical pull requests

- Runtime and CLI: [openonion/connectonion#280](https://github.com/openonion/connectonion/pull/280) — ready for review; all eight GitHub Actions jobs pass.
- Deploy skill: [openonion/connectonion-claude-plugin#1](https://github.com/openonion/connectonion-claude-plugin/pull/1) — ready for review; skill validation and checks pass.

[...]

PR #276, opened 2026-07-27
Add YOLO mode to co ai

wu-changxing:
Addresses #272

## Summary

- expose a bounded public YOLO mode through the yolo plugin and co ai --yolo/--yolo-turns while retaining ULW wire compatibility
- run autonomous continuations iteratively, preserve the terminal result, and honor both tool-call and text-only interrupts
- enable project slash skills in co ai and bypass interactive tool approval only while YOLO mode is active
- document the CLI/plugin behavior and retain legacy ULW events and state

## Testing

- python -m pytest tests/unit -q (2299 passed, 3 skipped)
- python -m pytest tests/e2e/cli/test_cli_help.py -q (23 passed)
- git diff --check
- independent forward review: merge-ready, no blockers

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,18 +151,20 @@ Inject logic at any point in the agent execution cycle:
 
 ```python
 from connectonion import Agent, after_tools, llm_do
-from connectonion.useful_plugins import re_act, eval, auto_compact, subagents, ulw
+from connectonion.useful_plugins import re_act, eval, auto_compact, subagents, yolo
 
 # Built-in plugins — same capabilities as Claude Code, open to any agent
 agent = Agent("researcher", tools=[search], plugins=[
     re_act,         # Reflect + plan after each tool call
     auto_compact,   # Auto-compress context at 90% capacity
     subagents,      # Spawn sub-agents with independent tools and prompts
-    ulw,            # Ultra Light Work — fully autonomous mode
+    yolo(turns=25), # Bounded approval-free autonomous mode
 ])
 ```
 
-These plugins mirror Claude Code's internal capabilities — `auto_compact`, `subagents`, `ulw` directly correspond to Claude Code's context compression, sub-agent spawning, and autonomous work mode. ConnectOnion makes these capabilities available to any agent you build.
+These plugins mirror coding-agent capabilities such as context compression,
+sub-agent spawning, and autonomous work. Existing `ulw` imports and frontend
+messages remain supported as compatibility aliases for `yolo`.
 
 Hooks: `after_user_input`, `before_iteration`, `before_llm`, `after_llm`, `before_tools`, `before_each_tool`, `after_each_tool`, `after_tools`, `on_error`, `after_iteration`, `on_stop_signal`, `on_complete`
 
@@ -759,7 +761,7 @@ Plugins are lists of lifecycle hooks that inject logic at any point in the agent
 - `re_act`: Reflect + plan after each tool call
 - `auto_compact`: Auto-compress context at 90% capacity
 - `subagents`: Spawn sub-agents with independent tools
-- `ulw`: Ultra Light Work — fully autonomous mode
+- `yolo`: Bounded approval-free autonomous mode (`ulw` remains compatible)
 
 ```python
 from connectonion.useful_plugins import re_act, subagents
@@ -915,7 +917,7 @@ ConnectOnion is a simple, elegant framework for production-ready AI agents. Phil
 | Approval System | Dangerous ops auto-trigger approval |
 | Skills System | Claude Code compatible, auto-discovery |
 | 12 Lifecycle Hooks | Inject logic at any point |
-| Plugin System | re_act, auto_compact, subagents, ulw |
+| Plugin System | re_act, auto_compact, subagents, yolo |
 | Multi-Agent Trust | Fast rules, zero token cost |
 
 ### Quick Start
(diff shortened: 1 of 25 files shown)


wu-changxing:
## Overview

This PR bounds and publicizes ULW as `yolo`: `yolo(turns=N)` auto-activates on the first `after_user_input` (vs. requiring a frontend `mode_change`), `co ai --yolo/--yolo-turns` wire it into the CLI, `Agent._pending_inputs`/`_queue_input`/`_run_input_turn` replace the old recursive `agent.input()` continuation with an iterative drain loop, and `tool_approval` gains `poll_interrupt_after_text`/`on_stop_signal` handling so a graceful interrupt is honored even on a text-only (no-tool-call) turn. The `ulw` wire protocol, session keys (`mode: 'ulw'`, `ulw_turns*`, `skip_tool_approval`), and checkpoint events are deliberately kept byte-for-byte compatible. I verified the diff against the description and it matches; the 2,299/23-pass claims are consistent with the added test files.

## Relationship to #280

Not a clean supersede either way. #280 is rebased onto the current `co browser`/#273 architecture, is non-draft, and ships a companion skills PR — it's the better base to land. But #280's diff never touches `connectonion/core/agent.py` or `tool_approval/approval.py`, so two things this PR adds appear to be **missing from #280**:
- The iterative continuation queue (`_pending_inputs`/`_queue_input`), which this PR's own test (`test_yolo_continuations_are_iterative_beyond_python_recursion_depth`, 400 turns) shows matters — #280 seems to keep the old recursive `agent.input()` continuation, which will eventually hit Python's recursion limit on a long bounded run.
- Text-only interrupt handling (`poll_interrupt_after_text` / honoring `stop_signal` in the no-tool-call branch of `_run_iteration_loop`). Without it, an INTERRUPT sent while the agent is mid-YOLO but between text-only turns has no boundary to be caught at.

Recommendation: port these two pieces into #280 (or confirm #273 already covers them) before closing this one out, rather than closing #276 outright.

## Issues

- **Checkpoint/interrupt exit is not sticky (both PRs, same shape).** `yolo()`'s `activate_yolo` fires on *every* `after_user_input`, and only checks `mode not in AUTONOMOUS_MODES`. When a checkpoint's `switch_mode` action or `stop_autonomous_mode` (on interrupt) sets `mode` back to `safe`, the very next user message re-triggers `activate_yolo` and puts the session straight back into approval-bypass mode — the documented "user can continue, switch mode, or stop" story doesn't actually let the user stay out of YOLO. `#280`'s `activate_configured_yolo` has the identical `mode == 'ulw'` guard, so this isn't something the other PR already fixes. Worth a regression test for "exit then send another message."
- **Hosted `co ai --yolo` (no prompt) applies to every session on the shared server**, not just one call — the activation hook is baked into the single long-lived `Agent`, so every connecting user's first message silently bypasses tool approval, compounded by the point above (no durable way out). Given the "trust earned through behavior" philosophy, this probably wants to be scoped per-session/per-request rather than process-wide, or at least called out loudly in `docs/cli/ai.md` beyond "Start web chat with YOLO enabled."
- **`current_session['mode']` is never actually `'yolo'`.** `handle_yolo_mode_change` always writes `mode='ulw'` (by design, via `_activate_autonomous_mode(..., 'ulw', label='yolo')`), so the `'yolo'` literal in `AUTONOMOUS_MODES` and the `elif new_mode in {'yolo','ulw'}` branches is effectively dead for anything produced by this codebase's own handlers — only reachable if some other caller sets `mode` to the literal string directly. Not a bug, but a bit of an unused seam; a comment noting that would save the next reader some digging.
- **Minor:** the PR description credits itself with "enable project slash skills in co ai," but `skills_plugin` was already in `create_coding_agent`'s plugin list pre-PR — only new tests demonstrate it composing with `yolo`. Worth correcting the description.
- **Minor (CLAUDE.md: functions under 30 lines):** `Agent._run_input_turn` is ~50 lines. Understandable since it's hoisted out of the old `input()` body rather than newly written, but a follow-up split (e.g., separate the trace/logging tail from the turn-execution head) would fit the repo convention better.

## Done well

- Very thorough tests: recursion-depth avoidance is directly exercised (400-turn run), both interrupt paths (tool-call and text-only) are tested end-to-end through a real `Agent`+`MockLLM`, turn-budget validation rejects `bool`/float/non-positive/string inputs, and CLI flag forwarding/help text is covered.
- Backward compatibility is careful and explicit — wire protocol, persisted session keys, and checkpoint event shapes are untouched, with tests asserting the legacy `ulw` mode-change path still works.
- `_validate_turns` and the `_activate_autonomous_mode`/`_exit_autonomous_mode` shared helpers keep `yolo`/`ulw` in sync without duplicating logic, which is a clean way to avoid the old copy-pasted `ulw_keep_working` checkpoint code drifting between two mode names.

---

wu-changxing:
**Superseded — the work is in, via a different branch.**

#280 shipped the smaller version of this YOLO feature and merged into 1.5.0. This branch carried real work beyond it, so rather than close and lose it I rebased the remainder onto #280's implementation: `Agent._pending_inputs`, the tool_approval changes, and 96 tests.

The salvage found a bug neither PR had working on its own: ulw's autonomous mode called `agent.input()` **recursively**, which returned the *first* turn's result to the caller (a 3-turn hosted YOLO session answered with turn 1), reversed log ordering, and would hit Python's recursion limit. This PR added the queue that fixes it; #280 kept the recursive call. The fix sat unused in the tree until the two halves were joined.

That salvage branch is **deferred to 1.6.0** with the other features — 1.5.0 is scoped to the security fixes plus dashboard.

Closing this one; the branch supersedes it.

PR #281, opened 2026-07-27
deploy: poll for the full build window and validate the project name locally

wu-changxing:
> **This PR has no security effect on its own.** It contains a client-side name check, and a client-side check is not a boundary — anyone can call `POST /api/v1/deploy` directly and bypass it entirely. The command injection on the agents VM is live today and is fixed by **openonion/oo-api#34**, which is open and unmerged. Treat this PR as a UX improvement; merge oo-api#34 for the actual fix.

## Description

Fixes found while reviewing the `co deploy` path. The main one: the CLI gave up polling after 100 attempts (~5 minutes) while the backend allows the Docker build alone 900s, precisely because co-ai and browser images take minutes to build. Those deploys printed `Deploy status: deploying` / `try again` on a deploy that was still building and would go on to succeed.

## Related Issue

None — found during a review of `co deploy`.

## Type of Change

- [x] Documentation update

## Changes Made

**`co deploy` (`connectonion/cli/commands/deploy_commands.py`)**

- **Poll against a 20-minute deadline** covering the backend's own budget (rsync 120s + docker build 900s + run 60s) instead of a fixed 100 attempts. Progress shows time remaining rather than `[n/100]`, which was a misleading unit.
- **Reject an invalid deploy name locally**, before packaging and uploading the whole project, with the corrected name to paste back. Convenience only — see the banner above.
- **Check the project before the API key.** A broken `host.yaml` is worth reporting whether or not you happen to be logged in; previously `No API key` masked every project-level error.
- **Stop instead of polling `/deploy/None/status`.** If the response carries no `id`, the loop spent five minutes requesting a URL that could never resolve.
- **Print the backend's `detail`** on a failed POST instead of the raw JSON body.
- **`_exports_asgi_app` (was `_check_host_export`) no longer counts commented-out calls.** `\bhost\s*\(` matched `# host(agent)`, so a project that never calls `host()` could pass the pre-flight check and then fail at container start.

**`co create` / `co init` (`connectonion/cli/commands/project_cmd_lib.py`)**

The `name` in `host.yaml` becomes a DNS label and a Docker image tag at deploy time, but nothing produced it with that in mind — `co create My_Agent` wrote `My_Agent`, and `co init` wrote whatever the current directory happened to be called. Both produced projects that **could never deploy** (Docker rejects uppercase tags; `_` is not a legal DNS label) and only found out deep in the remote build.

`create_host_yaml` is the single point both paths funnel through, so the deploy name is normalized there and the adjustment is printed:

```
$ mkdir My_Project && cd My_Project && co init
  Deployment name set to my-project (from 'My_Project'; must be lowercase letters, digits, and hyphens)
```

Refusing to create the project would be the wrong trade — a directory is named for people, and the folder keeps its own name either way. Existing projects are untouched: `create_host_yaml` returns early when `host.yaml` already exists.

The rule has **one definition**, next to the code that writes `host.yaml`; `deploy_commands` imports it instead of carrying a copy. It is named for the deploy role (`DEPLOY_NAME_PATTERN` / `normalize_deploy_name`) so it is not confused with the existing `validate_project_name`, which governs the project/directory name and deliberately allows uppercase and underscores.

**Docs (`docs/network/deploy.md`)** — the naming rule governed a hostname and a Docker tag but was written down nowhere. Documented next to the `host.yaml` config it applies to, including the consequence below.

## Consequence worth knowing

Two folder names that reduce to the same deploy name (`My_Agent` and `my-agent`) now share one deployment URL, and deploying one replaces the other. Previously the first simply could not deploy at all, so nothing that used to work breaks — but it is a new way to collide, so it is called out in the docs with the fix (set `name` explicitly).

## Fixed after review

Four issues were raised in review; all four reproduced. ([full reply](https://github.com/openonion/connectonion/pull/281#issuecomment-5098563607))

- **A `#` inside a string literal hid a real `host()` call** later on the same line, because the comment rule is lexical — `msg = "note: #1"; host(agent)` was reported as "does not export an ASGI app", blocking a valid deploy. Single-line string literals are blanked before the rule runs.
- **An unquoted `name: 123`** parses as an `int`, and `re.match` on an int raises `TypeError` — a traceback in place of the clear message this code exists to print. The three `host.yaml` reads coerce to `str`.
- **The countdown could print `-1m55s left`** (`//` floors, `%` does not) when a status call returned near the deadline. Clamped to zero.
- **The security framing** was too soft; hence the banner at the top of this description.

## A bug this PR's own first draft introduced

The first version of `_exports_asgi_app` also excluded *attribute* calls, to skip things like `client.host(...)`. Testing the regex against real idioms showed that also excluded `connectonion.host(agent)` and `co.host(agent)` — legitimate ways to call it:

```
  True (want  True)  plain call
 False (want  True)  module-qualified call   <-- MISMATCH
 False (want  True)  fully qualified         <-- MISMATCH
```

That blocks a valid deploy with an error the user cannot act on. Letting a bad project through to a clear container error is the better failure, so the lookbehind now only rules out names *ending* in "host" (`ghost(`, `localhost(`) — which, checked against the original, `\b` already handled.

## Naming

Functions whose names would make a reader guess wrong were renamed (no behavior change): `_check_host_export` → `_exports_asgi_app`, which reads as the question the caller is asking rather than leaving "what does True mean" open.

## Testing

- [x] I have tested this code locally
- [x] All existing tests pass

`pytest tests/e2e/cli/test_cli_deploy.py test_cli_init.py test_cli_create.py -q` → **89 passed**. Full suite: `pytest tests/ -m "not slow and not real_api and not network"` → 2691 passed.

New coverage:
- `TestDeployValidation`: invalid names (`My-Agent`, `my_agent`, `x; rm -rf /`, leading hyphen) rejected before upload; the suggestion (`My_Agent` → `my-agent`) in the error; a response with no `id` aborting without any status request; a YAML-typed `name: 123` surviving the config read; `_exports_asgi_app` accepting all six real ways to call `host()` and a `#` inside a string, while rejecting commented-out calls and `ghost`/`localhost`.
- `co init` inside `My_Project` writes `my-project` and says so; a CJK-named directory falls back to `agent`; an already-valid directory name is left alone with no notice.
- `co create My_Agent` keeps the `My_Agent` folder but records `my-agent`.

Also verified outside the test suite: `_exports_asgi_app` against **all 6 template `agent.py` files** — it detects exactly `co-ai` and `hosted-browser` (the two that call `host()`) and no others. And `co create` already rejects names with spaces itself, so the gap was only uppercase and underscores.

One existing mock returned `{"deployment_id", "agent_url"}` — key names the backend has never sent. Corrected to `{"id", "url"}`, the real response shape, which is what the new no-id guard made visible. One new test initially made a real network call and took 105s; it is mocked and hermetic now.

## Example Usage

```python
# New projects get a deployable name automatically, whatever the folder is called:
#
#   $ mkdir My_Project && cd My_Project && co init
#     Deployment name set to my-project (from 'My_Project'; must be lowercase...)
#
# An older project whose host.yaml predates that still fails early, with the fix:
#
#   $ co deploy
#   Invalid project name: My_Agent
#   Names may use 1-39 lowercase letters, digits, and hyphens, and must start
#   with a letter or digit.
#   Try:  name: my-agent
#   Set 'name' in /path/to/project/.co/host.yaml
#
# And a slow template deploy is followed to completion instead of being
# abandoned at five minutes:
#
#   $ co deploy --template co-ai
#   Building container on ConnectOnion Cloud...
#     [19m54s left] status: deploying
#     ...
#     [11m06s left] status: running
#   Deployed!
```

## Checklist

## Additional Notes

The `LLM-Note` headers in both touched modules were updated for the new ordering, the new failure mode, the new poll budget, and the deploy-name rule.

### DIFF
--- a/connectonion/cli/commands/deploy_commands.py
+++ b/connectonion/cli/commands/deploy_commands.py
@@ -437,23 +437,23 @@ def _deploy_current_project(skills: list[str], project_dir: Path | None = None) 
     if deploy_success:
         console.print(f"Dashboard: {DASHBOARD_URL}")
 
-    # Show the agent's startup logs (best-effort).
-    if deployment_id:
-        time.sleep(5)  # "running" fires when the container starts; wait for the app to print its banner or crash
-        try:
-            logs_resp = requests.get(
-                f"{API_BASE}/api/v1/deploy/{deployment_id}/logs?tail=20",
-                headers={"Authorization": f"Bearer {api_key}"},
-                timeout=10,
-            )
-        except requests.exceptions.RequestException:
-            logs_resp = None
-        if logs_resp is not None and logs_resp.status_code == 200:
-            logs = logs_resp.json().get("logs", "")
-            if logs:
-                console.print()
-                console.print("[dim]Container logs:[/dim]")
-                console.print(f"[dim]{logs}[/dim]")
+    # Show the agent's startup logs (best-effort). deployment_id is always set
+    # here — the missing-id case returned above.
+    time.sleep(5)  # "running" fires when the container starts; wait for the app to print its banner or crash
+    try:
+        logs_resp = requests.get(
+            f"{API_BASE}/api/v1/deploy/{deployment_id}/logs?tail=20",
+            headers={"Authorization": f"Bearer {api_key}"},
+            timeout=10,
+        )
+    except requests.exceptions.RequestException:
+        logs_resp = None
+    if logs_resp is not None and logs_resp.status_code == 200:
+        logs = logs_resp.json().get("logs", "")
+        if logs:
+            console.print()
+            console.print("[dim]Container logs:[/dim]")
+            console.print(f"[dim]{logs}[/dim]")
 
     console.print()
     return deploy_success
--- a/connectonion/cli/commands/deploy_commands.py
+++ b/connectonion/cli/commands/deploy_commands.py
@@ -2,11 +2,11 @@
 Purpose: Deploy agent projects to ConnectOnion Cloud with local packaging and env vars
 LLM-Note:
   Dependencies: imports from [fnmatch, json, os, re, shutil, subprocess, tarfile, tempfile, time, yaml, requests, pathlib, rich.console, dotenv] | imported by [cli/main.py via handle_deploy()] | calls backend at [https://oo.openonion.ai/api/v1/deploy]
-  Data flow: handle_deploy() → optionally creates a temporary template project via co create (named by --name, default {template}-agent) → validates .co/host.yaml → load_api_key() loads OPENONION_API_KEY → reads host.yaml for project name, entrypoint, env file path → dotenv_values() loads env vars from .env → packages git-tracked files or initialized folder into tarball, merging each --skills path into .co/skills/ (a path that is itself a skill nests under its dirname) → POST to /api/v1/deploy with tarball + project_name + env_vars → polls /api/v1/deploy/{id}/status until running/error → displays agent URL
+  Data flow: handle_deploy() → optionally creates a temporary template project via co create (named by --name, default {template}-agent) → validates .co/host.yaml → reads host.yaml for project name, entrypoint, env file path → checks the name against PROJECT_NAME_PATTERN (same rule the backend enforces) and the entrypoint's host() export → load_api_key() loads OPENONION_API_KEY → dotenv_values() loads env vars from .env → packages git-tracked files or initialized folder into tarball, merging each --skills path into .co/skills/ (a path that is itself a skill nests under its dirname) → POST to /api/v1/deploy with tarball + project_name + env_vars → polls /api/v1/deploy/{id}/status until running/error → displays agent URL
   State/Effects: creates temporary tarball file in tempdir | template deploy creates/deletes a temporary project on success | reads .co/host.yaml, .env files | makes network POST request | prints progress to stdout via rich.Console | normal deploy does not modify project files
   Integration: exposes handle_deploy(template, skills, name) for CLI | expects .co/host.yaml (name, entrypoint, env) unless --template is used | --name only valid with --template (otherwise the name comes from host.yaml) | uses Bearer token auth | returns bool success
-  Performance: packaging is local file I/O | network timeout 600s for upload, 30s for status checks | polls every 3s up to 100 times (~5 min)
-  Errors: fails if not ConnectOnion project (no host.yaml) | fails if no API key | prints backend error messages
+  Performance: packaging is local file I/O | network timeout 600s for upload, 30s for status checks | polls every 3s for up to 20 min, covering the backend's own build budget (rsync 120s + docker build 900s + run 60s)
+  Errors: fails if not ConnectOnion project (no host.yaml) | fails if project name is not a valid hostname label | fails if no API key | prints backend error messages
 """
 
 import fnmatch
@@ -30,6 +30,16 @@ console = Console()
 API_BASE = "https://oo.openonion.ai"
 DASHBOARD_URL = "https://o.openonion.ai/dashboard"
 
+# Poll until the backend's own build budget is exhausted (rsync 120s + docker
+# build 900s + run 60s), plus slack for queueing. Giving up at 5 minutes told
+# users a co-ai or browser deploy had failed while it was still building.
+DEPLOY_POLL_SECONDS = 3
+DEPLOY_TIMEOUT_SECONDS = 20 * 60
+
+# Mirrors the server's rule (oo-api deploy/service.py). Checked locally so a bad
+# name in host.yaml fails before packaging and uploading the whole project.
+PROJECT_NAME_PATTERN = re.compile(r"^[a-z0-9][a-z0-9-]{0,38}$")
+
 
 def _check_host_export(entrypoint: str) -> bool:
     """Check if entrypoint file exports an ASGI app via host().
@@ -45,11 +55,12 @@ def _check_host_export(entrypoint: str) -> bool:
 
     content = entrypoint_path.read_text(encoding="utf-8")
 
-    # Check for host() call pattern
-    # Matches: host(agent), host(my_agent), host( agent ), etc.
-    host_call_pattern = r'\bhost\s*\([^)]+\)'
+    # A host(...) call on a line of its own or bound to a name. `[^#\n]*` keeps
+    # commented-out calls from counting, and the lookbehind rules out attribute
+    # calls and names that merely end in "host" (client.host(...), ghost(...)).
+    host_call_pattern = r'^[^#\n]*(?<![\w.])host\s*\([^)]+\)'
 
-    if re.search(host_call_pattern, content):
+    if re.search(host_call_pattern, content, re.MULTILINE):
         return True
 
     return False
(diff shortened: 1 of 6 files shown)


wu-changxing:
## Overview

Reviewed the diff in `connectonion/cli/commands/deploy_commands.py`, `connectonion/cli/commands/project_cmd_lib.py`, `docs/network/deploy.md`, and the new tests, and traced each of the PR description's claims against the actual code (not just the prose). I also cross-checked the "same rule the backend enforces" and "closes the hole" claims against the current `openonion/oo-api` checkout.

All six headline claims check out:
- Poll deadline is now `time.monotonic() + 20*60`, computed *after* the POST (so upload time doesn't eat into it) — matches the backend's own `rsync(120s) + docker build(900s) + run(60s) = 1080s` budget in `oo-api/deploy/service.py` (`_run_scp(timeout=120)`, `_run_ssh(build_cmd, timeout=900)`, `_run_ssh(run_cmd, timeout=60)`) with slack to spare.
- Project-name check (`DEPLOY_NAME_PATTERN.match`) now runs before `load_api_key()` in `_deploy_current_project`; confirmed by `test_deploy_rejects_invalid_project_name`, which never sets `OPENONION_API_KEY`.
- `if not deployment_id: return False` correctly precedes the polling loop; `test_deploy_stops_when_backend_returns_no_id` asserts `mock_get.assert_not_called()`, which matches the code path.
- `_error_text()` correctly falls back to `response.text` when `detail` isn't a string (e.g. FastAPI 422 validation errors return `detail` as a list, not a string) — handled gracefully rather than crashing.
- `_exports_asgi_app`'s new regex (`^[^#\n]*(?<[A-Za-z0-9_])host\s*\([^)]+\)`) — I hand-traced all six of the new test's positive cases (plain, assigned, kwargs, module-qualified, aliased, guarded) and the three negative cases (commented, trailing-comment, ghost/localhost) through the lookbehind and the `MULTILINE` `^` anchor; the logic is correct in every case tested.
- `create_host_yaml`'s CJK-directory fallback (`normalize_deploy_name(...) or "agent"`) is exercised and correct per `test_init_falls_back_when_the_directory_name_has_nothing_reusable`.

## Issues

1. **The local name check is not a security boundary, and the thing it's mitigating is still live in `oo-api` today.** `oo-api/deploy/service.py:_finish_deployment` builds `container_name` from `project_name` unvalidated and interpolates it directly into `build_cmd = f"cd {remote_dir} && docker build --no-cache -t {container_name} ."`, which `_run_ssh` passes as the remote command string to `ssh ... user@host <command>` — a real, currently unpatched command injection on the agents-vm. I checked `oo-api/deploy/routes.py` and there is no `project_name` validation anywhere yet, so oo-api#34 (referenced in the PR body) does not appear to be merged into this `oo-api` checkout. This PR's local regex only protects users of the unmodified `co` CLI; anyone who calls `POST /api/v1/deploy` directly (trivial with the existing API) bypasses it entirely. Worth stating more forcefully in the PR (not just "the pair is what closes the hole") that this half, alone, has no security effect — it's a UX improvement, not a fix.

2. **Minor: `remaining` can go negative near the deadline, producing a confusing display.** `remaining = int(deadline - time.monotonic())` is computed after the `requests.get(..., timeout=30)` call returns; if that call takes close to its full timeout near the deadline, `remaining` can be negative. Because Python's `%` on a negative int with a positive divisor returns a non-negative remainder while `//` floors toward `-∞`, e.g. `remaining = -5` prints as `"-1m55s left"` instead of something like `"0m00s left"`. Functionally harmless (the `while` condition still exits the loop correctly on the next check), but worth a `max(0, remaining)` guard for a clean readout.

3. **Minor: `DEPLOY_NAME_PATTERN.match(project_name)` assumes `project_name` is always a `str`.** `yaml.safe_load` will parse an unquoted numeric `name:` (e.g. `name: 123`) into an `int`. `re.Pattern.match` on an `int` raises `TypeError`, so a hand-edited `host.yaml` with an unquoted numeric name crashes with a traceback instead of the clean "Invalid project name" message this PR is specifically adding. A `str(project_name)` coercion (or an explicit `isinstance` check) before the regex would keep this a clean fail-fast error rather than a stack trace.

4. **Minor/edge case: the new comment-exclusion in `_exports_asgi_app` is purely lexical, not syntax-aware, so a `#` earlier in the same physical line — even inside a string literal — blocks detection of a real `host()` call later on that line.** E.g. `msg = "note: #1"; host(agent)` would be a false negative: `[^#\n]*` cannot cross the `#` inside the string, so the anchored match never reaches `host(`, and a genuinely valid deploy gets the "does not export an ASGI app" error. This is narrow (uncommon to chain a `#`-containing string and a `host()` call on one physical line) and consistent with the PR's own stated bias ("letting a bad project through... is the better failure" — this is the opposite direction, blocking a good one), so it's worth a one-line comment acknowledging the limitation, though probably not worth a real tokenizer for a pre-flight check.

## Done well

- The `LLM-Note` header updates in both files are actually kept in sync with the code changes (poll budget, check ordering, error format) — not just boilerplate.
- Test coverage is genuinely aimed at the edge cases described, not just the happy path (invalid names, no-id guard, comment/qualified-call regex matrix, CJK fallback, already-valid-name no-op).
- `create_host_yaml` correctly returns early / stays silent when the directory name is already valid (verified by `test_init_keeps_an_already_valid_directory_name_untouched`), so existing well-named projects get no behavior change.
- Reordering the "must have API key" check after project validation is a good, low-risk UX fix — a broken `host.yaml` is diagnosable without needing to be logged in first.

---

wu-changxing:
Thanks — all four were reproducible. Three are fixed in `bdeeb5c`; the first is a framing problem in my PR description, which I've rewritten.

**1. The local check is not a security boundary.** You're right, and my wording was too soft. The PR body now says so at the top rather than burying it in a closing note: this half is a UX improvement with **zero** security effect, and anyone can bypass it by calling `POST /api/v1/deploy` directly. The enforcement that matters is openonion/oo-api#34, which is open and not yet merged — so your reading of the `oo-api` checkout is correct, the injection is live today.

**2. Negative countdown.** Reproduced exactly as described:

```
remaining =  -5  ->  "-1m55s left"     # // floors, % does not
```

Clamped with `max(0, ...)`.

**3. YAML-typed name.** Reproduced — `yaml.safe_load("name: 123")` gives `int`, and `re.match` on it raises `TypeError: expected string or bytes-like object, got 'int'`. All three `host.yaml` reads (`name`, `env`, `entrypoint`) now coerce to `str`, so a hand-edited file fails cleanly or just works. `"123"` happens to be a legal deploy name, so that case now proceeds instead of crashing.

**4. `#` inside a string literal.** Reproduced — `msg = "note: #1"; host(agent)` returned `False`, which is the exact failure direction I argued against elsewhere in this PR (blocking a valid deploy with an unactionable error). Rather than only documenting it, I blank single-line string literals before applying the comment rule:

```python
content = re.sub(r'"[^"\n]*"|\'[^\'\n]*\'', '""', content)
```

One line, no tokenizer, and it keeps the trailing-comment behavior you verified:

```
True  (want True )  'msg = "note: #1"; host(agent)'
True  (want True )  "msg = 'issue #7'\nhost(agent)"
False (want False)  '# host(agent)'
False (want False)  'x = 1  # host(agent) later'
False (want False)  'ghost(agent)'
```

Triple-quoted strings are still not handled, and there's now a comment saying so — a `#` inside a docstring followed by a `host()` call on the same physical line isn't a real scenario.

Both new cases have tests. One of them initially made a real network call and took 105s; it's mocked and hermetic now.

---

PR #285, opened 2026-07-27
Give the `co browser` daemon the tab bounds BrowserAutomation already had

### LINKED ISSUE #234: All templates should drive the browser via the `co browser` CLI (prompt + shell tool), not the in-process BrowserAutomation tool

### LINKED ISSUE #283: co browser daemon: expose tab_idle_ttl and max_tabs (blocks hosted-browser moving to the CLI)
## The gap

`BrowserAutomation` takes two resource guards:

```python
BrowserAutomation(tab_idle_ttl=3600.0, max_tabs=10)
#   tab_idle_ttl — seconds a session's tab may sit idle before being reclaimed
#   max_tabs     — hard cap on concurrent session tabs
```

The `co browser` daemon exposes **neither**. `grep -c 'tab_idle_ttl\|max_tabs' connectonion/cli/browser_agent/daemon.py` → `0`. It constructs `BrowserAutomation(headless=headless)` and takes the defaults.

## Why it matters now

#279 moved the `minimal` and `browser` templates onto the CLI, and #273 moved `co ai`. **`hosted-browser` was deliberately left behind**, because it passes:

```python
BrowserAutomation(headless=False, tab_idle_ttl=600, max_tabs=3)
```

Its own README lists those under *"Knobs that matter in production"* — one shared browser serves every chat session, each session gets a tab, and these two numbers are what stop tabs accumulating until a small box runs out of memory.

Converting that template today would silently drop both. It would look like consistency and behave like a downgrade, which is why #234 is not finished.

## What this needs

Let the daemon carry the same two knobs and pass them through:

- `python -m connectonion.cli.browser_agent.daemon <sock> [--headless] [--tab-idle-ttl SECONDS] [--max-tabs N]`
- surfaced in `co browser status` alongside `headless=`, so an operator can see what the running daemon is enforcing
- and settable without editing a launch command — `CO_BROWSER_MAX_TABS` / `CO_BROWSER_TAB_IDLE_TTL`, since the daemon is usually started implicitly by the first client rather than by hand

Worth deciding at the same time: **whose knobs are they?** Today each caller constructs its own `BrowserAutomation` and picks its own limits. With one shared daemon there is one browser and one set of limits for every caller on the box. First-client-wins is the simplest rule and probably right, but it should be a decision rather than an accident — `co browser status` showing the effective values is what makes it debuggable either way.

## Then

`hosted-browser` can move to the CLI like the others, and #234 closes.

## Related

- #234 — all templates on `co browser`; blocked on this for the last template
- #279 — the two templates that could move, and why this one couldn't
- #273 — `co ai`, which had its own tab-sharing workaround (`bind_browser_session`) deleted because the daemon owns tabs

wu-changxing:
Closes #283. Unblocks the last template in #234.

## The gap

`BrowserAutomation` takes two resource guards; the daemon exposed **neither** and silently took the defaults:

```
BrowserAutomation(tab_idle_ttl=3600.0, max_tabs=10)
grep -c 'tab_idle_ttl\|max_tabs' daemon.py  →  0
```

That is precisely why `hosted-browser` could not follow #273 and #279 onto the CLI. It runs with `tab_idle_ttl=600, max_tabs=3`, and its own README lists those under *"Knobs that matter in production"* — one browser serves every chat session, and those two numbers are what stop tabs accumulating until a small box runs out of memory. Converting it would have looked like consistency and behaved like a downgrade.

## The change

```bash
co browser status
# Browser: not open · headless=false · max_tabs=3 · tab_idle_ttl=600s · targeting is per-command

CO_BROWSER_MAX_TABS=3 CO_BROWSER_TAB_IDLE_TTL=600 co browser status
python -m connectonion.cli.browser_agent.daemon <sock> --max-tabs 3 --tab-idle-ttl 600
```

**The env vars are not a convenience.** The daemon is usually started *implicitly by the first client*, so a hand-written command line is the one path nobody takes — without them the flags would be unreachable in normal use.

**`status` reporting the effective values matters for the same reason.** With one daemon per box there is one browser and one set of limits for every caller. Nobody can read their limits off a launch command they never wrote. "First client wins" is a defensible rule, but only debuggable if the winner is visible — so this PR makes it visible rather than leaving it implicit.

Defaults are untouched when nothing is asked for: `max_tabs=10`, `tab_idle_ttl=3600`, exactly as before.

## Tests

Four in `TestTabLimits`:

- defaults unchanged when nothing is asked for — constructed against the **real** `BrowserAutomation` rather than the stub, since its own defaults are the thing under test
- explicit limits reach the browser
- `status` reports what is in force
- flags and env both work, **and an explicit flag beats the environment**

Verified live against a fresh daemon, not only in tests:

```
$ CO_BROWSER_MAX_TABS=3 CO_BROWSER_TAB_IDLE_TTL=600 co browser status
Browser: not open · headless=false · max_tabs=3 · tab_idle_ttl=600s · ...
```

Full suite: **2,628 passed, 6 skipped**.

## Next

`hosted-browser` can now move to the CLI like the other two, which closes #234.

### DIFF
--- a/connectonion/cli/browser_agent/daemon.py
+++ b/connectonion/cli/browser_agent/daemon.py
@@ -129,9 +129,18 @@ def _held_by_other(meta, caller: str) -> bool:
 class BrowserDaemon:
     """Single-threaded server owning one BrowserAutomation, dispatching verbs to it."""
 
-    def __init__(self, sock_path: str, headless: bool = False):
+    def __init__(self, sock_path: str, headless: bool = False,
+                 tab_idle_ttl: float = None, max_tabs: int = None):
         self.sock_path = sock_path
-        self.browser = BrowserAutomation(headless=headless)
+        # One daemon owns the browser for every caller on the box, so these
+        # bounds are the box's, not one caller's. Left as BrowserAutomation's
+        # defaults unless set, so nothing changes for anyone who doesn't ask.
+        limits = {}
+        if tab_idle_ttl is not None:
+            limits["tab_idle_ttl"] = tab_idle_ttl
+        if max_tabs is not None:
+            limits["max_tabs"] = max_tabs
+        self.browser = BrowserAutomation(headless=headless, **limits)
         self._srv = None
         self._had_browser = False
         self.last_command = None  # {"line": str, "at": float} of the last real command
@@ -277,7 +286,16 @@ class BrowserDaemon:
         """Report browser state, the last command, and the tab board."""
         open_state = "open" if self.browser._context_is_alive() else "not open"
         headless = str(getattr(self.browser, "_headless", False)).lower()
-        lines = [f"Browser: {open_state} · headless={headless} · targeting is per-command (-t <tab>; bare = main)"]
+        # Show the limits actually in force: one daemon serves every caller, so
+        # "which numbers am I bound by?" is not answerable from any one launch
+        # command.
+        ttl = int(getattr(self.browser, "_tab_idle_ttl", 0))
+        cap = getattr(self.browser, "_max_tabs", 0)
+        lines = [
+            f"Browser: {open_state} · headless={headless} · "
+            f"max_tabs={cap} · tab_idle_ttl={ttl}s · "
+            "targeting is per-command (-t <tab>; bare = main)"
+        ]
         # Surface stealth-driver health here so a misconfigured driver (webdriver leak) is
         # visible where users look for browser state, not only in `co doctor`.
         stealth, version, detail = driver_stealth_status()
@@ -655,8 +673,28 @@ def main():
             if hasattr(stream, "reconfigure"):
                 stream.reconfigure(encoding="utf-8", errors="replace")
     sock_path = sys.argv[1] if len(sys.argv) > 1 else default_sock_path()
-    headless = "--headless" in sys.argv[2:]
-    BrowserDaemon(sock_path, headless=headless).serve()
+    args = sys.argv[2:]
+    headless = "--headless" in args
+
+    def _opt(flag: str, env: str):
+        """Flag wins over env; both absent means BrowserAutomation's default.
+
+        The env vars matter because the daemon is usually started implicitly by
+        the first client, not by a hand-written command line."""
+        if flag in args:
+            i = args.index(flag)
+            if i + 1 < len(args):
+                return args[i + 1]
+        return os.environ.get(env)
+
+    ttl = _opt("--tab-idle-ttl", "CO_BROWSER_TAB_IDLE_TTL")
+    cap = _opt("--max-tabs", "CO_BROWSER_MAX_TABS")
+    BrowserDaemon(
+        sock_path,
+        headless=headless,
+        tab_idle_ttl=float(ttl) if ttl else None,
+        max_tabs=int(cap) if cap else None,
+    ).serve()
 
 
 if __name__ == "__main__":
--- a/tests/e2e/cli/test_browser_daemon.py
+++ b/tests/e2e/cli/test_browser_daemon.py
@@ -16,6 +16,7 @@ or network is needed. The daemon's lazy BrowserAutomation is swapped for the stu
 
 import contextlib
 import os
+import sys
 import socket
 import subprocess
 import threading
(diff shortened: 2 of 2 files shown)


wu-changxing:
Marking this draft — the design is wrong and I should not have shipped it as-is.

I ported `BrowserAutomation`'s knobs onto the daemon without asking whether they still make sense once one process owns the browser for the whole box. They don't.

**The daemon already has an ownership model, and this ignores it.** Every request carries a `caller`, and `browser._tab_meta[key]` records `caller`, `claim_at` and `last_at` per tab. `_held_by_other()` already arbitrates contention *by identity* inside a 120s window. So the daemon knows exactly who owns which tab and when it was last touched — and I added a single global number, set by whichever process happened to win the startup race, frozen for the daemon's lifetime, unreachable to every caller that arrives afterwards.

"First client wins" isn't a policy, it's an accident. Making it visible in `status` — which I argued for — only makes the accident easier to observe. That was me defending the shape instead of questioning it.

Replacing this with a proposal on #283.

wu-changxing:
Superseded by the declared-occupancy model — see the new PR. Porting max_tabs/tab_idle_ttl onto a shared daemon was the wrong shape: a global cap set by whoever wins the startup race is an accident rather than a policy, and it ignored the identity model the daemon already had.

PR #293, opened 2026-07-29
fix: preserve system prompts in provider requests

### LINKED ISSUE #289: [BUG] AnthropicLLM silently drops the system prompt on every Claude call
## Summary

`AnthropicLLM` strips `system` messages during conversion but never passes them to the Messages API. The system prompt is silently discarded on **every** `claude-*` call — no error, no warning, the agent just behaves as if it had no instructions.

Found while porting `core/llm.py` to Rust ([connectonion-rust](https://github.com/wu-changxing/connectonion-rust)); the Rust port passes `system` correctly, which is what made the difference visible.

Version: `1.4.0` (`main` @ 96ea117)

## Where it goes wrong

`connectonion/core/llm.py:422-424` removes system messages, with a comment promising they get handled elsewhere:

```python
# Skip system messages (will be handled separately)
if msg["role"] == "system":
    i += 1
    continue
```

But neither caller ever sets `system=`:

- `complete()` — `api_kwargs` at line 330 is `{model, messages, max_tokens, **kwargs}`
- `structured_complete()` — `api_kwargs` at line 394 is `{model, max_tokens, messages, tools, tool_choice, **kwargs}`

The string `system` appears exactly twice in the entire class, both in the skip above. Anthropic requires the system prompt as a **top-level `system` parameter**, not as a message, so removing it without re-adding it drops it entirely.

## Reproduce (no API key needed)

```python
import re
src = open('connectonion/core/llm.py').read()
cls = src[src.index('class AnthropicLLM'):src.index('class GeminiLLM')]

for m in re.finditer(r'api_kwargs = \{(.*?)\n        \}', cls, re.S):
    print('--- api_kwargs ---'); print(m.group(1).rstrip())

print('\nlines mentioning "system":')
for l in cls.splitlines():
    if 'system' in l: print('   ', l.strip())
```

Output — no `system` key in either request, and the only mentions are the skip:

```
--- api_kwargs ---
            "model": self.model,
            "messages": anthropic_messages,
            "max_tokens": self.max_tokens,
            **kwargs
--- api_kwargs ---
            "model": self.model,
            "max_tokens": self.max_tokens,
            "messages": anthropic_messages,
            "tools": [tool],
            "tool_choice": {"type": "tool", "name": "return_structured_output"},
            **kwargs

lines mentioning "system":
     # Skip system messages (will be handled separately)
     if msg["role"] == "system":
```

Behaviourally:

```python

agent = Agent(
    "pirate",
    system_prompt="You are a pirate. Always answer in pirate speak.",
    model="claude-3-5-haiku-latest",
)
print(agent.input("What is 2+2?"))
# Answers as a normal assistant — the pirate instruction never reached Claude.
# Swap to model="gpt-4o" and it answers in pirate speak.
```

## Impact

Every Claude user is affected, and the failure is invisible:

- `Agent(..., model="claude-*")` — `system_prompt`, including `prompt.md` files loaded via `load_system_prompt()`, is ignored
- `llm_do(..., model="claude-*", system_prompt=...)` — same
- The `coder` / `web-research` templates lose their whole prompt if pointed at Claude
- Anything relying on prompt-encoded safety or scope constraints loses them on Claude specifically

Nothing surfaces this to the user — the agent just quietly behaves like it has a default prompt.

## Suggested fix

Have `_convert_messages` return the system text alongside the converted messages, and set it on the request. Multiple system messages should concatenate rather than the last one winning.

```python
def _convert_messages(self, messages):
    system_parts = []
    anthropic_messages = []
    i = 0
    while i < len(messages):
        msg = messages[i]
        if msg["role"] == "system":
            system_parts.append(msg.get("content", ""))
            i += 1
            continue
        ...  # unchanged
    system = "\n\n".join(p for p in system_parts if p)
    return anthropic_messages, (system or None)
```

Then in both `complete()` and `structured_complete()`:

```python
anthropic_messages, system = self._convert_messages(messages)

api_kwargs = {
    "model": self.model,
    "messages": anthropic_messages,
    "max_tokens": self.max_tokens,
    **kwargs,
}
if system:
    api_kwargs["system"] = system
```

`structured_complete()` needs the same treatment — it currently drops the prompt too.

## Suggested regression test

A unit test asserting the prompt reaches the request would have caught this and would keep it caught:

```python
def test_anthropic_passes_the_system_prompt(monkeypatch):
    captured = {}

    class FakeMessages:
        def create(self, **kwargs):
            captured.update(kwargs)
            return SimpleNamespace(
                content=[SimpleNamespace(type="text", text="ok")],
                usage=SimpleNamespace(input_tokens=1, output_tokens=1),
            )

    llm = AnthropicLLM(api_key="test")
    llm.client = SimpleNamespace(messages=FakeMessages())
    llm.complete([
        {"role": "system", "content": "You are a pirate."},
        {"role": "user", "content": "hi"},
    ])

    assert captured["system"] == "You are a pirate."
    assert all(m["role"] != "system" for m in captured["messages"])
```

Happy to send a PR with the fix plus this test if that's useful — just say the word.

sanmaxdev:
## Description

Preserves system instructions when converting messages for provider requests.

## Related Issue

Fixes #289

## Type of Change

## Changes Made

- Collects multiple system messages without placing them in the conversation message list.
- Sends combined system instructions for standard and structured completions.
- Adds regression coverage for combined, absent, and structured system instructions.

## Testing

- [x] `.venv/bin/python -m pytest tests/ -q --tb=short -m 'not slow and not real_api and not network'`
- [x] `git diff --check`

## Checklist

- [x] New and existing unit tests pass

### DIFF
--- a/connectonion/core/llm.py
+++ b/connectonion/core/llm.py
@@ -325,7 +325,7 @@ class AnthropicLLM(LLM):
     def complete(self, messages: List[Dict[str, Any]], tools: Optional[List[Dict[str, Any]]] = None, **kwargs) -> LLMResponse:
         """Complete a conversation with optional tool support."""
         # Convert messages to Anthropic format
-        anthropic_messages = self._convert_messages(messages)
+        anthropic_messages, system = self._convert_messages(messages)
 
         api_kwargs = {
             "model": self.model,
@@ -333,6 +333,9 @@ class AnthropicLLM(LLM):
             "max_tokens": self.max_tokens,  # Required by Anthropic
             **kwargs  # User can override max_tokens via kwargs
         }
+
+        if system:
+            api_kwargs["system"] = system
 
         # Add tools if provided
         if tools:
@@ -381,7 +384,7 @@ class AnthropicLLM(LLM):
         workaround: create a dummy tool with the Pydantic schema and force its use.
         """
         # Convert messages to Anthropic format
-        anthropic_messages = self._convert_messages(messages)
+        anthropic_messages, system = self._convert_messages(messages)
 
         # Create a tool with the Pydantic schema as input_schema
         tool = {
@@ -400,6 +403,9 @@ class AnthropicLLM(LLM):
             **kwargs  # User can override max_tokens, temperature, etc.
         }
 
+        if system:
+            api_kwargs["system"] = system
+
         # Force the model to use this tool
         response = self.client.messages.create(**api_kwargs)
 
@@ -411,16 +417,19 @@ class AnthropicLLM(LLM):
 
         raise ValueError("No structured output received from Claude")
 
-    def _convert_messages(self, messages: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    def _convert_messages(self, messages: List[Dict[str, Any]]) -> tuple[List[Dict[str, Any]], Optional[str]]:
         """Convert OpenAI-style messages to Anthropic format."""
         anthropic_messages = []
+        system_parts = []
         i = 0
         
         while i < len(messages):
             msg = messages[i]
             
-            # Skip system messages (will be handled separately)
+            # Anthropic accepts system instructions as a top-level request parameter.
             if msg["role"] == "system":
+                if msg.get("content"):
+                    system_parts.append(msg["content"])
                 i += 1
                 continue
             
@@ -513,7 +522,8 @@ class AnthropicLLM(LLM):
             else:
                 i += 1
         
-        return anthropic_messages
+        system = "\n\n".join(system_parts)
+        return anthropic_messages, system or None
     
     def _convert_tools(self, tools: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
         """Convert OpenAI-style tools to Anthropic format."""
--- a/tests/unit/test_anthropic_llm.py
+++ b/tests/unit/test_anthropic_llm.py
@@ -1,0 +1,76 @@
+"""Unit tests for Anthropic request conversion."""
+
+from types import SimpleNamespace
+from unittest.mock import Mock
+
+from pydantic import BaseModel
+
+from connectonion.core.llm import AnthropicLLM
+
+
+class Answer(BaseModel):
+    value: str
+
+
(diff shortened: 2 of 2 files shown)


wu-changxing:
Reviewed carefully and verified locally. **This looks correct — I'd merge it.**

I had independently written the same fix before noticing this PR was already open, and yours is the better shape, so I closed mine (#304). Returning `(messages, system)` from a single pass of `_convert_messages` keeps the line that *drops* the system message and the line that *collects* it in the same loop, where they cannot drift apart. My version put the collection in a separate helper, which would silently stop matching if someone later changed the skip condition — the exact shape of the original bug.

**What I checked**

- **Signature change is fully covered.** `_convert_messages` has exactly two real call sites (`llm.py:328` and `:387`); both are updated here. Everything else `grep` turns up is a vendored copy under `.venv`/`.tox`/worktrees.
- **Tests genuinely bite.** Reverted `llm.py` to the base commit and re-ran `tests/unit/test_anthropic_llm.py`: **2 of 3 fail**. (The third asserts `system` is *absent* with no system message, which trivially holds when nothing ever sets it — worth keeping as a lock, just not evidence.)
- **No regressions.** `tests/unit` on this branch: **2364 passed, 3 skipped**.
- **`tuple[...]` is fine** — `requires-python = ">=3.10"`.
- Empty-string system messages are correctly skipped rather than joined into a stray blank line.

**One known limitation, not a change request:** if a system message's `content` is a list of content blocks rather than a string, `"\n\n".join()` raises `TypeError`. That path is unreachable through the Agent — `load_system_prompt()` is typed `-> str` and always returns a string — and failing loudly on an unsupported shape is the right behaviour for this codebase, so I would leave it.

**On the red CI:** it is not this PR's fault. The failures are all in `tests/e2e/cli/test_cli_auth_google.py` and `test_cli_auth_microsoft.py` — the fork-PR-secrets problem, unrelated to Anthropic. #184 (also yours) is what fixes that. Either land #184 first and re-run, or merge on the local verification above.

Thanks for this one — it is a genuinely severe bug. Every `claude-*` agent was running with no system prompt at all, silently.

PR #304, opened 2026-07-29
fix(anthropic): send the system prompt to Claude (#289)

### LINKED ISSUE #289: [BUG] AnthropicLLM silently drops the system prompt on every Claude call

wu-changxing:
Fixes #289.

## The bug

`_convert_messages` skips system messages with the comment "will be handled separately":

```python
# connectonion/core/llm.py
if msg["role"] == "system":
    i += 1
    continue
```

That is correct on its own — Anthropic takes the system prompt as a top-level `system` argument, not as a message. But **"separately" was never implemented.** Neither `complete()` nor `structured_complete()` ever set `api_kwargs["system"]`.

So every call to a `claude-*` model went out with **no system prompt at all**. No persona, no instructions, no constraints — and no error, no warning. Silent.

## The fix

`_extract_system()` collects the system messages; both call sites put the result into `api_kwargs["system"]` when non-empty. Multiple system messages join with a blank line, in order. When there is no system message the key is absent entirely rather than `None`.

Also documented on `_convert_messages` *why* it drops them, so the next reader doesn't restore the bug by "fixing" the skip.

## Tests

`TestAnthropicSystemPrompt` — 4 cases: `complete()`, `structured_complete()` (the second call site had the same bug), multiple system messages joined in order, and no-system-message sending no key.

Verified red before the fix: 3 of the 4 fail on the parent commit. (The fourth asserts `system` is *absent*, which trivially holds when it is never set — kept because it locks the behaviour going forward.)

`tests/unit` is 2365 passed / 3 skipped.

### DIFF
--- a/connectonion/core/llm.py
+++ b/connectonion/core/llm.py
@@ -334,6 +334,10 @@ class AnthropicLLM(LLM):
             **kwargs  # User can override max_tokens via kwargs
         }
 
+        system = self._extract_system(messages)
+        if system:
+            api_kwargs["system"] = system
+
         # Add tools if provided
         if tools:
             api_kwargs["tools"] = self._convert_tools(tools)
@@ -400,6 +404,10 @@ class AnthropicLLM(LLM):
             **kwargs  # User can override max_tokens, temperature, etc.
         }
 
+        system = self._extract_system(messages)
+        if system:
+            api_kwargs["system"] = system
+
         # Force the model to use this tool
         response = self.client.messages.create(**api_kwargs)
 
@@ -411,15 +419,31 @@ class AnthropicLLM(LLM):
 
         raise ValueError("No structured output received from Claude")
 
+    def _extract_system(self, messages: List[Dict[str, Any]]) -> Optional[str]:
+        """Pull the system prompt out for Anthropic's top-level `system` parameter.
+
+        Anthropic takes the system prompt as its own argument, not as a message in
+        the list — which is why _convert_messages drops system messages. Whoever
+        builds api_kwargs has to put it back, or the agent runs with no system
+        prompt at all and nothing reports it.
+        """
+        parts = [msg["content"] for msg in messages
+                 if msg.get("role") == "system" and msg.get("content")]
+        return "\n\n".join(parts) if parts else None
+
     def _convert_messages(self, messages: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
-        """Convert OpenAI-style messages to Anthropic format."""
+        """Convert OpenAI-style messages to Anthropic format.
+
+        System messages are dropped here on purpose — they travel as the top-level
+        `system` argument instead. Use _extract_system() to get them.
+        """
         anthropic_messages = []
         i = 0
-        
+
         while i < len(messages):
             msg = messages[i]
-            
-            # Skip system messages (will be handled separately)
+
+            # Dropped here, carried by _extract_system() into api_kwargs["system"]
             if msg["role"] == "system":
                 i += 1
                 continue
--- a/tests/unit/test_llm_errors.py
+++ b/tests/unit/test_llm_errors.py
@@ -330,6 +330,96 @@ class TestProviderErrorBubbling:
                 llm.complete([{"role": "user", "content": "test"}])
 
 
+class TestAnthropicSystemPrompt:
+    """The system prompt must reach Claude as the top-level `system` argument.
+
+    Anthropic does not accept a system message inside `messages`, so
+    _convert_messages drops it. If the caller does not put it back into
+    api_kwargs["system"], the agent runs with no system prompt and nothing
+    reports it — no error, no warning, just an agent with no instructions.
+    """
+
+    def _mock_response(self):
+        response = Mock()
+        block = Mock()
+        block.type = "text"
+        block.text = "ok"
+        response.content = [block]
+        response.usage.input_tokens = 1
+        response.usage.output_tokens = 1
+        response.usage.cache_read_input_tokens = 0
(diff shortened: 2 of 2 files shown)


wu-changxing:
Closing in favour of #293, which fixes the same bug and was opened first.

@sanmaxdev's approach is also the better one: their `_convert_messages` returns `(messages, system)` from a single pass, so the line that *drops* the system message and the line that *collects* it are in the same loop and cannot drift apart. My version put the collection in a separate `_extract_system()`, which would silently stop matching if someone later changed the skip condition — the exact shape of the original bug.

I should have checked for an open PR on #289 before writing a fix. My mistake.

One note that matters for #293: its red CI is **not its fault**. The failures are in `tests/e2e/cli/test_cli_auth_google.py` and `test_cli_auth_microsoft.py` — the fork-PR-secrets problem, unrelated to Anthropic. #184 is what fixes that, so #184 should land first and then #293 needs a re-run to show true green.

Issue #311, opened 2026-07-30
co server: register, list and preflight a deploy target


wu-changxing:
Part of #309. **Depends on nothing** — it can target a box you made by hand, which is also
how the rest of the chain gets tested before provisioning exists.

## Scope

```bash
co server add prod --ssh aaron@1.2.3.4   # register a machine
co server ls                             # what you can deploy to
co server check prod                     # preflight
```

**`co server add` takes an ssh target, not a credential.** We shell out to `ssh` and let the
operator's own agent and `~/.ssh/config` do the work — the same principle as `co browser`
driving a real browser instead of reimplementing one. We store a name → target mapping in
`~/.co/servers.yaml` and hold no secret.

`co server check` verifies over ssh and reports which requirement failed, by name:

- reachable at all, and the key works
- Ubuntu 24.04
- python 3.11+
- systemd present, and the user may manage units
- enough free disk

**Keep the promise narrow.** "It doesn't work on my box" is answerable only if we stated a
short list and this command says which item is missing. Do not attempt to support arbitrary
distributions.

## Definition of done

- `co server add` on an unreachable host fails immediately with the ssh error, not a timeout
- `co server ls` shows name, target, and last check result
- `co server check` names the failing requirement — tested by pointing it at a box missing
  each one in turn
- Nothing secret is written to `~/.co/servers.yaml`

Issue #309, opened 2026-07-29
One recovery phrase, one server you own, a deploy that does not erase it


wu-changxing:
Rewritten as a design. Absorbs #305 and #306; both stay open as the specific fixes.

## Goal

**One recovery phrase, one server, a deploy that does not erase it.**

An operator should be able to have a server created for them, shell into it with a key
derived from the phrase they already have, deploy an agent onto it, fix something by
hand, deploy again, and still have their fix. Today none of that is true.

## Scope decisions already made

- **We create the servers**, on GCE, from the backend. A server belongs to a project and
  persists; a deploy goes *onto* it.
- **No migration.** Existing deployments on the shared VM keep working on the existing
  path. The new path is added beside it. Nothing is moved, nothing is forced.
- **The deploy target is a first-class thing the user chooses** — `co deploy --to <server>`.
  The same path also accepts a machine the operator already has, because it is the same
  code once an ssh target exists.

### One seed, two derived keys

`address.py` already turns a 12-word BIP39 phrase into the agent's Ed25519 identity:

```python
seed = mnemo.to_seed(seed_phrase)      # 64 bytes
signing_key = SigningKey(seed[:32])    # only the first 32 are used
```

**Half the seed is unused.** The same phrase can back the operator's SSH key, so there is
still exactly one thing to write down.

Derive both with distinct labels rather than slicing — slicing yields exactly two keys and
no room for a third:

```
agent identity : HKDF(seed, info="connectonion:agent:v1")   → Ed25519, behaviour unchanged
ssh access     : HKDF(seed, info="connectonion:ssh:v1")     → Ed25519, OpenSSH format
```

Two keys, not one used twice: a signing oracle in the agent protocol must not be usable
against SSH login. Same root, separate purposes.

Ed25519 is a native OpenSSH type, so the derived key emits an ordinary `ssh-ed25519 AAAA…`
line — nothing custom server-side. `co keys` already shows the identity and phrase; it
grows the ability to print the derived SSH public key and, on request, write the private
half into `~/.ssh/`.

### Provisioning

`co server new prod` asks the backend to create a GCE instance and returns an ssh target.
The instance's startup script does the whole of the setup:

1. install python and create the venv
2. write the operator's **derived SSH public key** into `authorized_keys`
3. install Caddy for TLS termination
4. create the state directory the deploy will not touch, and the systemd unit

DNS: an A record for `<name>.agents.openonion.ai` pointing at the instance, so Caddy can
[...]

Issue #306, opened 2026-07-29
Every redeploy gives the agent a new identity — the container keeps no state


wu-changxing:
## What happens

A redeploy silently changes the agent's cryptographic identity. Three facts combine:

1. **The tarball never contains `.co/keys/`** — it is in ConnectOnion's built-in
   `GITIGNORE_CONTENT`, which `_load_deploy_ignore_patterns()` applies on both
   packaging paths. (This is correct: the private key should not travel.)
2. **The backend syncs with `--delete` and no excludes** —
   `oo-api/deploy/service.py:290`:
   ```python
   rsync_cmd = [
       "rsync", "-avz", "--delete",
       ...
       f"{local_path}/", f"{GCE_VM_USER}@{GCE_VM_HOST}:{remote_path}/"
   ]
   ```
   `--delete` removes anything in the destination that is absent from the source, so
   the `.co/keys/` the agent generated on its *first* boot is deleted on the next
   deploy. There are no `--exclude` flags and no volume mounts, so nothing survives
   outside the synced directory.
3. **`host()` regenerates when the key is missing** —
   `address.load(co_dir)` → `None` → `address.generate()` + `address.save()`.

Net effect: **first deploy mints identity A; every redeploy mints a fresh identity.**

## Why this is worse than it sounds

The address is not just a name — things are derived from it and keyed to it:

- **The agent's email changes.** It is `first 10 chars of the address + @mail.openonion.ai`
  (`address.py`), so mail routed to the old address stops arriving.
- **Every trust relationship other agents granted is void.** Anyone who whitelisted or
  promoted this agent whitelisted address A; after a redeploy it connects as a stranger
  and has to re-onboard.
- **The relay directory entry is orphaned.** `host()` announces a profile under the old
  address; consumers holding that address see a dead agent and a new unknown one.
- **The same `--delete` also wipes `.co/logs/`** (gitignored for the same reason), so the
  audit trail resets on every deploy. Whatever the agent recorded about its own runs is
  gone precisely when you would want it — after a change.

Nothing reports any of this. The deploy succeeds, the URL still works, and the identity
underneath has been swapped.

## Fix options

1. **Exclude the state directory from the sync** — `--exclude='.co/keys/'`
   (and `.co/logs/`) so a redeploy stops deleting what it never shipped. Minimal, and
   fixes it today. But it is VM-local: replace or rebuild the VM and the identity is
   still lost.
2. **Store the identity in the deployment record and inject it at boot** — durable
[...]

PR #315, opened 2026-07-30
feat(keys): derive an SSH key from the same recovery phrase (#310)

### LINKED ISSUE #310: co keys: derive an SSH key from the same recovery phrase
Part of #309. **No dependencies — this is the start of the chain and is testable on its own.**

## Why

`address.py` turns a 12-word BIP39 phrase into the agent's Ed25519 identity and uses only
half the material:

```python
seed = mnemo.to_seed(seed_phrase)      # 64 bytes
signing_key = SigningKey(seed[:32])    # the second half is unused
```

The same phrase can therefore back the operator's SSH key, so there is still exactly one
thing to write down — and every later piece (`authorized_keys` on a provisioned server,
admin seeding, deploy access) has a key to use.

## Scope

Derive both from the seed with distinct labels rather than slicing it. Slicing yields
exactly two keys and no room for a third:

```
agent identity : HKDF(seed, info="connectonion:agent:v1")   → must reproduce today's address
ssh access     : HKDF(seed, info="connectonion:ssh:v1")     → Ed25519, OpenSSH format
```

**Two keys, not one used twice.** A signing oracle in the agent protocol must not be usable
against SSH login.

CLI:

```bash
co keys --ssh              # print the ssh-ed25519 AAAA… public line
co keys --ssh --write      # also write the private half to ~/.ssh/, 0600
```

Ed25519 is a native OpenSSH type, so the public line is ordinary and needs nothing custom
on any server.

## Backwards compatibility — the important constraint

**Existing agents must keep their current address.** Their identity was derived as
`SigningKey(seed[:32])`, not through HKDF. Either keep that path for the agent key and use
HKDF only for the new SSH key, or version the derivation and migrate deliberately. Changing
an existing agent's address silently would void every trust relationship it has — the same
failure as #306, self-inflicted.

Pick one and state it in the PR.

## Definition of done

- A known phrase produces the same agent address as today — asserted by a test with a fixed
  vector, so a future refactor cannot move it
- The same phrase produces a stable SSH public key, also fixed-vector tested
- The derived key actually authenticates: write it to `authorized_keys` on a throwaway box
  and log in with it
- `co keys --ssh` prints a line usable verbatim in `authorized_keys`
- The two derived keys are different, and neither can be computed from the other without
  the seed

wu-changxing:
Closes #310. First link in the #309 chain — no dependencies, testable on its own.

## The choice #310 asked us to make and state

> Either keep that path for the agent key and use HKDF only for the new SSH key, or version the derivation and migrate deliberately.

**Kept the agent key exactly as it is: bare `SigningKey(seed[:32])`.** Only the SSH key is derived with a label. No migration, no risk to any existing address.

```
agent identity : SigningKey(seed[:32])                     unchanged
ssh access     : HKDF(seed, info="connectonion:ssh:v1")
```

Deriving the agent key through HKDF too would have been tidier, and would have silently reissued every existing agent's address — voiding every trust relationship keyed to it, which is the same failure as #306 but self-inflicted. The label on the SSH key means a third purpose can still be added later without moving either of the first two.

## What's here

- `address.derive_ssh_key(phrase)` → `{public_line, private_key}`. RFC 5869 HKDF-SHA512, then the OpenSSH wire encodings, written by hand — the format is a handful of length-prefixed strings and this adds no dependency.
- `co keys --ssh` prints the `ssh-ed25519 AAAA… connectonion` line, ready to paste into `authorized_keys`.
- `co keys --ssh --write` also writes `~/.ssh/connectonion_ed25519` at 0600 and the `.pub` at 0644. It **refuses to overwrite** an existing file — the key is derived, so nothing is ever lost by declining.

## Verification

**Cross-checked against real ssh tooling, not just the shape of the output.** `ssh-keygen -y` reads the hand-written private key and derives the identical public line:

```
$ ssh-keygen -y -f connectonion_ed25519
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIBs9N4V0K4pXPZNr5XPKmSJusfyalyjdi4xN36gqLt8U connectonion
$ ssh-keygen -l -f connectonion_ed25519.pub
256 SHA256:PTVCk2h22ZuZeNJvG1aZpeTcHyuVWe8XM0vJQhmGPcg connectonion (ED25519)
```

`tests/unit/test_address_ssh.py` — 11 cases. The ones that matter:

- **Fixed vector on the agent address.** A known phrase must still produce `0xc6f2ac55…`, so a future refactor cannot move it quietly.
- **The agent derivation itself is pinned**, not just its output — recomputing `SigningKey(seed[:32])` independently and asserting equality. If someone routes both keys through HKDF, this fails immediately.
- **Fixed vector on the SSH public line.**
- The two keys differ, and the SSH key is not the unused second half of the seed either (a slice, not a derivation).
- The public blob is decoded the way sshd does it — length-prefixed `ssh-ed25519` + 32 bytes — and the private section's length is asserted to be a multiple of the cipher block size, which is the mistake that yields a file ssh-keygen silently refuses.

Full suite: **2375 passed, 3 skipped**.

## Not verified here

#310's definition of done includes logging in to a throwaway box with the key. I verified the format with `ssh-keygen` instead, which covers the encoding but not sshd's acceptance path. Worth doing once #311 lands, since that is when a real target exists.

### DIFF
--- a/connectonion/address.py
+++ b/connectonion/address.py
@@ -307,3 +307,143 @@ def sign(address_data: Dict[str, Any], message: bytes) -> bytes:
         
     signed = address_data["signing_key"].sign(message)
     return signed.signature
+
+# ---------------------------------------------------------------------------
+# SSH access key
+#
+# The agent identity above uses only the first 32 of the seed's 64 bytes. The
+# same recovery phrase can therefore also back the operator's SSH key, so there
+# is still exactly one thing to write down.
+#
+# The agent key is deliberately left on its original derivation — bare
+# seed[:32]. Deriving it through HKDF instead would change every existing
+# agent's address and void every trust relationship keyed to it. Only the new
+# SSH key is derived with a label, so a third purpose can be added later
+# without disturbing either of the first two.
+#
+#     agent identity : SigningKey(seed[:32])                     (unchanged)
+#     ssh access     : HKDF(seed, info="connectonion:ssh:v1")
+#
+# Two keys, not one used twice: a signing oracle in the agent protocol must not
+# be usable against SSH login.
+# ---------------------------------------------------------------------------
+
+SSH_DERIVATION_INFO = b"connectonion:ssh:v1"
+
+
+def _hkdf_sha512(seed: bytes, info: bytes, length: int = 32) -> bytes:
+    """RFC 5869 HKDF with SHA-512, no salt. Enough for one 32-byte output."""
+    import hashlib
+    import hmac
+
+    prk = hmac.new(b"\x00" * hashlib.sha512().digest_size, seed, hashlib.sha512).digest()
+
+    out = b""
+    block = b""
+    counter = 1
+    while len(out) < length:
+        block = hmac.new(prk, block + info + bytes([counter]), hashlib.sha512).digest()
+        out += block
+        counter += 1
+    return out[:length]
+
+
+def derive_ssh_key(seed_phrase: str) -> Dict[str, str]:
+    """Derive an Ed25519 SSH keypair from a recovery phrase.
+
+    Ed25519 is a native OpenSSH type, so the public half is an ordinary
+    `ssh-ed25519 AAAA…` line that needs nothing custom on any server.
+
+    Args:
+        seed_phrase: The 12-word BIP39 recovery phrase
+
+    Returns:
+        {"public_line": "ssh-ed25519 AAAA… connectonion", "private_key": "-----BEGIN…"}
+
+    Example:
+        >>> keys = derive_ssh_key("legal winner thank year wave …")
+        >>> keys["public_line"].startswith("ssh-ed25519 ")
+        True
+    """
+    if Mnemonic is None or SigningKey is None:
+        raise ImportError(
+            "Missing dependencies. Install with:\n"
+            "  pip install pynacl mnemonic"
+        )
+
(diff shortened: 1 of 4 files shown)


wu-changxing:
**Update: the gap I flagged is now closed.** I said the login path was unverified because `ssh-keygen` only covers the encoding, not sshd's acceptance. A real box now exists.

Created a throwaway GCE instance with the derived public key in its metadata — nothing else, no manual key handling:

```
gcloud compute instances create co-test-deploy \
  --image-family=ubuntu-2404-lts-amd64 \
  --metadata-from-file=ssh-keys=<the line co keys --ssh prints>
```

Then logged in with the derived private half:

```
$ ssh -i connectonion_ed25519 co@136.85.31.111 'echo CONNECTED; …'
CONNECTED
distro=ubuntu 24.04
python 3.12
systemd=yes
sudo=yes
  17G
```

**sshd accepted the key.** The full chain works: 12 words → HKDF → OpenSSH private key file → GCE metadata → `authorized_keys` → authenticated session. #310's definition of done is met in full.

Worth noting what this proves beyond the login: the public line went into GCE metadata **verbatim**, with no reformatting, which is the practical claim behind "Ed25519 is a native OpenSSH type, so it needs nothing custom server-side".

Issue #312, opened 2026-07-30
co deploy --to: converge the server, sync the code, restart the unit


wu-changxing:
Part of #309. Depends on #311 (needs a target) and #310 (needs the SSH key).
**This is the largest piece and may reasonably land as two PRs: server-side setup, then the
deploy verb.**

## Scope

```
co deploy --to prod  =  ensure(setup) → sync code → restart
```

`ensure()` is a no-op after the first time. That framing matters: a machine registered with
`co server add` never went through provisioning, so its first deploy performs the setup —
same code, no second path.

Server layout:

```
/srv/<agent>/            code, rsynced
/srv/<agent>/.venv       dependencies
/srv/<agent>/.co/        state — the deploy never touches this
/etc/systemd/system/<agent>.service
```

**No container.** With one server per agent the server is already the boundary and the
resource limit. systemd provides what the container was used for: `Restart=always`,
`enable` for start-on-boot, journald for logs. And Chrome does not need `--shm-size` or
sandbox flags when it is not inside one.

## The invariant

**The deploy must never recreate the machine's state.**

- rsync **without** `--delete` over `.co/`, or sync into a release path and leave state alone
- `.co/keys/`, `.co/logs/` and `.co/evals/` survive every deploy
- restart must not require a rebuild

This is what makes #306 stop being a bug: identity, audit trail and replayable history
persist because nothing removes them. It is also what makes a hand fix over ssh survive.

## Convergence marker

```json
// /srv/<agent>/.co/provision.json
{ "schema": 1, "provisioned_by": "connectonion 1.6.0" }
```

One ssh read decides the path: schema current → straight to rsync and restart; older → run
only the missing steps. Keeps the common deploy fast without letting the server drift.

What runs when:

| | When |
|---|---|
| python, venv, Caddy | once, unless the required version changes |
| systemd unit | on change |
| SSH `authorized_keys` | ensure every deploy — self-heals the only way back in |
| rsync code, `pip install` if `requirements.txt` changed, restart | every deploy |

## Definition of done

- Two consecutive deploys leave the agent's address, `.co/logs/` and `.co/evals/` intact —
  asserted by a test that reads the address before and after
- Editing a file over ssh and deploying again keeps the edit
- A deploy onto a server registered with `co server add` (never provisioned) succeeds
- `pip install` runs only when `requirements.txt` actually changed
- A crashed agent is restarted by systemd without a deploy
- A browser agent survives sustained real page loads with no shm tuning

PR #316, opened 2026-07-30
feat(server): register, list and preflight a deploy target (#311)

### LINKED ISSUE #311: co server: register, list and preflight a deploy target

wu-changxing:
Closes #311. Second link in the #309 chain. No dependencies — it can target a box you made by hand, which is how #312 gets tested before provisioning exists.

## What's here

```bash
co server add prod --ssh aaron@1.2.3.4   # register
co server ls                             # what you can deploy to
co server check prod                     # preflight
```

**`add` takes an ssh target, not a credential.** We shell out to the system `ssh`, so the operator's own agent, `~/.ssh/config`, jump hosts and key selection all keep working — same principle as `co browser` driving a real browser instead of reimplementing one. `~/.co/servers.yaml` holds a name, a target and the last check result. Nothing secret.

**`check` names the requirement that failed.** One probe, one ssh round trip, one `KEY=value` line per fact — so a missing tool reports as an empty value instead of derailing the probe. The list is deliberately short: Ubuntu 24.04, python 3.11+, systemd **the user may actually manage**, ≥5 GB free.

That systemd distinction is worth calling out: `systemd` existing and *being allowed to restart a unit* are different requirements, and a box can pass the first and fail the second. They are reported separately, and a box with systemd but no passwordless sudo is not told "systemd not present" — that would send someone debugging the wrong thing.

## Verification

**`tests/unit/test_server_commands.py` — 23 cases.** The requirement checks are tested by feeding the probe exactly what a box missing each requirement in turn would return, which is what #311's definition of done asks for: wrong distro, wrong Ubuntu version, python absent, python too old, systemd absent, systemd present but units unmanageable, disk too small, `df` unreadable, and several missing at once.

Also asserted:

- **Nothing secret is written.** The saved file is read back and scanned for `private`, `begin openssh`, `password`, `secret`, `token`, `key`.
- **An unreachable host saves nothing** — the registry file is not even created.
- **ssh is invoked with `BatchMode=yes` and a `ConnectTimeout`**, asserted on the actual argv, so it can never hang on a password prompt.

Real run against a hostname that does not resolve:

```
$ time co server add nope --ssh user@does-not-exist.invalid
Cannot reach user@does-not-exist.invalid.
  ssh: Could not resolve hostname ... not known
Nothing was saved. Fix the ssh target and try again.
Tip: your derived key is co keys --ssh — it has to be in authorized_keys on that host.
6.4s total
```

Fails with ssh's own words, in seconds, not a timeout. Most of those 6.4s are Python startup.

Full suite: **2387 passed, 3 skipped**. Confirmed the tests do not touch the real `~/.co/servers.yaml`.

## Depends on

Nothing in code. Pairs with #315 (`co keys --ssh`) in practice — the error message points there, since the derived key has to be in `authorized_keys` for any of this to connect.

### DIFF
--- a/connectonion/cli/commands/server_commands.py
+++ b/connectonion/cli/commands/server_commands.py
@@ -1,0 +1,244 @@
+"""
+Purpose: Register, list and preflight the machines `co deploy --to` can target
+LLM-Note:
+  Dependencies: imports from [subprocess, shutil, pathlib, yaml, rich] | imported by [cli/main.py via handle_server_*] | tested by [tests/unit/test_server_commands.py]
+  Data flow: handle_server_add(name, ssh_target) → _load()/_save() ~/.co/servers.yaml → handle_server_list() renders the table | handle_server_check(name) → _ssh(target, probe script) → parses one KEY=value line per requirement → prints the first failure by name
+  State/Effects: reads and writes ~/.co/servers.yaml (name → ssh target, last check result) | shells out to the system ssh binary | never stores a credential
+  Integration: exposes handle_server_add(), handle_server_list(), handle_server_check(), load_server() for deploy | requirement list is Ubuntu 24.04, python 3.11+, systemd the user may manage, free disk
+  Performance: one ssh round trip per check; the probe is a single command, not one per requirement
+  Errors: an unreachable host fails with ssh's own message and a short timeout, never a hang | a missing requirement is named
+"""
+
+import shutil
+import subprocess
+from pathlib import Path
+from typing import Dict, Optional
+
+import yaml
+from rich.console import Console
+from rich.table import Table
+
+console = Console()
+
+SERVERS_FILE = Path.home() / ".co" / "servers.yaml"
+
+# Deliberately short. "It doesn't work on my box" is only answerable if we
+# stated a narrow promise and `co server check` says which item is missing.
+MIN_PYTHON = (3, 11)
+MIN_FREE_DISK_GB = 5
+SUPPORTED_UBUNTU = "24.04"
+
+# ssh must fail fast rather than hang: an unreachable host is the common case
+# when a name is typed wrong, and a hang looks like a bug in us.
+SSH_TIMEOUT_SECONDS = 10
+
+
+def _load() -> Dict[str, dict]:
+    if not SERVERS_FILE.exists():
+        return {}
+    data = yaml.safe_load(SERVERS_FILE.read_text()) or {}
+    return data.get("servers", {})
+
+
+def _save(servers: Dict[str, dict]) -> None:
+    SERVERS_FILE.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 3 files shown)


wu-changxing:
**Update: verified against a real server, not only mocks.**

Created a throwaway GCE box (Ubuntu 24.04, e2-small, `asia-southeast1-b`) and ran the commands against it.

**`add` accepts a `~/.ssh/config` Host name, not just `user@ip`:**

```
$ co server add test --ssh co-test-deploy
✓ Added test → co-test-deploy

$ co server check test
Checking test (co-test-deploy) …
✓ test is ready to deploy to

$ co server ls
  NAME    TARGET            LAST CHECK
  test    co-test-deploy    ok
~/.co/servers.yaml
```

That is the payoff of shelling out to `ssh` instead of reimplementing it: the Host block carries the hostname, user and IdentityFile, and `co` never had to learn any of them. A registry that stored credentials could not do this.

**The unreachable path, on a real unroutable address:**

```
$ co server add dead --ssh co@10.255.255.1
Cannot reach co@10.255.255.1.
  …
Nothing was saved. Fix the ssh target and try again.
Tip: your derived key is co keys --ssh — it has to be in authorized_keys on that host.

$ co server ls        # 'dead' absent — confirmed in servers.yaml too
```

The probe read the real box correctly — `ubuntu 24.04`, `python 3.12`, `systemd=yes`, `sudo=yes`, `17G` free — matching what a manual ssh reports.

The eight requirement-failure cases stay mock-driven, deliberately: reproducing them for real would mean damaging a box (uninstalling python, revoking sudo) and each one would need a differently-broken machine. The mocks feed the probe exactly what such a box returns, which is the same input the parser sees either way.

PR #317, opened 2026-07-30
Show the agent's Home page before onboarding, not after

### LINKED ISSUE #314: The dashboard is a Home page nobody can see — it is delivered only after the trust gate
## Verified with a controlled test

Same agent, same `dashboard.html`, one variable changed:

| `trust` | Result in oo-chat |
|---|---|
| `"careful"` (the default) | no dashboard — the visitor sees only the default agent card |
| `"open"` | HOME panel renders correctly: agent name, Quick actions, the skill buttons |

So the whole path works — host generates the starter file, pushes `DASHBOARD_SNAPSHOT`, the
relay forwards it, and the frontend renders it in its sandboxed iframe. **The delivery is
simply never reached on the default trust level.**

## Why

`send_dashboard` is called from `ws_router/connect.py:101`, *after* `establish_connection`.
On `careful`, a first-time visitor never gets that far:

```
⚡ ws+ 127.0.0.1 (0 active)
[FAST_RULES] Evaluating client_id=0xb2df62c9094c42eb54...
config={'allow': ['whitelisted','contact'], 'deny': ['blocked'],
        'onboard': {'invite_code': ['OpenOnion'], 'payment': 10}, 'default': 'ask'}
[FAST_RULES] No match, using default=ask
[FAST_RULES] Returning None (needs LLM)
```

The connection stops at the gate, so the snapshot is never sent.

## Why this is a design gap and not just a setting

The dashboard is the agent's **Home page** — the thing meant to greet someone who has not
interacted yet. The starter template offers "Quick actions" precisely as an entry point for
a new visitor. Under the default trust level that visitor must onboard (invite code or
payment) *before* being allowed to see the page whose job is to invite them in.

The content is already public by construction: `featured_skills()` filters to published
project-tree skills specifically because those are the ones carried in the agent's public
profile, with user and builtin skills deliberately excluded as private. So the starter Home
is built entirely from data the agent already publishes to the relay directory — and is then
withheld behind private authorization.

## The tension to resolve

A hand-written `dashboard.html` is agent-authored and may contain anything, so showing an
arbitrary one to strangers could leak. The starter one cannot. That argues for making the
choice explicit rather than picking a side:

- a public Home shown pre-auth and an authenticated dashboard after — two surfaces, or
- one surface with an explicit opt-in, e.g. `host(agent, public_home=True)`, or
- always ship the *generated* starter pre-auth and gate only a file the agent wrote itself

Worth deciding deliberately, because the default currently produces a Home page that no new
visitor can ever load — and nothing reports that it was withheld.

## Note for deploy testing

This is the state a deployed agent is in by default, so it will be hit immediately when
testing #312/#313: the agent is online and reachable and appears to have no dashboard.

## Definition of done

- A first-time visitor on the default trust level either sees a Home page, or is told why
  one is not shown
- The choice between public and gated is explicit in `host()` and documented
- A test covers the pre-auth path, so a future change to the CONNECT sequence cannot
  silently withhold the dashboard again

wu-changxing:
Closes #314.

## The bug

On the default `careful` trust level a first-time visitor never saw the dashboard.
`send_dashboard` only ran inside `establish_connection`, and the trust gate answers
`ONBOARD_REQUIRED` and returns before reaching it — so the page whose whole job is to
invite someone in was withheld until after they had already been let in.

Verified in oo-chat against a real hosted agent, changing one variable:

| `trust` | Before this PR |
|---|---|
| `careful` (default) | no Home panel — just the bare agent card |
| `open` | Home renders correctly |

The host log shows where it stops:

```
[FAST_RULES] No match, using default=ask
[FAST_RULES] Returning None (needs LLM)
```

## Why showing it is the right default

The starter Home is public by construction. `featured_skills()` keeps only project-tree
skills *precisely because* those are the ones already carried in the agent's published
relay profile — user and builtin skills are excluded as private. So the page was built
entirely from data the agent already publishes, and was then gated behind private
authorization.

## Changes

- send the snapshot on the `ONBOARD_REQUIRED` path, **after** that frame so the client
  learns it is gated before any content arrives and cannot read a snapshot as a
  successful connect
- `public_home` in `.co/host.yaml`, default `true`, for the case where a dashboard is
  written for the operator rather than for visitors
- `send_dashboard` stamps `conn`, so the resend from `establish_connection` after a
  successful onboard is skipped as an unchanged file rather than shipping the page twice

## Testing

End to end in the browser against a real relay-connected agent:

| trust | `public_home` | Result |
|---|---|---|
| `careful` | default | Home appears |
| `careful` | `false` | Home hidden |
| `open` | — | Home appears (unchanged) |

Plus four unit tests in `test_dashboard.py` covering the gated path, the opt-out, the
no-file case, and the no-double-send guarantee. Full unit suite: 2391 passed.

### DIFF
--- a/connectonion/cli/commands/server_commands.py
+++ b/connectonion/cli/commands/server_commands.py
@@ -1,0 +1,244 @@
+"""
+Purpose: Register, list and preflight the machines `co deploy --to` can target
+LLM-Note:
+  Dependencies: imports from [subprocess, shutil, pathlib, yaml, rich] | imported by [cli/main.py via handle_server_*] | tested by [tests/unit/test_server_commands.py]
+  Data flow: handle_server_add(name, ssh_target) → _load()/_save() ~/.co/servers.yaml → handle_server_list() renders the table | handle_server_check(name) → _ssh(target, probe script) → parses one KEY=value line per requirement → prints the first failure by name
+  State/Effects: reads and writes ~/.co/servers.yaml (name → ssh target, last check result) | shells out to the system ssh binary | never stores a credential
+  Integration: exposes handle_server_add(), handle_server_list(), handle_server_check(), load_server() for deploy | requirement list is Ubuntu 24.04, python 3.11+, systemd the user may manage, free disk
+  Performance: one ssh round trip per check; the probe is a single command, not one per requirement
+  Errors: an unreachable host fails with ssh's own message and a short timeout, never a hang | a missing requirement is named
+"""
+
+import shutil
+import subprocess
+from pathlib import Path
+from typing import Dict, Optional
+
+import yaml
+from rich.console import Console
+from rich.table import Table
+
+console = Console()
+
+SERVERS_FILE = Path.home() / ".co" / "servers.yaml"
+
+# Deliberately short. "It doesn't work on my box" is only answerable if we
+# stated a narrow promise and `co server check` says which item is missing.
+MIN_PYTHON = (3, 11)
+MIN_FREE_DISK_GB = 5
+SUPPORTED_UBUNTU = "24.04"
+
+# ssh must fail fast rather than hang: an unreachable host is the common case
+# when a name is typed wrong, and a hang looks like a bug in us.
+SSH_TIMEOUT_SECONDS = 10
+
+
+def _load() -> Dict[str, dict]:
+    if not SERVERS_FILE.exists():
+        return {}
+    data = yaml.safe_load(SERVERS_FILE.read_text()) or {}
+    return data.get("servers", {})
+
+
+def _save(servers: Dict[str, dict]) -> None:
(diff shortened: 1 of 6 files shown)


wu-changxing:
Closing — this fixes something that was not broken, and reverses the intended behaviour.

I observed that Home does not appear before onboarding and concluded it was a bug, without testing what happens *after* onboarding. `session.py:78-83` pops the stashed CONNECT on a successful ONBOARD_SUBMIT and calls `establish_connection`, which sends the snapshot — so the existing design already is:

- not onboarded → Home hidden
- enter the invite code once, in chat → Home appears

One invite code for both surfaces, which is the intended behaviour. This PR would have made Home public to anyone holding the address, which is the opposite.

The four tests added here would have locked that wrong behaviour in place. My apologies for the noise.

Issue #313, opened 2026-07-30
co server new: provision a GCE instance, with DNS and TLS


wu-changxing:
Part of #309. Depends on #310 (the SSH public key to install) and #311 (registers the
result). Spans this repo and openonion/oo-api.

## Scope

```bash
co server new prod        # create on GCE, return an ssh target, register it
```

Backend creates the instance and does **only what cannot be done over ssh afterwards**:

1. the instance itself — Ubuntu 24.04, smallest allowlisted machine type
2. write the operator's **derived SSH public key** (#310) into `authorized_keys`
3. an A record for `<name>.agents.openonion.ai` at the instance

**Nothing is installed.** No python, no venv, no Caddy, no systemd unit — `ensure(setup)`
in #312 does all of that on the first deploy.

*Scope corrected per #318.* The original version had the startup script install the
runtime, which contradicted this issue's own closing line and #309's no-branch principle: a
machine registered by hand with `co server add` has no startup script, so that logic has to
exist in `ensure(setup)` regardless. Duplicating it here would mean two implementations of
one thing, reachable only by two differently-created machines — and since `ensure()` is
idempotent by construction (it runs on every deploy), running it on a freshly provisioned
box costs nothing. The duplication buys nothing and risks divergence.

The test that decides future additions, from #318:

> **Can `ensure(setup)` do it over ssh, on a machine we did not create?**
> Yes → it belongs to `ensure()`. No → provisioning.

The three steps above are exactly the ones that fail that test: writing the operator's key
requires the access it grants, and the DNS record needs our zone credentials.

DNS is here rather than in `ensure()` for that reason, and because Caddy needs the name to
exist before it can obtain a certificate. **One name per server, not per agent** — several
agents on one server are separated by path or subdomain at deploy time.

Provisioning ends by handing an ssh target to `co server add`, so everything downstream is
the code from #312 with no branch for "we made this one".

## Backend access

The backend needs to reach the servers it creates. A single long-lived key in every
`authorized_keys` is one credential that opens the whole fleet.

**An SSH CA is the intended answer** — the backend signs a short-lived certificate per
operation and servers trust the CA, so there is no standing credential and revocation does
not mean editing N machines.

If that slips, a static key is acceptable for the first release **only if** it is recorded
here as debt with a follow-up issue. The operator's own derived key stays a plain
`authorized_keys` entry either way: it is their ownership and the backend must not be able
to revoke it.

## Open

[...]

Issue #305, opened 2026-07-29
A deployed agent has no reachable admin — super-admin is the agent's own key, which only the server holds


wu-changxing:
## The part that already works

Worth stating first, because it looks like a problem and isn't. A deployed agent
does **not** inherit the deploying machine's identity:

- `.co/keys/` is in ConnectOnion's built-in `GITIGNORE_CONTENT`, and
  `_load_deploy_ignore_patterns()` seeds from that list regardless of whether the
  project is a git repo — so the private key never enters the tarball on either
  packaging path.
- On the server `host()` does `address.load(co_dir)` → `None` →
  `address.generate()` + `address.save()`.

So every deployed agent already gets its own fresh Ed25519 identity, and one
machine can deploy many agents without them colliding. No change needed there.

## The part that is broken

Nobody can become admin on a deployed agent.

```python
# network/trust/tools.py
def is_super_admin(client_id: str, co_dir: Path = None) -> bool:
    return client_id == get_self_address(co_dir)        # the agent's OWN address

def add_admin(admin_id: str) -> str:
    """Add an admin to ~/.co/admins.txt. Super admin only."""
    admins_file = CO_DIR / "admins.txt"                # CO_DIR = Path.home()/".co"
```

`ADMIN_ADD` is gated on `is_super_admin`, and super-admin is defined as *the agent
itself*. On a deployed agent that address belongs to the keypair generated **on the
server**, whose private key exists only there. The operator who deployed it holds a
different key, so:

- they cannot sign an `ADMIN_ADD` that passes the gate,
- and there is no other path in, because promoting anyone requires an existing admin.

Chicken-and-egg: the only account that could grant admin is one nobody can sign as.

## A second, related defect

`add_admin()` writes `~/.co/admins.txt` — the **global** home directory — while
`is_admin(client_id, co_dir)` reads through `load_admins(co_dir)`, which is
directory-scoped. The write and the read are not guaranteed to be the same file, and
the global write means every agent on one machine shares one admin list. That is the
multi-agent collision people expect from the keypair, hiding one file over.

## Proposal: seed the deployer as admin, the way `ssh-copy-id` does

Treat the deploying identity like a key in `authorized_keys`:

1. **`co deploy` sends the deployer's public address** (from `~/.co/keys/`) with the
   deploy request — the address only, never the private key.
2. **The agent writes it to its own `.co/admins.txt`** when it first generates its
   identity, so a fresh agent starts with exactly one admin: whoever deployed it.
[...]

Issue #302, opened 2026-07-29
User-level skills silently vanish in remote environments — the loader never says so


wu-changxing:
## What happened

`discover_skills()` searches three tiers, in priority order:

1. `.co/skills/` — project
2. `~/.co/skills/` — user
3. builtin

Tier 2 is the one that breaks the moment the agent isn't on your laptop. Verified today against the Claude Code cloud docs: user-level skill directories do not travel to a cloud session — only what is committed to the repository does. The same is true of any deployed agent, a CI runner, or a teammate's checkout.

The loader doesn't fail. It finds fewer skills and says nothing. The agent then behaves differently than it did locally, for a reason nothing in the output explains.

## Why this is worse than a missing file

Most of our own skills live at the user tier by symlink — `~/.co/skills/<name>` pointing into a separate skills repo. That arrangement is invisible to anything that packages or deploys the project, so a skill can look installed, work perfectly for months, and be simply absent everywhere else.

While auditing that skills repo I also found a symlink pointing at its own parent directory. Nothing caught it, because nothing validates what the link tier resolves to.

## Why 1.7

Direct dependency for #147 (*deploying co ai templates with skills*): before we can ship skills with a deploy, we need to know which of them are actually shippable and which only exist on the author's machine.

## Proposal

- **Say which tier each skill came from.** `co status` / `co doctor` list skills as project / user / builtin, so "this one won't travel" is visible before deploy day.
- **Warn when a user-tier skill is about to not travel** — on `co deploy`, and when running somewhere remote (`CLAUDE_CODE_REMOTE`, or the deployed-agent equivalent).
- **Validate links**: a symlinked skill whose target is missing, or which resolves inside itself, is reported rather than silently skipped.
- **Document the fix**: to make a skill travel, it belongs in the repo's `.co/skills/`.

## Definition of done

- Every discovered skill's tier is visible in `co doctor`
- Deploying with user-tier-only skills warns, naming them
- A broken or self-referential skill symlink is reported, not skipped
- Docs state plainly which tier survives a deploy

Issue #301, opened 2026-07-29
Epic: portable skill dependency contracts from install through deploy


wu-changxing:
## What happened

Ran a skill (`make-video`) on a machine that had everything, then looked at what a fresh machine would need. The SKILL.md said nothing about it. The skill needs `ffmpeg`, `ffprobe`, a Python venv, ~395MB of npm packages, and a Chrome binary discoverable at a specific env var. None of that is expressible today.

Frontmatter carries `name` and `description`. That's it. So the failure mode is: the agent reads a skill that looks ready, follows it, and dies partway through on a missing binary — after doing real work.

## Why 1.7

#147 is *Support deploying co ai templates with skills*. Shipping a skill to a machine you don't control means knowing what that machine has to have. Right now the deploy has no way to ask and the skill has no way to answer.

## Proposal

Let a skill declare prerequisites in frontmatter, and check them before the agent commits to the workflow:

```yaml
---
name: make-video
description: ...
requires:
  commands: [ffmpeg, ffprobe, node, python3]
  python: requirements.txt
  node: remotion/package.json
  env: [GEMINI_API_KEY]
  setup: scripts/bootstrap.sh
---
```

- **Check before starting, not midway.** `skill(name)` verifies `requires` and returns what's missing *with the command to fix it*, instead of a traceback three steps in.
- **`co doctor` reports it** for every discovered skill (related: #213).
- **`co deploy` uses it** to build the remote image — the direct link to #147.

Keep it optional: a skill with no `requires` behaves exactly as today.

## Not in scope

Auto-installing anything. The skill states its needs; a human or an explicit setup step satisfies them. Silently running installers because a markdown file asked is worse than the current failure.

## Definition of done

- A skill with unmet `requires` is reported before its workflow starts, naming the missing item and how to get it
- A skill without `requires` is unaffected
- `co doctor` lists unmet prerequisites per skill
- Tests cover: missing command, missing env var, satisfied, and absent-`requires`

wu-changxing:
## Delivery breakdown

This issue is now the architecture/acceptance epic. Implementation should land as small PRs that close these child issues:

- [ ] #470 — Define a versioned dependency manifest for skills
- [ ] #471 — Validate skill requirements in co doctor, skills install, and co ai startup
- [ ] #472 — Package and install declared skill runtime dependencies during deploy

A child can merge independently when its own definition of done passes. Close this epic only after every required child is complete or an explicit architecture decision removes it from scope.

[...]

PR #320, opened 2026-07-30
feat(deploy): co deploy --to converges a server, syncs code, restarts the unit (#312)

### LINKED ISSUE #312: co deploy --to: converge the server, sync the code, restart the unit

wu-changxing:
Closes #312, and with it #305 and #306. The largest piece of #309.

```
co deploy --to prod  =  ensure(setup) → sync code → restart
```

No container. A directory, a venv, a systemd unit. `Restart=always` for crash recovery, `enable` for start-on-boot, journald for logs — and Chrome needs no `--shm-size` when it is not inside one.

## The invariant, and how it is enforced

**A deploy must never recreate the machine's state.**

```
rsync -az --delete --exclude .co/ --exclude .venv/ --exclude .git/
```

`--delete` so code removed locally goes away; `.co/` excluded so identity, logs and evals survive. **That pairing is the whole design** — it is what stops a redeploy from reissuing the agent's address (#306) and what lets a fix made by hand over ssh survive (#305).

Four tests guard it directly, including one asserting the setup script never redirects into `.co/`.

## Verified end to end on a real server

Not mocks. A live Ubuntu 24.04 GCE box, reached with the key derived in #315.

**First deploy:**

```
deploytest → test (co-test-deploy)
  converging server …
  syncing code …
  installing dependencies …
  writing systemd unit …
  restarting …
✓ deploytest is running on test
```

**Then planted the state a deploy must not touch** — an agent key, an audit log, and a file edited by hand:

```
/srv/deploytest/.co/keys/agent.key   MY-AGENT-IDENTITY-12345
/srv/deploytest/.co/logs/agent.log   important audit log
/srv/deploytest/.co/hand-edit.txt    HAND EDIT BY OPERATOR
```

**Changed the code and deployed again:**

```
  converging server …
  syncing code …
  restarting …
```

Note what is missing: no dependency install, no unit write. `requirements.txt` was unchanged and the unit content was identical, so both were skipped.

**The result — the definition of done for #312:**

```
=== new code live? ===
version=2-UPDATED

=== state survived? ===
agent identity:  MY-AGENT-IDENTITY-12345
audit log:       important audit log
hand edit:       HAND EDIT BY OPERATOR

=== service ===
active
```

**Crash recovery, without a deploy:**

```
$ kill -9 <MainPID>
6 seconds later: active
new PID differs, NRestarts: 1
```

## Design decisions worth reviewing

**`ensure()` is keyed on `.co/provision.json`, not on "did we provision this box".** One ssh read decides the path: current schema → straight to rsync and restart; older → run only the missing steps. This is what makes a hand-registered machine need no separate code path — its first deploy is the setup, which is #309's no-branch requirement and the reason #313's scope was corrected to install nothing.

**`authorized_keys` is re-asserted on every deploy, not once.** It is the only way back into the machine, so it self-heals rather than needing a rescue path. Appended with `grep -qxF` so it does not accumulate duplicates.

**The restart is verified with `systemctl is-active`, not the exit code.** `systemctl restart` returns 0 for a unit that starts and immediately dies — trusting it would report a broken deploy as a success, which is the worst outcome a deploy command can have. On failure it prints journald, because the traceback is what the operator needs.

**`--to` refuses to combine with `--template`, `--skills` or `--name`.** Those belong to the cloud path. Silently ignoring them would let someone believe their skills were bundled when they were not.

## Tests

`tests/unit/test_deploy_to_server.py` — **27 cases**, organised around the invariant: the rsync exclusions, setup skipped at the current schema, `authorized_keys` ensured anyway, corrupt marker reading as schema 0, pip skipped on an unchanged hash, unit not rewritten when identical, a unit that dies reported as failure, journald surfaced, and `provision.json` written **only** after every step succeeded.

Full suite: **2425 passed, 3 skipped**.

## Not covered

The last item in #312's definition of done — "a browser agent survives sustained real page loads with no shm tuning" — is untested. It needs a browser agent deployed and driven for a while, which is a separate exercise from this code. The claim it rests on is structural (there is no container capping `/dev/shm` at 64 MB), but structural is not measured.

### DIFF
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -1,0 +1,350 @@
+"""
+Purpose: Deploy an agent onto a server you own — converge the setup, sync the code, restart the unit
+LLM-Note:
+  Dependencies: imports from [hashlib, json, shlex, subprocess, pathlib, yaml, rich, server_commands, project_cmd_lib] | imported by [cli/main.py via handle_deploy_to] | tested by [tests/unit/test_deploy_to_server.py]
+  Data flow: handle_deploy_to(server, project_dir) → load_server() resolves the ssh target → _read_provision() reads /srv/<agent>/.co/provision.json → _ensure_setup() runs only the missing steps → _sync_code() rsyncs excluding .co/ → _install_deps() only when requirements.txt changed → _write_unit() on change → systemctl restart
+  State/Effects: on the server creates /srv/<agent>/{,.venv,.co} and /etc/systemd/system/<agent>.service | rsyncs code | NEVER writes inside /srv/<agent>/.co except provision.json | locally reads .co/host.yaml and the project files
+  Integration: exposes handle_deploy_to(server, project_dir) | requires a server registered by `co server add` (#311) reachable with the key from `co keys --ssh` (#310)
+  Performance: one ssh round trip to read provision.json, one rsync, one restart | setup steps are skipped entirely once the schema matches
+  Errors: an unreachable or unprepared server fails before anything is changed | a failed unit start surfaces journalctl output rather than a bare exit code
+"""
+
+import hashlib
+import json
+import shlex
+import subprocess
+from pathlib import Path
+from typing import Optional
+
+import yaml
+from rich.console import Console
+
+from .server_commands import SSH_TIMEOUT_SECONDS, load_server
+
+console = Console()
+
+# Bumped when a setup step is added or changes shape. A server reporting an
+# older schema runs the missing steps; one reporting this schema goes straight
+# to rsync and restart.
+PROVISION_SCHEMA = 1
+
+SRV = "/srv"
+
+
+def _ssh(target: str, command: str, timeout: int = 300) -> subprocess.CompletedProcess:
+    """Run one command on the server. Longer default timeout than a preflight:
+    apt and pip are slow, and killing them halfway leaves a half-installed box."""
+    return subprocess.run(
+        [
+            "ssh",
+            "-o", "BatchMode=yes",
+            "-o", f"ConnectTimeout={SSH_TIMEOUT_SECONDS}",
+            "-o", "StrictHostKeyChecking=accept-new",
+            target,
+            command,
+        ],
+        capture_output=True,
+        text=True,
+        timeout=timeout,
+    )
+
+
+def _read_project(project_dir: Path) -> Optional[dict]:
+    """Read the agent name and entrypoint from .co/host.yaml.
+
+    Same contract as the cloud deploy, so a project does not need a second
+    config to be deployable both ways.
+    """
+    host_yaml = project_dir / ".co" / "host.yaml"
+    if not host_yaml.exists():
+        console.print("[red]Not a ConnectOnion project. Run 'co init' first.[/red]")
+        return None
+
+    config = yaml.safe_load(host_yaml.read_text(encoding="utf-8")) or {}
+    from .project_cmd_lib import DEPLOY_NAME_PATTERN, normalize_deploy_name
+
+    name = str(config.get("name") or "unnamed-agent")
+    if not DEPLOY_NAME_PATTERN.match(name):
+        console.print(f"[red]Invalid agent name: {name}[/red]")
+        # The name becomes a directory and a systemd unit name, so the same rule
+        # the cloud path enforces applies here.
+        suggestion = normalize_deploy_name(name)
+        if suggestion:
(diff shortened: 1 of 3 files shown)


wu-changxing:
## Self-review found a real bug, now fixed in 98730ce

**Excluding all of `.co/` meant skills never reached the server.**

A project's skills live in `.co/skills/`, and the cloud deploy path bundles them (that is what `--skills` does). This path excluded the whole directory, so `co deploy --to` shipped an agent with none of its skills, reported success, and ran it that way. **A deploy that quietly drops what the agent is made of is worse than one that fails** — the agent starts, answers, and is simply less capable than the operator believes.

Confirmed on the live box before fixing:

```
$ ls /srv/deploytest/.co/skills
ls: cannot access '/srv/deploytest/.co/skills': No such file or directory
```

### The fix, and why it is fiddly

Skills are **what the agent is**, not state it accumulated, so they belong in the sync. But the state protection has to survive unchanged:

```
--include .co/  --include '.co/skills/***'  --exclude '.co/*'
```

Two things that are easy to get wrong:

- The exclusion has to be `.co/*`, not `.co/`. Excluding the directory itself prunes it before rsync ever descends, so the skills include can never match.
- rsync applies the **first** matching rule, so the includes must precede the exclude.

### Verified on the live box, both halves at once

Added a skill locally and deployed. The server has state files that do **not** exist locally, so if `--delete` escaped the skills subtree they would be destroyed:

```
=== skill arrived? ===
This skill must reach the server.

=== state survived? ===
agent identity:  MY-AGENT-IDENTITY-12345
audit log:       important audit log
hand edit:       HAND EDIT BY OPERATOR
heartbeat:       version=2-UPDATED

=== service ===
active
```

Both hold: the skill arrived, nothing was deleted.

### Tests

Two added, plus one corrected. The rule is order-sensitive and reads like something to tidy into a single blanket exclusion, so the tests assert the **sequence** (`argv.index("--include") < argv.index("--exclude")`) and explicitly that `("--exclude", ".co/")` is *not* present. One pre-existing test asserted the old blanket pattern and now asserts `.co/*` — it failed when the fix landed, which is the behaviour I want from it.

`tests/unit/test_deploy_to_server.py` is 29 cases. Full suite: **2427 passed, 3 skipped**.

### Relation to #301 and #302

This is not those. #302 is about skills that are *symlinks* into another repo — rsync follows or skips those depending on flags, and this change does not address it. #301 is a skill declaring what it needs to run. Both remain open and both now matter more, because skills actually reach the server.

PR #294, opened 2026-07-29
co-ai: make the agent general, ship one template, and stop treating an unanswered confirmation as approval

wu-changxing:
## Description

`main.md` opened with "You are a coding agent" and followed it with five sections of code-editing doctrine — and `assemble_prompt()` loads `main.md` unconditionally. So every deployed agent, including one shipped as a poster or a support bot, was told it was a software engineer before it ever saw the skill it was deployed with.

This splits **how to work** from **what to work on**, rewrites the generic half to what a general agent actually needs, and collapses six templates into one.

Four bugs surfaced from running and building the result rather than reading it — one of them would have broken the next release for everyone. All fixed here.

## Related Issue

Fixes #292

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

Breaking: five `--template` names are gone (see *Breaking changes* below). `co ai` behaviour is unchanged.

### The prompt split

- `prompts/main.md` is now domain-neutral. `assemble_prompt(..., role=)` appends `roles/{role}.md` on top; an unknown role raises with the list of valid ones rather than silently serving a prompt with no domain.
- `roles/coding.md` carries the coding identity and doctrine. `co ai` passes `role="coding"`, so its assembled prompt is the same content it always had. Verified by probing all 19 distinct concepts from the old `main.md` against the new file set — every one survives.
- "Remote Login Requests" moved from `main.md` to `browser.md`, where its subject already lived. An agent that only posts to a website still needs to log in, and shouldn't have to be a coding agent to know how. Fixed a paragraph there that had been split mid-sentence by an inserted block.
- `create_coding_agent()` → `create_agent()`. Old name kept as an alias.

### Two sections the prompt was missing entirely

**Executing Actions with Care.** The old `Security` section was 16 lines on SQL injection, XSS, and path traversal — the safety of code the agent *writes*. For an agent that can post publicly, send mail, and deploy, that is the wrong half of "security", and there was nothing at all about the consequences of what the agent *does*. That section moved to `roles/coding.md`; confirm-before-irreversible-or-outward-facing is now in the shared prompt.

**Delivering Work** — scope discipline, finishing the whole task, and reporting what actually happened. `Persistence` ("try your best, don't give up") was already there with no counterweight. An agent pushed to be autonomous, with nothing telling it to be honest about failure, is being pushed toward claiming success it does not have.

[...]

PR #321, opened 2026-07-30
fix: an unanswered confirmation is not approval (split from #294)

wu-changxing:
Split out of #294 so the security fix can ship without waiting on that PR's breaking template removal. Everything here is independent of the template work and applies to `main` as it stands.

Milestone: **1.5.3**.

## 1. An unanswered confirmation was being read as approval

`ask_user` with no `io` attached returned:

```
No interactive input available (one-shot mode); decide from the request context.
```

The model read that as permission. **An agent that correctly stopped to confirm an irreversible or outward-facing action was told to go ahead anyway.**

Reproduced on `gemini-3.6-flash` (credit to #294, which found it): in a folder containing a script that posts to the company `#general` channel, prompted *"Announce the v2 release to the company."* The agent did everything right — read the changelog, read the script, recognised it was outward-facing, drafted the message, called `ask_user` with Yes/No. Then the tool answered its own question and it posted. **2/2 before, 0/3 after.**

**This is not only one-shot mode.** Every deployed agent served over plain `POST /input` has no `io` either — precisely the population that cannot take back a bad send.

The replacement says what it means and names the actions it gates:

```
NOT ANSWERED — nobody is available to reply. This is not approval. Do not send,
post, publish, delete, overwrite, deploy, or spend anything that this question
was gating. Continue with any part of the task that does not depend on the
answer, then end your turn by stating the question and what you would have
done, so the user can decide.
```

It does not just refuse — it says what to do instead, so the agent still makes progress on the reversible part rather than stalling.

## 2. Prompts taught two templates that never existed

`prompts/connectonion/agent-design.md` instructed:

```bash
co create my-browser --template playwright
co create my-emailer --template email-agent
```

Neither `playwright` nor `email-agent` has ever existed in this tree. The templates are `browser, co-ai, coder, hosted-browser, minimal, web-research`. An agent following its own always-loaded prompt ran a command that could not work.

Fixed, and a test now asserts every `--template` named anywhere in the prompts resolves to a directory that exists — so prompts cannot drift from the templates again.

## 3. `co create --template nope` printed an error and exited 0

```
$ co create x --template playwright; echo $?
❌ Template 'playwright' not found!
0
```

A script asking for a template it misspelled got **success and an empty directory**. Now:

```
$ co create x --template playwright; echo $?
❌ Template 'playwright' not found!
   Available: browser, co-ai, coder, hosted-browser, minimal, web-research, custom
1

$ co create good --template minimal; echo $?
0
```

Verified both paths, and that no directory is left behind on failure.

## 4. A dangling symlink in the package breaks the wheel build silently

`prompts/connectonion/` is a tree of symlinks mirroring `docs/` — that is how `load_guide` serves documentation without a second copy. hatchling `os.stat()`s every included file, so **one dangling link fails the build**:

```
FileNotFoundError: .../prompts/connectonion/templates/meta-agent.md
ERROR Backend subprocess exited when trying to invoke build_wheel
```

The trap is that `pip install -e .` archives nothing, so it keeps working and the ordinary test jobs stay green. **The release breaks for every user while seven of eight CI jobs pass.**

`tests/unit/test_packaging.py` asserts no dangling symlinks, in the fast suite, on every platform. It passes on `main` today — this is a guard, not a fix. #294 is where it would have earned its keep.

## What stayed in #294

The prompt restructure (`main.md` split into a domain-neutral half plus `roles/coding.md`), the six-to-one template collapse, and the two tests that only make sense once templates are retired (`test_prompts_do_not_teach_a_template_that_no_longer_exists`, `test_guide_index_does_not_advertise_the_retired_templates`). Those are a breaking change plus a coordinated docs-site update across 16 places, which is a separate decision from shipping this.

## Tests

`tests/unit/test_ask_user.py` extended, `tests/unit/test_packaging.py` added (2 cases). Full suite: **2438 passed, 3 skipped**.

### DIFF
--- a/connectonion/cli/co_ai/prompts/connectonion/agent-design.md
+++ b/connectonion/cli/co_ai/prompts/connectonion/agent-design.md
@@ -156,8 +156,7 @@ co create my-agent
 cd my-agent
 
 # With template
-co create my-browser --template playwright
-co create my-emailer --template email-agent
+co create my-browser --template browser
 co create my-researcher --template web-research
 ```
 
--- a/connectonion/cli/commands/create.py
+++ b/connectonion/cli/commands/create.py
@@ -257,9 +257,14 @@ def handle_create(name: Optional[str], ai: Optional[bool], key: Optional[str],
     template_dir = cli_dir / "templates" / template
 
     if not template_dir.exists() and template != 'custom':
+        # Naming a template that does not exist printed this and exited 0, so a
+        # script that asked for one got "success" and an empty directory. Say
+        # what is available and let the caller fail.
+        available = sorted(p.name for p in (cli_dir / "templates").iterdir() if p.is_dir())
         console.print(f"[red]❌ Template '{template}' not found![/red]")
+        console.print(f"   [dim]Available: {', '.join(available + ['custom'])}[/dim]")
         shutil.rmtree(project_dir)
-        return
+        return False
 
     # Copy template files
     files_created = []
--- a/connectonion/cli/main.py
+++ b/connectonion/cli/main.py
@@ -125,7 +125,11 @@ def create(
 ):
     """Create new project."""
     from .commands.create import handle_create
-    handle_create(name=name, ai=None, key=key, template=template, description=description, yes=yes)
+    # An explicit False means the project was not created — exit non-zero so a
+    # script can tell. Other return values keep the previous behaviour.
+    if handle_create(name=name, ai=None, key=key, template=template,
+                     description=description, yes=yes) is False:
+        raise typer.Exit(1)
 
 
 @app.command()
--- a/connectonion/useful_tools/ask_user.py
+++ b/connectonion/useful_tools/ask_user.py
@@ -31,8 +31,19 @@ def ask_user(
         The user's answer (or comma-separated answers if multi_select)
     """
     if not agent.io:
-        # One-shot mode (co ai "prompt") runs without io — let the model decide.
-        return "No interactive input available (one-shot mode); decide from the request context."
+        # One-shot mode (co ai "prompt", and every deployed agent) has nobody to
+        # answer. This used to say "decide from the request context", which read
+        # as approval: an agent that correctly stopped to confirm an irreversible
+        # or outward-facing action was told to go ahead anyway. Unanswered is not
+        # yes — proceed with reversible work, report back on the rest.
+        return (
+            "NOT ANSWERED — nobody is available to reply. This is not approval. "
+            "Do not send, post, publish, delete, overwrite, deploy, or spend "
+            "anything that this question was gating. Continue with any part of "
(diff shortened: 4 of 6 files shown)


wu-changxing:
Split makes sense — the `ask_user` fix should not have been sitting behind a breaking change. Extraction checks out against what was in #294, and two of your versions are better than mine:

- **`main.py` converting `False` → `typer.Exit(1)`** is the right shape. I had `raise typer.Exit(1)` inside `handle_create`, which forced `_deploy_template_project` to wrap the call in `try/except typer.Exit` because deploy's contract is a bool. Yours keeps CLI-exit concerns in the CLI layer and that try/except disappears. I'll adopt it on rebase and drop mine.
- **`test_prompts_only_name_templates_that_exist`** resolves `--template` names against the actual templates directory, where mine hardcoded the retired list. Yours keeps working as templates change; mine needs editing. On rebase I'll delete `test_prompts_do_not_teach_a_template_that_no_longer_exists` — once #294 collapses templates to one, your test covers that case automatically and having both is just a list to maintain.

Two things I think belong here rather than in #294, by your own "independent of the template work" line. Happy to push either to this branch or hand you a patch — your call, I don't want to touch your PR uninvited.

**1. The exit-0 fix has no regression test.** The PR verifies it by hand. That is the one behaviour here a future refactor is most likely to undo silently — `handle_create` has several other `return` paths that still mean failure and still exit 0 (invalid project name, directory exists, custom template with no API key), so the next person touching this reasonably reads `return` as the house style. Something like:

```python
def test_unknown_template_exits_nonzero(tmp_path, monkeypatch):
    from typer.testing import CliRunner
    from connectonion.cli.main import app

    monkeypatch.chdir(tmp_path)
    result = CliRunner().invoke(app, ["create", "x", "--template", "nope", "--yes"])

    assert result.exit_code == 1
    assert not (tmp_path / "x").exists()
```

**2. The check still runs after the network call.** `handle_create` resolves the template at the top but only validates it ~200 lines later — after `authenticate()` has made its round trip and after `project_dir.mkdir()`. So the failing path in this PR is:

```
co create x --template nope
  → authenticate()      2-5s over the network
  → mkdir project_dir
  → "Template 'nope' not found!"
  → rmtree
  → exit 1
```

The name is knowably wrong before any of that. Moving the check to where the template is resolved takes it to ~1.9s with nothing written:

```python
    if not template:
        template = 'minimal'

    if template != 'custom' and not (TEMPLATES_DIR / template).exists():
        console.print(...)
        return False
```

This is independent of the template work in the same way item 3 already is — it is wrong today with the current template set, and it gets more visible, not less, once five names retire and every stale script lands on this path. Doing it here also avoids a second round of conflicts in `create.py` on my rebase.

Otherwise: agreed on what stays in #294, and agreed the template retirement is Aaron's call given the 16 docs-site references. I'll rebase once this lands — overlap is exactly the five files you listed.

---

PR #403, opened 2026-07-31
fix: report malformed trust policy YAML clearly

### LINKED ISSUE #381: Malformed YAML in a trust policy file crashes the whole agent host instead of failing safely
## Bug

`connectonion/network/trust/fast_rules.py:44` — `yaml.safe_load(yaml_content)` inside `parse_policy()` has no exception handling. A `yaml.YAMLError` propagates straight up through `TrustAgent._load_policy()` (`trust_agent.py:131-149`) into `TrustAgent.__init__`, uncaught.

## Failure scenario

An operator points `host(..., trust="path/to/policy.md")` at a hand-edited policy file with a YAML typo in its frontmatter (e.g. `allow: [whitelisted, contact` — an unclosed bracket). Instead of a clear config error or a safe fallback, the entire agent host process crashes at startup (or at first request, if `TrustAgent` is constructed lazily). A one-character typo in a markdown policy takes down the whole host — a real availability bug, not cosmetic.

(Note: missing frontmatter entirely is already handled gracefully — `parse_policy` returns `({}, policy_text)` in that case. This is specifically about *malformed* frontmatter, not absent frontmatter.)

## Fix direction

Catch `yaml.YAMLError` in `parse_policy()` and either fall back to a safe default trust level with a loud warning, or raise a clear, specific "malformed policy file" error at `host()` startup rather than letting a raw YAML parser exception surface.

## Definition of done

- A policy file with malformed YAML frontmatter produces a clear, actionable error (or safe fallback) instead of an uncaught `yaml.YAMLError`
- A test covering malformed frontmatter (distinct from the already-covered missing-frontmatter case)

Roovelrz:
## Description

Handle malformed YAML frontmatter in trust policy files as a clear configuration error instead of exposing an uncaught `yaml.YAMLError`.

The new error includes available YAML line and column information and tells the operator to fix the policy before starting the agent host. Existing behavior for policies without frontmatter and policies with empty frontmatter remains unchanged.

## Related Issue

Fixes #381

## Type of Change

- [x] Bug fix

## Changes Made

- Catch `yaml.YAMLError` inside `parse_policy`
- Raise an actionable `ValueError` with YAML location and parser details
- Suppress the raw PyYAML traceback with `from None`
- Add a distinct unit test for malformed YAML frontmatter
- Update the module error-handling note
- Preserve existing missing-frontmatter and empty-frontmatter behavior

## Testing

- [x] `python -m pytest tests/unit/test_fast_rules.py::TestParsePolicy::test_parse_malformed_yaml_raises_actionable_error -q`
- [x] `python -m pytest tests/unit/test_fast_rules.py -q`
- [x] Manually verified that malformed YAML raises an actionable `ValueError`
- [x] Manually verified that valid, missing, and empty frontmatter behavior is unchanged
- [x] Verified `fast_rules.py` targeted coverage at 95 percent
- [x] Ran `git diff --check`

### Local Environment Note

Full unit and non-real API test suites were attempted but do not complete cleanly on this Windows host because of existing browser, symbolic-link permission, and CLI environment failures.

These failures are outside `connectonion/network/trust/fast_rules.py` and `tests/unit/test_fast_rules.py`. The affected fast-rules test file passes all 29 tests locally.

## Example Usage

A malformed policy such as:

```yaml
---
allow: [whitelisted, contact
---
```

now produces an error similar to:

```text
ValueError: Malformed YAML frontmatter in trust policy at line 1, column 29: expected ',' or ']', but got '<stream end>'. Fix the YAML syntax before starting the agent host.
```

instead of exposing an uncaught PyYAML parser exception.

## Checklist

- [x] My code follows the existing style of the affected files
- [x] I have performed a self-review
- [x] I have added a docstring entry for the new error behavior
- [x] My changes generate no new warnings in the affected test suite
- [x] New and existing fast-rules unit tests pass locally
- [x] No project dependencies were added
- [x] No public documentation changes are required

## Screenshots

Not applicable.

## Additional Notes

This change intentionally fails fast instead of silently falling back to another trust configuration. Continuing with an unintended policy could change security behavior without the operator noticing.
```

### DIFF
--- a/connectonion/network/host/server.py
+++ b/connectonion/network/host/server.py
@@ -76,14 +76,15 @@ def _parse_trust_config(trust: Union[str, "Agent"]) -> dict | None:
     if trust.lower() in TRUST_LEVELS:
         policy_path = PROMPTS_DIR / f"{trust.lower()}.md"
         if policy_path.exists():
-            config, _ = parse_policy(policy_path.read_text(encoding='utf-8'))
+            config, _ = parse_policy(policy_path.read_text(encoding='utf-8'),
+                                     source=str(policy_path))
             return config
         return None
 
     # Check if it's a file path
     path = Path(trust)
     if path.exists() and path.is_file():
-        config, _ = parse_policy(path.read_text(encoding='utf-8'))
+        config, _ = parse_policy(path.read_text(encoding='utf-8'), source=str(path))
         return config
 
     # Inline policy text
--- a/connectonion/network/trust/fast_rules.py
+++ b/connectonion/network/trust/fast_rules.py
@@ -24,12 +24,18 @@ from typing import Optional
 from .tools import is_whitelisted, is_blocked, is_contact, is_admin, promote_to_contact
 
 
-def parse_policy(policy_text: str) -> tuple[dict, str]:
+def parse_policy(policy_text: str, source: str = None) -> tuple[dict, str]:
     """
     Parse YAML frontmatter from markdown policy file.
 
     Returns:
         (config_dict, markdown_body)
+
+    Args:
+        policy_text: The policy file's contents.
+        source: Where it came from, named in the error. A line number is only
+            half an answer when an operator has several policies and the
+            built-in ones as well.
 
     Raises:
         ValueError: If the YAML frontmatter is malformed.
@@ -57,8 +63,9 @@ def parse_policy(policy_text: str) -> tuple[dict, str]:
         )
         detail = f': {problem}' if problem else ''
 
+        where = f' in {source}' if source else ' in trust policy'
         raise ValueError(
-            f'Malformed YAML frontmatter in trust policy{location}{detail}. '
+            f'Malformed YAML frontmatter{where}{location}{detail}. '
             'Fix the YAML syntax before starting the agent host.'
         ) from None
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by wu-changxing** [comment]
Thanks for this — you found something real, and the part you identified as missing is the part we shipped.

We went a different way with it in #422, and I want to be clear about why, because it is not a criticism of the diagnosis.

## What we kept

Your read of #381 was right: the error was unhelpful. `yaml.safe_load` on a bare string produces

```
in "<unicode string>", line 1, column 8
```

and an operator with policies under `.co/trust/` plus the built-in ones has no way to know which file to open. That is the actual gap, and it is what #422 fixes.

## What we changed

PyYAML names the stream it is handed. Giving it a named stream instead of a string gets the file into the error with nothing caught:

```
while parsing a flow sequence
  in ".co/trust/custom.md", line 3, column 8
      allow: [whitelisted, contact
             ^
expected ',' or ']', but got '<stream end>'
```

Three reasons we preferred that to catching and re-raising:

1. **Both versions crash the host.** Nothing degrades and nothing recovers either way, so the change is to the wording rather than to the behaviour — and #381's "fails safely" framing invites the reading that something now survives, which is not the case in either patch.
2. **This codebase avoids `try/except` deliberately.** `CLAUDE.md` says so directly: let it crash and show the error. A wrapper that converts one fatal error into another fatal error is the case that rule is aimed at.
3. **`from None` discards a better message than the one it builds.** The parser quotes the offending line and points at the column. The replacement keeps the line and column as numbers and drops the rest.

## One thing to flag

While reviewing I pushed a commit to your branch (`4dc98fa`, adding the filename to the message) before settling on the stream approach. Sorry for the noise — it is on `fix/malformed-trust-policy-yaml` in your fork and you can drop it.

Closing in favour of #422. The credit for spotting it is yours.

Issue #393, opened 2026-07-31
co deploy --to rsyncs the operator's personal OAuth refresh tokens to the server in plaintext


wu-changxing:
## What happened

`co deploy --to <server>` rsyncs a copy of the user's personal OAuth refresh tokens onto
the server, in plaintext, and never mentions it.

Two facts combine.

**1. A project's `.env` is a full copy of the user's global credentials.**

```python
# Create .env file - copy from global keys.env
env_path = project_dir / ".env"

# Always copy from global keys.env (includes AGENT_ADDRESS, AGENT_EMAIL, and API keys)
if global_keys_env.exists() and global_keys_env.stat().st_size > 0:
```
`cli/commands/create.py:313-317`

On a real project here, that copy is 17 keys:

```
GOOGLE_REFRESH_TOKEN        MICROSOFT_REFRESH_TOKEN
GOOGLE_ACCESS_TOKEN         MICROSOFT_ACCESS_TOKEN
GOOGLE_EMAIL                MICROSOFT_EMAIL
OPENONION_API_KEY           AGENT_EMAIL   AGENT_ADDRESS   …
```

Refresh tokens for the operator's personal Gmail and Outlook — not project scoped, not
short lived.

**2. The `--to` path has no concept of a secret.**

`deploy_to_server.py` does not contain the string `.env` anywhere. `_sync_code()` excludes
exactly four things:

```python
"--exclude", ".co/*", "--exclude", ".venv/",
"--exclude", ".git/", "--exclude", "__pycache__/",
```
`cli/commands/deploy_to_server.py:369-374`

`.env` is not among them, so it ships as ordinary source — as do `.env.local`,
`.env.production`, `*.pem`, `credentials.json`, and anything else a `.gitignore` was
holding back. rsync does not read `.gitignore`.

It then works by accident: systemd sets `WorkingDirectory=/srv/<agent>`, and
`connectonion/__init__.py` calls `load_dotenv()` on the cwd. The agent gets its keys
because a file happened to land next to it.

## Why this is a bug and not a convenience

The other deploy path teaches the opposite, in both code and prose. `co deploy` parses
`.env` with `dotenv_values()`, uploads the pairs as `secrets`, and keeps the file out of
the package; the docs say plainly that *"Local-only files such as `.env` … are skipped."*

So one verb, two paths, and the same file is a secret on one and source on the other.
A user who has read the documentation has every reason to believe `.env` stays on their
laptop.

## Three further consequences

[...]

Issue #324, opened 2026-07-30
The agent never calls the skill tool on its own — the skill list is injected with nothing that tells it to


wu-changxing:
## What happened

A project skill sat at `.co/skills/contract-ledger/SKILL.md`. Its description names the
trigger phrases explicitly — "use when the user says 整理合同 / 更新台账 / 补录合同 / 合同对账".
Then I asked for exactly that, in the agent's own words:

```
co ai "帮我整理一下合同台账，先告诉我第一步该做什么"
```

The agent never called `skill()`. It went looking for files instead:

```
▸ glob(pattern="*contract-ledger*")        ✓ 0.00s   → 0 results
▸ glob(pattern="**/*skill*/**")            ✓ 0.01s
▸ bash(command="find ~/.co ~/.co...")      ✓ 1.4s
▸ bash(command="grep -rn "contra...")      ✓ 120.0s  ← two minutes
▸ bash(command="ls -la ~/.co/ski...")      ✓ 0.03s
```

One `grep -rn` burned **120 seconds** and still never produced the skill content.

Model: `gemini-3.6-flash`, the co-ai default. Reproduced with three different phrasings
of the same intent. Told explicitly — `co ai "use the skill tool to load contract-ledger"` —
it fires on the first try. **The tool works. The agent just doesn't know when to reach for it.**

## Why

`get_skills_for_prompt()` injects a bare list. No instruction attached:

```python
lines = ["<available_skills>"]
for name, info in SKILLS_REGISTRY.items():
    lines.append(f'  <skill name="{name}" description="{desc}"/>')
lines.append("</available_skills>")
```

And `grep -ri skill prompts/` returns exactly one hit — an unrelated mention in
`connectonion/concepts/trust.md`. Nothing in the system prompt ever says *when a task
matches a skill's description, call `skill(name=...)` first*.

The only place that sentence exists is the docstring of `skills/tool.py`, sharing space
with 23 other tool descriptions:

> When you recognize a task matches a skill's description, call this tool to load the full instructions.

So the model has to independently connect "there is an `<available_skills>` list in my
system prompt" with "there is a `skill()` entry in my tool list." Frontier models manage it.
Flash-tier models — the default here — do not.

## What makes it worse: project skills drown in user skills

Measured in one project:

| Source | Count |
|---|---|
| `~/.co/skills/` (user) | 57 |
| builtin | 4 |
| `.co/skills/` (project) | **1** |

[...]

PR #417, opened 2026-07-31
fix(deploy): carry the config the author wrote, delete nothing the server owns

### LINKED ISSUE #392: Deploying to a server silently drops host.yaml, OO.md and commands/ — the .co exclusion defaults to losing config
## What happened

`_sync_code()` excludes `.co/` and re-includes one path:

```python
"--include", ".co/",
"--include", ".co/skills/***",
"--exclude", ".co/*",
```
`cli/commands/deploy_to_server.py:369-372`

That exclusion is correct and deliberate — it is the #306 fix, and it is what keeps
`.co/keys/` from being overwritten. The problem is the shape of the rule, not the intent:
the default is **exclude**, so every file under `.co/` that an author writes has to be
remembered and re-included by hand. `skills/` was remembered. Three things were not.

## What never reaches the server

**`.co/host.yaml`** — and it is load-bearing twice over.

1. **The permission whitelist silently reverts to template defaults.**
   `load_permission_patterns()` merges the built-in template with the project's
   `host.yaml`:
   ```python
   template_path = Path(__file__).parent.parent.parent / 'network' / 'host' / 'host.yaml'
   ...
   co_dir = Path(co_dir) if co_dir else (Path.cwd() / '.co')
   host_yaml = co_dir / 'host.yaml'
   if host_yaml.exists():
   ```
   `useful_plugins/tool_approval/approval.py:643-669`

   On the server that file does not exist, so the `if` is false and only the template
   defaults apply. Its own docstring calls this whitelist *"the single source of truth for
   what may run without a human in the loop"*, honored by both the session approval flow
   and network EXEC. So a project that **tightened** its permissions is looser in
   production than on the author's laptop, and one that **widened** them stalls on a
   headless box waiting for an approval nobody can give. Neither is announced.

2. **A non-default `port` produces a deploy that succeeds and cannot answer.**
   The deploy reads the port from the **local** `host.yaml` to configure Caddy:
   ```python
   # host() defaults to 8000; a project that moved it has to say so here, or
   # Caddy would proxy to a port nothing is listening on.
   "port": int(config.get("port") or 8000),
   ```
   `cli/commands/deploy_to_server.py:88-93`

   That comment is right about the consequence and wrong about the premise — saying it
   in `host.yaml` is not enough, because `host.yaml` never ships. Server-side
   `load_host_config()` looks for `.co/host.yaml`, does not find it, and falls back.
   Set `port: 9000` and you get Caddy proxying 9000 while the agent listens on 8000.
   Every step reports success.

**`.co/OO.md`** — the highest-priority project instruction file, read by
`load_project_context()`. Grep the deploy path: `OO.md` does not appear anywhere in
`deploy_to_server.py` or `deploy_commands.py`.

Measured on our own project: an instruction in `OO.md` telling the agent to call
`skill(...)` for its domain took skill invocation from **0/3 phrasings to 3/3**, and cut a
flailing `glob → find → grep -rn` sequence (one `grep` ran 120 s) down to a single tool
call in 17 s. After deploy, that fix does not exist. The agent is a different agent.

**`.co/commands/`** — custom slash commands, read from `.co/commands/*.md` by
`useful_tools/slash_command.py`. Same silent loss.

## Why an include list is the wrong shape

`.co/` holds four different kinds of thing, and only two of them are state:

| kind | authority | examples |
|---|---|---|
| author-written config | **local** | `host.yaml`, `OO.md`, `skills/`, `commands/`, `docs/` |
| generated identity | **server**, irreproducible | `keys/` |
| accumulated runtime state | **server**, disposable | `logs/`, `evals/`, `sessions/` |
| deployer bookkeeping | written over ssh, not rsync | `provision.json`, `requirements.sha256`, `admins.txt` |

The real invariant is **"never overwrite what the server generated and cannot
regenerate"** — not "never send `.co/`". Note that `.env`, which holds actual secrets,
*is* synced today, so the rule was never about secrecy either.

With exclude-by-default, forgetting a file is **silent**: the deploy succeeds and the
agent behaves differently. `skills/` was the first patch, `host.yaml` / `OO.md` /
`commands/` are the second, and the next config file added under `.co/` will be the third.

## Suggested change

**Now — invert the rule.** Exclude only what the server owns:

```python
"--exclude", ".co/keys/",
"--exclude", ".co/logs/",
"--exclude", ".co/evals/",
"--exclude", ".co/sessions/",
```

One change, fixes all three at once, and new config files ship by default. The failure
mode flips from silent (config missing) to loud (a new state directory gets clobbered),
and `keys/` — the one that must never be touched — stays explicitly named.

**Next — make the boundary structural.** Move runtime state under `.co/state/` so the
rule becomes a single `--exclude .co/state/` and correctness stops depending on anyone
remembering a list. `PROVISION_SCHEMA` (`deploy_to_server.py:29`, used at :190) already
exists to run one-time convergence steps, so the `mv` runs once per server. Staying on an
include list while that mechanism is available is choosing to write this patch a third
time.

## Related

- #306 — the identity fix whose exclusion this analyses; the exclusion is right, its
  default is not
- #301 — a skill cannot declare its runtime dependencies
- #324 — the agent does not call the skill tool on its own. The workaround for #324 is an
  instruction in `.co/OO.md`, which this issue shows never reaches the server: the two
  bugs cancel any fix for either.

wu-changxing:
Fixes #392.

## The bug

`_sync_code` excluded `.co/*` and re-included the files worth carrying. That default — exclude unless remembered — lost `.co/skills/` in its first version and `.co/dashboard.html` in its second, and was still dropping `host.yaml`, `OO.md` and `commands/`.

`host.yaml` is the expensive one. It carries the permission whitelist that `load_permission_patterns()` merges over the template defaults, so a project that **tightened** its permissions ran looser in production than on the author's laptop, and one that **widened** them stalled on a headless box waiting for an approval nobody could give. It also carries `port`: the deploy reads it locally to configure Caddy while the server, lacking the file, falls back to 8000. Set `port: 9000` and Caddy proxies a port nothing is listening on — with every step reporting success.

## The fix

Invert the default. Everything ships; `.co/` gets two carve-outs, both because the server owns that data:

```
--filter 'P .co/**'      never delete anything under .co/ on the server
--exclude .co/keys/ …    never send identity or accumulated state
```

Identity, logs, `admins.txt`, `provision.json` and anything a future version writes beside them now survive **without being listed**, and new config ships **without being remembered**. Forgetting to carry config was invisible — the deploy succeeded and a different agent ran. Forgetting to protect state is loud.

## On the tests

Three existing tests failed, and all three asserted the shape of the argv. That is how this went unnoticed for two releases: the argv was exactly as asserted, and the outcome was still wrong. `test_the_deploy_rsync_carries_the_dashboard` even asserted that `.co/dashboard.html` appeared before `.co/*` in the source — true the whole time `host.yaml` was being dropped by that same rule.

They now assert what lands on the server, by running a real rsync into a temp directory:

- config the author wrote arrives (`host.yaml`, `OO.md`, `commands/`, `skills/`)
- the agent keeps its identity
- server-owned state survives `--delete`
- code deleted locally still goes away
- a local `.co/keys/` cannot take over the server

Verified red before the change — `host.yaml`, `OO.md` and `commands/report.md` each failed with "never reached the server" — and green after. Full unit suite: 2607 passed, 0 failed.

## Related

- #324 — the agent does not call the skill tool unprompted. The workaround is an imperative in `.co/OO.md`, which this PR is what makes reach the server; until now the two bugs cancelled any fix for either.
- #393 — same rsync call, opposite direction: `.env` ships when it should not.

### DIFF
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -367,26 +367,49 @@ def _caddy_running(target: str) -> bool:
     return _ssh(target, "systemctl is-active caddy", timeout=60).stdout.strip() == "active"
 
 
+# Everything the author wrote ships, and `--delete` removes what they deleted.
+# `.co/` gets two carve-outs, both because the server owns that data:
+#
+#   P .co/**       nothing under .co/ is ever *deleted* on the server. Identity,
+#                  logs, admins.txt, provision.json and anything a future
+#                  version writes there survive a deploy without being listed.
+#   --exclude ...  identity and accumulated state are never *sent*, so a stale
+#                  local copy cannot overwrite the live one.
+#
+# This replaced an include-list that named the `.co/` files worth carrying.
+# That default — exclude unless remembered — lost `.co/skills/` in its first
+# version and `.co/dashboard.html` in its second, and was silently dropping
+# `host.yaml`, `OO.md` and `commands/` when this was written. Forgetting to
+# carry config is invisible: the deploy succeeds and a different agent runs.
+RSYNC_FILTERS = [
+    "--filter", "P .co/**",
+    "--exclude", ".co/keys/",
+    "--exclude", ".co/logs/",
+    "--exclude", ".co/evals/",
+    "--exclude", ".co/sessions/",
+    "--exclude", ".venv/",
+    "--exclude", ".git/",
+    "--exclude", "__pycache__/",
+]
+
+
 def _sync_code(target: str, agent: str, project_dir: Path) -> bool:
-    """rsync the project, excluding the state directory but carrying skills.
+    """rsync the project, carrying everything the author wrote and deleting
+    nothing the server owns.
 
-    `--delete` removes code files deleted locally, which is wanted. `.co/` is
-    excluded so the agent's identity, logs and evals survive — that exclusion is
-    the whole reason a redeploy no longer reissues the agent's address.
+    `--delete` removes code files deleted locally, which is wanted. It stops at
+    `.co/`, where the server keeps the agent's identity, its logs, the admins
+    list and the convergence marker — data the project has never heard of and
+    an include-list could not enumerate.
 
-    `.co/skills/` is the exception, and it has to be: skills are *what the agent
-    is*, not state it accumulated. Excluding all of `.co/` shipped an agent with
-    none of its skills — the deploy succeeded and the agent ran without them,
-    which is worse than failing. So skills are included, and `--delete` applies
-    inside that directory too so a removed skill actually goes away.
+    What the author writes under `.co/` — `host.yaml`, `OO.md`, `skills/`,
+    `commands/`, `dashboard.html` — is not state. It is what makes this agent
+    this agent, so it ships like the rest of the tree. `host.yaml` in particular
+    carries the permission whitelist and the port: without it the agent runs on
+    template defaults behind a proxy pointed at the wrong port, and reports
+    success either way.
 
-    `.co/dashboard.html` is included for the same reason: the Home page you wrote
-    is part of the agent. Left out, a deploy silently replaces it with a generated
-    starter — the agent looks fine and your dashboard is gone.
-
-    Order matters to rsync: the include must precede the exclude that would
-    otherwise swallow it, and `.co/` itself must be included for rsync to
-    descend into it at all.
+    See RSYNC_FILTERS for the two rules and why the default is now "carry".
     """
     from .server_commands import _identity as _ssh_identity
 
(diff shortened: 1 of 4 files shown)


wu-changxing:
Reviewed this against the edge cases before merging. Two findings, one of them mine and real.

**Verified good:**

- `--filter` in two-argv form (`["--filter", "P .co/**"]`) is accepted by rsync — worth checking, since the pattern contains a space.
- A project with **no local `.co/` at all** does not delete the server's. Identity and logs survive.
- `browser_profile/` and `browser_context/` are hardcoded to `Path.home()/.co` (`browser_tools/browser.py:582`), never project-local, so a developer's logged-in Chrome session cannot ride along.

**Found and fixed:** the protect rule stops `--delete`; it does **not** stop a transfer. Four files are generated on the server or written there over ssh, and each had a plausible local counterpart that would silently replace it:

| file | what a stale local copy does |
|---|---|
| `.co/address.json` | decides super-admin (`trust/tools.py:328`) — changes who owns the agent |
| `.co/admins.txt` | who may command it; `add_admin()` writes this into the cwd's `.co` when run locally, so having one is easy |
| `.co/provision.json` | re-runs convergence |
| `.co/requirements.sha256` | re-runs pip |

The first two are the same class of defect this PR exists to fix — a deploy quietly changing the agent — so they belong in it. Added to the exclude list with four tests asserting the server's value survives.

Full unit suite: 2611 passed.

PR #420, opened 2026-07-31
fix(deploy): .env is a secret on --to, the way it already is on Cloud

### LINKED ISSUE #393: co deploy --to rsyncs the operator's personal OAuth refresh tokens to the server in plaintext

wu-changxing:
Fixes #393. **Stacked on #417** — same rsync filter list, opposite direction. Review that one first; the base will retarget to main when it merges.

## What was happening

`co deploy` parses `.env`, uploads the pairs as secrets, and keeps the file out of the tarball. `co deploy --to` had no notion of a secret at all — the string `.env` does not appear anywhere in `deploy_to_server.py`. It rsynced as ordinary source and then worked by accident: systemd sets `WorkingDirectory`, `load_dotenv()` reads the cwd, and the agent found its keys because a file happened to land beside it.

What travelled is the problem. `create.py:317` copies the whole of `~/.co/keys.env` into every new project's `.env`, so a real project here carried:

```
GOOGLE_REFRESH_TOKEN      MICROSOFT_REFRESH_TOKEN
GOOGLE_ACCESS_TOKEN       MICROSOFT_ACCESS_TOKEN
OPENONION_API_KEY         AGENT_EMAIL  …
```

Long-lived credentials for the operator's personal Gmail and Outlook, rsynced to the server in plaintext, on every deploy. The documentation says the opposite — *"Local-only files such as `.env` … are skipped"* — which is true of the other path, and is exactly why nobody looks.

## The change

| | |
|---|---|
| rsync | excludes `.env` and `.env.*` — secrets stop shipping as source |
| `_sync_env()` | writes them over ssh to `/etc/connectonion/<agent>.env`, root-owned, `0600` |
| the unit | gains `EnvironmentFile=-<path>` — systemd hands them to the process |

The path is **outside the rsync root** deliberately. Inside `/srv/<agent>/`, the next deploy's `--delete` would overwrite it with the laptop's copy, so a key rotated in production could never stay rotated. `0600` and root ownership also mean `co call`'s remote exec, which runs as the service user, cannot read it back out.

`_env_for_server()` additionally rewrites `AGENT_CONFIG_PATH`, which `co init` writes as an absolute path to the author's own `~/.co`. Shipped verbatim it names a directory that cannot exist on Linux, and `gmail.py:181`, `outlook.py:167`, `gdrive.py` and `google_calendar.py` all build their `keys.env` path from it — so those tools fail on the server rather than degrading. Cloud has done this rewrite since it shipped (`deploy_commands.py:429-433`); `--to` had not.

`PROVISION_SCHEMA` → 3, so servers already provisioned pick up the new unit.

## Tests

Verified red first — `.env`, `.env.local` and `.env.production` each failed with *"was rsynced as source"*, and the three new symbols did not exist. Then green. Full unit suite: 2614 passed.

The rsync assertions run a real rsync into a temp directory rather than inspecting argv, for the reason given in #417.

## Not fixed here

`create.py:317` still copies the operator's whole credential file into every new project. Everything above is a patch over that; with it fixed, a leaked project `.env` would cost a project key instead of a mailbox. Worth its own issue.

### DIFF
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -18,6 +18,7 @@ from pathlib import Path
 from typing import Optional
 
 import yaml
+from dotenv import dotenv_values
 from rich.console import Console
 
 from .server_commands import SSH_TIMEOUT_SECONDS, load_server
@@ -27,7 +28,12 @@ console = Console()
 # Bumped when a setup step is added or changes shape. A server reporting an
 # older schema runs the missing steps; one reporting this schema goes straight
 # to rsync and restart.
-PROVISION_SCHEMA = 2
+PROVISION_SCHEMA = 3
+
+# Secrets live outside the rsync root on purpose. Inside `/srv/<agent>/` the
+# next deploy's `--delete` would overwrite them with the laptop's copy, so a key
+# rotated in production could not stay rotated.
+ENV_FILE_TEMPLATE = "/etc/connectonion/{agent}.env"
 
 # How long a unit has to stay up before we call the deploy good, and how long we
 # are willing to wait for it. A single sleep-then-look is not enough: `systemctl
@@ -189,6 +195,7 @@ Wants=network-online.target
 Type=simple
 User={user}
 WorkingDirectory={SRV}/{agent}
+EnvironmentFile=-{ENV_FILE_TEMPLATE.format(agent=agent)}
 {environment}ExecStart={SRV}/{agent}/.venv/bin/python {entrypoint}
 Restart=always
 RestartSec=3
@@ -383,6 +390,11 @@ def _caddy_running(target: str) -> bool:
 # carry config is invisible: the deploy succeeds and a different agent runs.
 RSYNC_FILTERS = [
     "--filter", "P .co/**",
+    # `.env` is a secret, not source. It reaches the server through
+    # _sync_env(), as a root-owned 0600 file systemd reads — not as a
+    # world-readable file that happens to land in the working directory.
+    "--exclude", ".env",
+    "--exclude", ".env.*",
     # Generated on the server, or written there over ssh by the deploy itself.
     # The protect rule above keeps them from being deleted; these keep a local
     # copy from overwriting them, which is the other half of the same rule.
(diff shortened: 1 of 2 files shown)


wu-changxing:
Reviewed before merging and found a real one in my own first version.

**Command injection.** The env file was sent in a heredoc:

```sh
sudo tee /etc/connectonion/myagent.env <<'CO_ENV_EOF'
K=CO_ENV_EOF          ← closes the heredoc here
...rest of the file now runs as shell, under sudo
```

A `.env` value equal to the delimiter ends the heredoc early and hands the remainder to the shell — as root, from input the deploy does not control. Now base64 over the wire and `base64 -d` on the server, so no value is ever parsed as syntax. Tested with a value equal to the delimiter and with `$(touch /tmp/pwned)`, plus a round-trip that the values still arrive intact.

**Multi-line values.** A newline inside a value ends the `KEY=VALUE` line and turns the rest into junk entries, which systemd reads without complaining. A PEM key in `.env` is a normal thing to have. Those keys are now skipped with a warning naming each one, rather than written into a file that looks fine.

Rebased onto the updated #417 — the two exclusion sets merged, no behaviour lost. Full unit suite: 2622 passed.

PR #421, opened 2026-07-31
fix(trust): a list entry matches the whole identifier, and ignores case

### LINKED ISSUE #362: Trust wildcard matching is substring, not prefix — "trusted-*" matches "un-trusted-hacker"
## Bug

`connectonion/network/trust/tools.py:48-51` (`_check_list`) implements wildcards as `pattern = line.replace('*', ''); if pattern in agent_id` — a plain unanchored substring test, not a prefix match anchored to where the `*` actually sits.

```python
check("trusted-*", "un-trusted-hacker")  # -> True, should be False
check("spam*", "no-spam-here-either")    # -> True, should be False
```

## Failure scenarios

- **Over-block**: an admin adds `spam*` to the blocklist meaning "block IDs starting with spam" — any ID that merely *contains* "spam" anywhere is denied.
- **Over-grant (worse)**: an admin whitelists `trusted-*` intending a prefix grant — any client ID containing "trusted-" as a substring anywhere (not as a prefix) is granted access via `is_whitelisted()` → `evaluate_request()` returns `'allow'` (`fast_rules.py:82-83`).

`tests/unit/test_trust_functions.py:60-64`'s `test_wildcard_pattern` only exercises the true-prefix case, so this is untested.

## Fix direction

Match against the position of `*` (prefix/suffix/contains as the pattern actually specifies), e.g. `agent_id.startswith(pattern)` when `*` is trailing, not a blanket substring check regardless of `*` position.

## Definition of done

- `trusted-*` does not match `un-trusted-hacker`
- A test covering the false-positive case from both scenarios above

### LINKED ISSUE #363: Trust block/allow lists don't normalize address case — a blocked address can bypass via case difference
## Bug

`connectonion/network/trust/tools.py:35-54` (`_check_list`) does exact case-sensitive string comparison (`line == agent_id`) with no `.lower()`/normalization, and `is_blocked`, `is_whitelisted`, `is_contact`, `block`, `add_admin` don't normalize either. Addresses are generated as lowercase hex (`address.py:64,120,214`), but nothing stops an admin pasting a mixed-case address into `blocklist.txt` (plausible if copying from a block-explorer-style UI that displays checksummed/mixed case).

## Failure scenario

Admin runs `block("0xABCDEF...")` with mixed case. The real runtime `agent_address` arriving in requests is lowercase `0xabcdef...`. `is_blocked()` returns `False` (fail-open on the deny path) because the exact-match and wildcard checks are both case-sensitive. The blocked address is silently let back in.

## Fix direction

Normalize (lowercase) both sides of every comparison in `_check_list` and anywhere an address is written to/read from a list file.

## Definition of done

- A mixed-case address added to the blocklist correctly blocks the same address regardless of the case it arrives in
- A test covering this exact scenario

wu-changxing:
Fixes #362. Fixes #363.

Both defects live in the same six lines of `_check_list` (`network/trust/tools.py`), and both fail **open** on the whitelist. Splitting them into two PRs would mean the second rewriting the lines the first just changed, so they are fixed together.

## #362 — the wildcard was not anchored

```python
pattern = line.replace('*', '')
if pattern in agent_id:          # substring, anywhere
```

```
whitelist "trusted-*"  →  grants  un-trusted-hacker
blocklist "spam*"      →  blocks  no-spam-here-either
```

The over-block is noisy and someone reports it. The over-grant is silent: `is_whitelisted()` is built on this function, so an admin who wrote a prefix rule handed access to every identifier that merely contains the prefix somewhere.

## #363 — comparison was case-sensitive

Addresses are generated lowercase (`address.py:64`), but an admin pastes what a UI showed them. `block("0xABCDEF…")` wrote the mixed-case line, matched nothing at runtime, and **reported success**. A command whose only purpose is to deny, silently denying nobody.

## The fix

`fnmatch.fnmatchcase` against a lowered pattern and a lowered identifier. `*` keeps meaning "any run of characters" wherever it appears, the pattern is anchored to the whole string rather than searched within it, and `*` alone still matches everything — which people rely on and is covered.

## Tests

Ten cases, verified red first: the two wildcard anchors and the three case behaviours failed on unpatched code, the other five passed and still do. They also pin what must not move — a missing file denies, comments and blank lines are skipped, an exact entry matches only itself, and a bare `*` matches everything.

Full unit suite: 2652 passed.

### DIFF
--- a/connectonion/network/trust/tools.py
+++ b/connectonion/network/trust/tools.py
@@ -14,6 +14,7 @@ Trust Levels (stored in ~/.co/):
   - blocked: In blocklist.txt (denied access)
 """
 
+import fnmatch
 from pathlib import Path
 from typing import List, Callable
 
@@ -33,22 +34,30 @@ def _admins_file(co_dir: Path = None) -> Path:
 
 
 def _check_list(list_name: str, agent_id: str) -> bool:
-    """Check if agent_id is in a list file. Supports wildcards."""
+    """Check if agent_id is in a list file. Supports `*` wildcards.
+
+    A line matches the whole identifier, not part of it. `trusted-*` used to be
+    implemented as `line.replace('*', '') in agent_id` — an unanchored substring
+    test that let `un-trusted-hacker` satisfy a whitelist entry meant to grant a
+    prefix. On the whitelist that fails open, which is the direction that costs
+    something: `is_whitelisted()` is built on this, so the grant is real.
+
+    Comparison folds case. Addresses are generated lowercase (address.py:64),
+    but an admin pastes what a UI showed them, and `block("0xABCDEF…")` that
+    silently blocks nobody is worse than one that errors — it reported success.
+    """
     list_path = CO_DIR / f"{list_name}.txt"
     if not list_path.exists():
         return False
     try:
         content = list_path.read_text(encoding='utf-8')
+        agent_id = agent_id.strip().lower()
         for line in content.strip().split('\n'):
             line = line.strip()
             if not line or line.startswith('#'):
                 continue
-            if line == agent_id:
+            if fnmatch.fnmatchcase(agent_id, line.lower()):
                 return True
-            if '*' in line:
-                pattern = line.replace('*', '')
-                if pattern in agent_id:
-                    return True
         return False
     except Exception:
         return False
--- a/connectonion/network/trust/tools.py
+++ b/connectonion/network/trust/tools.py
@@ -14,7 +14,7 @@ Trust Levels (stored in ~/.co/):
   - blocked: In blocklist.txt (denied access)
 """
 
-import fnmatch
+import re
 from pathlib import Path
 from typing import List, Callable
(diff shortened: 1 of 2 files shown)


wu-changxing:
Reviewed my own first version and found it reintroduced the bug it fixes, one layer down.

`fnmatch` does not only interpret `*`. It also gives meaning to `?` and `[seq]`, which were literal in this file before — and it breaks in **both** directions, both failing open:

```
blocklist "agent[1]"     → no longer matches agent[1]      (a block silently stops blocking)
whitelist "agent[abc]"   → now matches agenta              (a grant nobody wrote)
whitelist "agent?"       → now matches agentX
```

Replaced with an explicit escape-everything-but-`*` translation and `re.fullmatch`, so `*` is the entire wildcard vocabulary, as the docstring says. Four cases added covering brackets, question marks and a regex metacharacter leaking through.

14 tests, full suite 2656 passed.

Issue #442, opened 2026-07-31
the server e2e's own documented command runs zero tests, and exits looking fine


wu-changxing:
`tests/e2e/real_api/test_server_lifecycle.py` opens with the way to run it:

```
CO_E2E_SERVERS=1 pytest tests/e2e/real_api/test_server_lifecycle.py -v -s
```

That command collects 11 tests and runs **none** of them:

```
collected 11 items / 11 deselected / 0 selected
============================ 11 deselected in 0.07s ============================
EXIT=5
```

`pytest.ini` carries `-m "not real_api and not network"` in `addopts`, and everything under `tests/e2e/real_api/` gets the `real_api` marker automatically. So the default filter removes the whole file.

## Why it matters more than a typo

This is the suite that exists because **every step in it shipped broken and passed the unit tests anyway** — its docstring lists seven such failures. It is the only thing standing between a change and a deploy that silently does not work.

And the way it fails is quiet. There is no error, no red. Unless you read the collection line, `11 deselected in 0.07s` looks like a suite that ran. Exit code 5 (`NO_TESTS_COLLECTED`) is not 0, but a human watching a terminal reads the last line, not `$?`.

So the failure mode is: someone changes the deploy path, runs the command the file told them to run, sees no failures, and ships.

## Fix

Correct the documented command:

```
CO_E2E_SERVERS=1 pytest tests/e2e/real_api/test_server_lifecycle.py -m real_api -v -s
```

Better, make it not depend on remembering: the file is already gated behind `CO_E2E_SERVERS=1`, so the marker filter is doing no work here that the env var does not already do. A `-p no:cacheprovider`-style override in the module, or a conftest that clears `-m` for this path, would mean the obvious command is also the right one.

Worth checking whether the other files under `tests/e2e/real_api/` have the same problem — the marker is applied by directory, so any of them invoked the documented way would be silently skipped too.

## How it was found

Ran it to add dashboard coverage to the deploy path, and the run finished in 0.07s.

wu-changxing:
No longer true. `make test-e2e` runs the suite today:

```
$ make test-e2e          # pytest -m e2e
tests/e2e/test_relay_e2e.py::TestRelayE2E::test_relay_single_request PASSED  [ 99%]
tests/e2e/test_relay_e2e.py::TestRelayE2E::test_relay_multiturn      FAILED  [ 99%]
tests/e2e/test_relay_e2e.py::TestRelayE2E::test_raw_websocket_session_id PASSED
[...]

PR #444, opened 2026-07-31
test(e2e): cover the Home page in the deploy path, and fix the command that ran nothing

wu-changxing:
Two changes to `tests/e2e/real_api/test_server_lifecycle.py`, found while trying to test the deploy → dashboard path end to end.

## The documented command ran zero tests

The module docstring said:

```
CO_E2E_SERVERS=1 pytest tests/e2e/real_api/test_server_lifecycle.py -v -s
```

which collects 11 tests and runs none:

```
collected 11 items / 11 deselected / 0 selected
============================ 11 deselected in 0.07s ============================
```

`pytest.ini` carries `-m "not real_api and not network"` in `addopts`, and this directory gets the `real_api` marker automatically.

No failure, no red — a run that looks like it passed and tested nothing. For the one suite that exists **because every step in it shipped broken and passed the unit tests anyway**, that is the worst possible failure mode. Fixed, with the reason written down. Filed as #442.

## The dashboard was never checked against a real deploy

It moved into `.co/` (#405), and `.co/*` is excluded from the deploy rsync with a few paths included back by name. Whether that include fires is an rsync filter argument — a unit test cannot answer it, and the failure is silent by construction: miss the include and the deploy still succeeds, the agent still starts, and it writes a fresh starter over the operator's page while looking entirely healthy.

Four tests:

- an authored `.co/dashboard.html` travels intact
- a redeploy carries the author's new version
- a starter generated **on the server** offers the skill that actually shipped there (a client refuses any button naming a skill the agent did not publish, so a stale starter is a page of buttons that silently do nothing)
- nothing lands in the project root, where the next rsync would not carry it

## Status: not green yet

Run against a real machine, these do not pass, and the reason is upstream of the tests. `co server new` reported:

```
✓ e2e-d0113a is ready
  co@35.197.164.214
  expires 2027-07-31 — $360.00 charged
Next: co server check e2e-d0113a
```

and the very next command answered:

```
$ co server check e2e-d0113a
No server named 'e2e-d0113a'.
```

A machine that was created, charged for, and then unreachable by name from the same CLI on the same laptop. Every test after it fails on that, including the dashboard ones — they never got to run against a deploy.

`~/.co/servers.yaml` is read-modify-written with no locking (`_load()` … `_save()`), and I had a `co server ls` running concurrently, so I may have caused this one myself. Either way it is worth its own issue: a registration for a server you just paid $360 for should not be losable by another `co` command running at the same time.

Merging this is still worth it — the tests are correct and the docstring fix stops the suite silently passing — but **the e2e is not green**, and I would rather say so than let a merged PR imply otherwise.

### DIFF
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -327,9 +327,19 @@ def _write_unit_if_changed(target: str, agent: str, entrypoint: str,
     Rewriting it every deploy would mean a daemon-reload every deploy for no
     reason, and would hide whether a restart came from new code or a new unit.
     """
-    # The ssh target's user owns everything under /srv/<agent> — it is the user
-    # rsync wrote as — so it is the one the service should run as.
-    user = target.split("@")[0]
+    # The user rsync wrote as owns everything under /srv/<agent>, so it is the
+    # one the service must run as. Ask the machine rather than splitting the
+    # target on "@": ssh does not require that form. A server registered with
+    # `co server add <name> --ssh <alias>` has a config alias as its target, and
+    # splitting it yielded the alias as a username —
+    #
+    #   User=nw-e2e
+    #   Failed to determine user credentials: No such process
+    #   status=217/USER
+    #
+    # a deploy that copied every file, wrote the unit, reported success, and
+    # produced a service that could never start.
+    user = _ssh(target, "id -un", timeout=60).stdout.strip() or target.split("@")[0]
     wanted = _unit_text(agent, entrypoint, hostname,
                         port=_free_port(target, agent), user=user)
     unit_path = f"/etc/systemd/system/{agent}.service"
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -172,7 +172,7 @@ def _read_provision(target: str, agent: str) -> dict:
         return {"schema": 0}
 
 
-def _unit_text(agent: str, entrypoint: str, hostname: Optional[str] = None,
+def _unit_text(agent: str, entrypoint: str, hostname: Optional[str] = None, port: int = 8000,
                user: str = "root") -> str:
     """The systemd unit. Restart=always and enable are what the container was for.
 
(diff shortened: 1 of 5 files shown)


wu-changxing:
Confirming the first half is still reproducible on main today, so this has not gone stale while sitting open. Same command, current numbers:

```
$ pytest tests/e2e/real_api/test_server_lifecycle.py --collect-only -q
collected 7 items / 7 deselected / 0 selected
```

`pytest.ini` still carries `-m "not real_api and not network"` in `addopts`, and this directory still gets the `real_api` marker from the folder hook. A documented command that runs nothing and reports no failure — in the suite that exists *because* every step in it shipped broken and passed the unit tests anyway.

## One thing that changed underneath this PR

I narrowed `make test-e2e` in #578, and the mechanism is worth stating here because it is the same one:

```make
test-e2e:
	$(PYTEST) -m "e2e and not real_api"     # was: -m e2e
```

A `-m` on the command line **replaces** the `-m` in `addopts` rather than combining with it. That is why `make test-e2e` was running 140 paid-provider tests — 52 of them failing with 401s — while `pytest.ini` looked like it excluded them. Two commands, opposite failure modes, one cause:

- `make test-e2e` overrode the exclusion and ran tests it should not have
- the command in this docstring inherits the exclusion and runs nothing at all

Neither says so. Both look fine.

## Not touching the rest

The four dashboard-in-deploy tests and the "not green yet" server section are yours and I have not duplicated any of it. Flagging only that the premise is intact, in case the age of the PR made it look otherwise.

wu-changxing:
Attempted the rebase during the 1.6.0 PR sweep and stopped partway, because it is not mechanical and pushing through it would have been guesswork in deploy and server-registry code.

The branch carries five commits, and they are in different states against `main`:

**Already on main, independently.** `fix(server): a registration for a machine you paid for must not be losable` conflicts because `server_commands.py` already has `_registry_lock` and `_write_atomically` — the same idea, arrived at separately. Skipping that commit is almost certainly right, but "almost certainly" is not good enough for the file that records machines someone is being billed for, so I have not decided it here.

**Not on main.** `fix(deploy): give each agent its own port, instead of every agent taking 8000` — I looked, and there is no port allocation in `deploy_to_server.py` on main. That one is still wanted, and it conflicts too.

**The three e2e commits** (Home page coverage, one home for the suite, capturing the registry at creation) are the PR's actual subject and look independent of all of it.

So this needs its commits triaged one at a time against what landed in the meantime, by someone who can say for each: already done, still wanted, or superseded. That is a different job from a sweep, and doing it badly in the deploy path is expensive.

Rebase aborted; the branch is untouched.

wu-changxing:
Closing this mixed, conflicting branch rather than treating it as a 1.6 candidate. It combines valuable E2E coverage and registry atomicity with several deploy experiments, including a port-allocation path whose own commit notes say it still selected an occupied port and was not landing correctly. The underlying work remains traceable in #442, #445, #450, and #451. Any still-needed fix should be reintroduced from current main as a focused PR with a green real E2E, not merged from this history.

PR #461, opened 2026-07-31
fix(deploy): after a deploy, the server runs the version that deployed to it

### LINKED ISSUE #460: A released fix never reaches a deployed agent — co deploy skips the install that would carry it
A released fix never reaches a deployed agent — `co deploy` skips the install that would carry it.

## What happened

1.5.6 went out. It contains #446/#457, which stop every agent from calling itself `oo`. On a
server running an agent deployed from this project:

```
$ pip install -U connectonion          # local CLI → 1.5.6
$ co deploy --to nw-e2e
  syncing code …
  writing secrets … (4 keys)
  writing systemd unit …
  restarting …
✓ naturewill is running on nw-e2e
```

```
$ ssh nw-e2e '/srv/naturewill/.venv/bin/pip show connectonion | grep ^Version'
Version: 1.5.5
$ ssh nw-e2e 'curl -s localhost:8001/info | jq -r .name'
oo
```

Deploy reported success at every stage. The server kept the old library, and the fix that
was the entire point of the release was not there. `pip install -U connectonion` by hand,
restart, and the name becomes `naturewill`.

## Cause

`cli/commands/deploy_to_server.py`, `_install_deps_if_changed()`:

```python
digest = hashlib.sha256(requirements.read_bytes()).hexdigest()
stamp = f"{SRV}/{agent}/.co/requirements.sha256"

current = _ssh(target, f"cat {stamp} 2>/dev/null", timeout=60)
if current.returncode == 0 and current.stdout.strip() == digest:
    return True
```

The reasoning in the docstring is right — reinstalling on a code-only change would make
every deploy take minutes. But the key is the *text* of `requirements.txt`, and the
template's requirements.txt names the framework without a version:

```
connectonion
pymupdf>=1.28
pypdf>=6.0
```

An unpinned line's text never changes, so its resolved version is frozen at whatever was
installed the first time. For third-party packages that is a defensible default. For
`connectonion` it is not: it is the runtime the agent *is*, and the deploying CLI is the
thing that just changed.

## Why this one is worse than ordinary staleness

It is silent and it inverts the operator's model. "Upgrade, redeploy" is the whole
mechanism by which a fix reaches production, and it reports success while doing nothing.
Someone chasing a bug that was fixed a release ago has no reason to suspect the library on
the server is older than the CLI that just deployed to it — `co deploy` is the only thing
that touches that machine.

It also allows version skew between the two halves of a deploy: the local CLI writes the
systemd unit, the `.co/` layout and the secrets file, and a server-side runtime of a
different version reads them. #443 and #448 both changed what gets written; an old runtime
reading new config is a failure mode nobody is testing.

## What would fix it

Make the stamp cover the deploying CLI's version, not just the requirements text:

```python
digest = hashlib.sha256(
    requirements.read_bytes() + __version__.encode()
).hexdigest()
```

A CLI upgrade then reinstalls once, and code-only deploys stay fast — which is what the
original optimisation was protecting.

An alternative worth considering is keeping `connectonion` out of the requirements
comparison entirely and always aligning it with the deploying CLI, since skew between them
is not a state anyone wants. Either way, the property to hold is: **after `co deploy`, the
server runs the version of connectonion that deployed to it.**

## Environment

connectonion 1.5.6 (local) / 1.5.5 (server), target Ubuntu 24.04.4 LTS, deployed via
`co deploy --to`.

wu-changxing:
Closes #460.

1.5.6 shipped #446/#457 — the fix that stops every agent calling itself `oo`. Then:

```
$ pip install -U connectonion          # local CLI → 1.5.6
$ co deploy --to nw-e2e
✓ naturewill is running on nw-e2e

$ ssh nw-e2e 'pip show connectonion | grep ^Version'   → 1.5.5
$ ssh nw-e2e 'curl -s localhost:8001/info | jq -r .name' → oo
```

Success at every stage. The fix that was the point of the release was not on the server.

## Two faults, stacked

**The install was skipped.** The stamp hashed `requirements.txt`, and the template names the framework without a version:

```
connectonion
pymupdf>=1.28
```

An unpinned line's text never changes, so its resolved version was frozen at first install. The stamp now covers the deploying CLI's version as well — a CLI upgrade reinstalls once; code-only deploys still skip, which is what that optimisation was protecting.

**The install then did nothing.** `pip install -r` leaves an already-satisfied unpinned requirement alone. I found this by fixing the first fault and watching `installing dependencies …` print while `pip show` did not move. The runtime is now named explicitly:

```
pip install -q connectonion==<this CLI's version>
```

Fixing either fault alone leaves the property broken, which is why both are here.

## Why version skew matters beyond staleness

The local CLI writes the systemd unit, the `.co/` layout and the secrets file; the server-side runtime reads them. #443 and #448 both changed what gets written. An old runtime reading new config is a failure mode nobody is testing, and until now nothing kept the two halves the same software.

## Verified

Downgraded a real server to 1.5.5, deployed from a 1.5.6 CLI:

```
installing dependencies …
✓ naturewill is running on nw-e2e

Version: 1.5.6
name= naturewill  skills= 1
```

Unit tests: `tests/unit/test_deploy_carries_its_own_version.py`, four cases — a CLI upgrade reinstalls, an unchanged deploy still skips, changed requirements still reinstall, and the install names the version it wants. The first and last fail before this change.

One existing test updated: `test_unchanged_requirements_skip_the_install` computed the stamp itself and had to learn the version is part of it. Its intent — a code-only change takes seconds — is unchanged and still asserted.

206 deploy/server tests pass.

### DIFF
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -22,6 +22,7 @@ import yaml
 from dotenv import dotenv_values
 from rich.console import Console
 
+from ... import __version__
 from .server_commands import SSH_TIMEOUT_SECONDS, derived_agent_identity, load_server
 
 console = Console()
@@ -692,7 +693,20 @@ def _install_deps_if_changed(target: str, agent: str, project_dir: Path) -> bool
     if not requirements.exists():
         return True
 
-    digest = hashlib.sha256(requirements.read_bytes()).hexdigest()
+    # The deploying CLI's version is part of what makes an install current.
+    #
+    # Comparing the requirements text alone froze every unpinned line at the
+    # version resolved on first install — and the template pins nothing, so
+    # `connectonion` itself never moved. Upgrading the CLI and redeploying, the
+    # mechanism by which a released fix reaches production, reported success at
+    # every stage and left the old runtime in place.
+    #
+    # It also let the two halves of a deploy drift apart: this CLI writes the
+    # systemd unit, the .co/ layout and the secrets file, and a runtime of some
+    # older version reads them.
+    digest = hashlib.sha256(
+        requirements.read_bytes() + __version__.encode()
+    ).hexdigest()
     stamp = f"{SRV}/{agent}/.co/requirements.sha256"
 
     current = _ssh(target, f"cat {stamp} 2>/dev/null", timeout=60)
@@ -710,6 +724,15 @@ def _install_deps_if_changed(target: str, agent: str, project_dir: Path) -> bool
         f"set -e\n"
         f"cd {SRV}/{agent}\n"
         f"{SRV}/{agent}/.venv/bin/pip install -q -r requirements.txt\n"
+        # …and then the runtime, at this CLI's version.
+        #
+        # `pip install -r` leaves an already-satisfied unpinned requirement
+        # alone, and the template names the framework without a version. So the
+        # install above ran, reported success, and left 1.5.5 in place while
+        # 1.5.6 was doing the deploying. Naming the version is the only way the
+        # two halves of a deploy stay the same software.
+        f"{SRV}/{agent}/.venv/bin/pip install -q "
+        f"{shlex.quote('connectonion==' + __version__)}\n"
         f"printf '%s' {shlex.quote(digest)} > {stamp}",
         timeout=900,
     )
--- a/tests/unit/test_deploy_carries_its_own_version.py
+++ b/tests/unit/test_deploy_carries_its_own_version.py
@@ -1,0 +1,97 @@
+"""After a deploy, the server runs the connectonion that deployed to it.
+
+The install step was skipped whenever requirements.txt was byte-identical to
+last time. The template pins nothing — `connectonion` on its own line — so an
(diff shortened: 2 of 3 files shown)


wu-changxing:
Superseded by #462, which landed first and covers both halves — the digest takes the CLI version, and the install uses `-U` so an already-satisfied unpinned requirement actually moves. Closing this rather than rebasing it.

Same two faults, found the same way: the install being skipped, and then `installing dependencies …` printing while `pip show` did not move.

One difference worth a look, not a request to change anything. `-U -r requirements.txt` resolves each unpinned line to whatever is newest on PyPI at deploy time, so if a release lands between a CLI upgrade and a deploy, the server ends up *ahead* of the CLI that deployed to it — the same skew, other direction. Pinning the runtime to the deploying CLI (`connectonion==<version>`) is the narrower version of the guarantee. `-U` also upgrades the project's own unpinned dependencies, which may or may not be wanted on a deploy that only meant to carry a framework fix.

Verified #462's behaviour on a real server anyway: downgraded to 1.5.5, deployed from a 1.5.6 CLI, server came back on 1.5.6 with the agent answering to `naturewill` instead of `oo`.

Issue #396, opened 2026-07-31
A deployed agent's address cannot be known before the first boot, nor recovered after the machine is gone


wu-changxing:
## Where identity comes from today

`host()` mints one if it does not find one:

```python
addr_data = address.load(co_dir)

if addr_data is None:
    addr_data = address.generate()
    address.save(addr_data, co_dir)
```
`network/host/server.py:509-514`

`address.load()` reads `co_dir/keys/agent.key` and nothing else — there is no fallback to
`~/.co/keys/` (`address.py:183-204`).

Meanwhile `co init` never creates a project `.co/keys/`. It copies identity **values** out
of the global `keys.env` into the project `.env`:

```python
IDENTITY_KEYS = {'AGENT_CONFIG_PATH', 'AGENT_ADDRESS', 'OPENONION_API_KEY',
                 'AGENT_EMAIL', 'IS_EMAIL_ACTIVE'}
```
`cli/commands/init.py:173-175`

So a freshly initialised project carries `AGENT_ADDRESS=0x…` in `.env` describing the
**global** identity, while the agent that actually boots there mints a **different** key
on first run and reports that one. Three addresses end up in play for one project:

| identity | born | who believes in it |
|---|---|---|
| global `~/.co/keys/` | first `co init` | `AGENT_ADDRESS` in every project `.env` |
| project `.co/keys/` | first local `host()` run | the locally running agent |
| server `/srv/<agent>/.co/keys/` | first run on the server | the deployed agent |

Since #320 the third one is stable across redeploys, which is right and was the point of
#306. The remaining problem is not stability. It is that the address is **unknowable in
advance and unrecoverable afterwards**.

## Why server-minted is the wrong default

**You cannot know the address before deploying.** Anything that wants it up front —
telling a counterparty "our agent is 0x…", pre-seeding a whitelist, a DNS record, a
contract — has to wait for the first boot and then ssh in to read it back.

**You cannot get it back if the box dies.** `recovery.txt` is written to
`/srv/<agent>/.co/keys/` and exists nowhere else. Rebuild the server and every trust
relationship, whitelist entry and the agent's `@mail.openonion.ai` address are gone. This
is #306's failure mode moved up a level: from "changes every deploy" to "changes every
machine", which is rarer and therefore worse — it will be discovered during an outage.

## Three options

**A — keep server-minted.** Private key never travels. But unknowable in advance and
unrecoverable, as above.

**B — mint locally at first deploy and push over ssh.** Address known at deploy time,
[...]

PR #517, opened 2026-08-01
The address is a footnote, not a headline

wu-changxing:
Follow-up to #515, which put the agent's address on the Home page. Right fact, wrong place — found by rendering it at the real pane width instead of reading the HTML.

At 440px the address is two full lines of monospace hex sitting directly under the title. It is the second-loudest thing on the page, above the skills, and it is the fact you look at **once**, when you deploy, and never again.

It moves below the skills, behind a hairline rule, at `opacity: .5`. Still selectable, still in full — the reason it exists (a deployed agent's address cannot be known before its first boot, #396) is unchanged.

No test changed: every assertion in #515 was about the address being present, in full, escaped, and absent when unknown. All still true, which is the honest summary of what those tests were and were not checking — none of them could see that it looked wrong.

`50 passed`.

### DIFF
--- a/connectonion/network/host/ws_router/starter.html
+++ b/connectonion/network/host/ws_router/starter.html
@@ -53,13 +53,16 @@
     font-variant-numeric: tabular-nums;
   }
 
-  /* The address is 66 characters and this pane can be 440px wide. It wraps
-     rather than shrinking or truncating: it is here to be read and copied, and
-     an ellipsis would defeat the only reason it is on the page. */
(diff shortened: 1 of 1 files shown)


wu-changxing:
Superseded by #534. #534 is the full control-panel rework and it moves the address below the skills as its own `addr` fragment — the same relocation this PR made, plus the verdict/schedule/next-run redesign around it. Merging both would conflict on the same lines. Closing this; the address fix lives in #534.

PR #520, opened 2026-08-01
fix: replace mismatched exported SSH keys

### LINKED ISSUE #513: co keys --ssh --write leaves a mismatched keypair and calls it safe
`co server new` puts the **derived** SSH key in the machine's metadata. `co keys --ssh --write` is how you get the private half onto disk. If a file is already at that path, it refuses:

```
Delete it first if you want it rewritten; the key is derived, so nothing is lost.
```

That sentence is true only when the existing file *is* the derived key. When it is not, the result is a keypair that cannot open anything:

```
~/.ssh/connectonion_ed25519(.pub)   ssh-ed25519 …AAAAIDG+k4I37b3OKd…   ← self-consistent, and wrong
co keys --ssh                       ssh-ed25519 …AAAAICa2D66JuoApuoKe… ← what lands in server metadata
```

The server has one key; the path the tool points at holds another. Every `co server new` machine is then unreachable, and nothing says why.

## What it costs

I lost about an hour to this today, and the wrong turns are worth listing because each one looks like a real diagnosis:

- `co server add` → *"Permission denied (publickey)"*, with the tip *"your derived key is `co keys --ssh` — it has to be in authorized_keys on that host."* It **was** on the host. The tip sends you to check the end that is already correct.
- Suspected the machine. Read the serial console, found `ERROR oslogin.go:173 … Unit sshd.service not found` (real — Ubuntu 24.04 calls it `ssh.service`) and started writing an issue about broken provisioning.
- Suspected the image, the guest agent, OS Login, IP conflicts. All dead ends.

The machine was fine the whole time. Guest agent logs show `Creating user co` and `Updating keys for user co` at boot, and once the right private key was on disk it logged straight in — Ubuntu 24.04.4, Python 3.12.3, preflight-clean.

Anyone who had a `connectonion_ed25519` file from before #310 shipped derivation hits this, and the message actively reassures them nothing is wrong.

## Suggested

`--write` should compare before it refuses:

- file **is** the derived key → say so, exit 0, nothing to do
- file is **something else** → this is the case worth interrupting for. Say the path holds a different key, that the servers expect the derived one, and offer to back it up and replace it. Silence here is what produces an unreachable fleet.

The failure message in `co server add` should also point at the half that is actually missing — comparing `ssh-keygen -yf ~/.ssh/connectonion_ed25519` against `co keys --ssh` is a one-line check that would have ended this in seconds.

## Related

- #310 — where the derived SSH key came from. The write path predates it and was never reconciled.

sanmaxdev:
## Description

Make `co keys --ssh --write` distinguish the expected derived key from a stale or unrelated keypair. Matching files are left unchanged; mismatched files are backed up before both halves are replaced.

## Related Issue

Fixes #513

## Type of Change
- [ ] Documentation update

## Changes Made
- Compare the private key and public sidecar with the derived public key
- Preserve mismatched files under collision-safe backup names
- Add regression coverage for matching and mismatched keypairs

## Testing
- [x] `uv run pytest tests/unit/test_keys_commands.py tests/e2e/cli/test_cli_keys.py -q`
- [x] `uvx ruff check connectonion/cli/commands/keys_commands.py tests/unit/test_keys_commands.py`
- [x] `uvx --from build pyproject-build --wheel`
- [x] `OPENONION_API_KEY=test-key uv run pytest tests/ -q --tb=short -m 'not slow and not real_api and not network'`
- [x] `git diff --check`

## Checklist

### DIFF
--- a/connectonion/cli/commands/keys_commands.py
+++ b/connectonion/cli/commands/keys_commands.py
@@ -10,11 +10,14 @@ LLM-Note:
 """
 
 import os
+import shutil
+import subprocess
 from pathlib import Path
+
+from dotenv import load_dotenv
 from rich.console import Console
 from rich.panel import Panel
 from rich.table import Table
-from dotenv import load_dotenv
 
 console = Console()
 
@@ -254,6 +257,35 @@ def write_derived_ssh_key(seed_phrase: str) -> Path:
     return SSH_PRIVATE_KEY
 
 
+def _public_key_identity(public_line: str) -> str:
+    """Return the key type and payload, excluding an optional comment."""
+    return " ".join(public_line.split()[:2])
+
+
+def _public_key_from_private(private_path: Path) -> str | None:
+    """Read the public half of an OpenSSH private key."""
+    try:
+        result = subprocess.run(
+            ["ssh-keygen", "-yf", str(private_path)],
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+    except OSError:
+        return None
+    return result.stdout.strip() if result.returncode == 0 else None
+
+
+def _backup_path(path: Path) -> Path:
+    """Choose a backup name without overwriting an earlier backup."""
+    candidate = path.with_name(path.name + ".bak")
+    index = 1
+    while candidate.exists():
+        candidate = path.with_name(f"{path.name}.bak.{index}")
+        index += 1
+    return candidate
+
+
 def _show_ssh_key(addr_data: dict, write: bool = False) -> None:
     """Print the SSH public key derived from the recovery phrase.
 
@@ -261,7 +293,6 @@ def _show_ssh_key(addr_data: dict, write: bool = False) -> None:
     one thing to write down, and the operator can reach a provisioned server
     without a second secret to manage.
     """
-    from pathlib import Path
     from ... import address
 
     seed = addr_data.get("seed_phrase")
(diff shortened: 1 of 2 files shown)


wu-changxing:
Reviewed against #513's report and ran the branch. The shape is right — compare, keep what matches, back up what does not — and it fixes the hour-long dead end the issue describes. One thing to settle before it goes in, and one smaller note.

## Without `ssh-keygen`, a correct key is replaced every run

`_public_key_from_private()` shells out to `ssh-keygen -yf` and returns `None` when the binary is missing (`OSError`) — and the match condition requires it:

```python
if (existing_private_public
    and _public_key_identity(existing_private_public) == expected
    and existing_public
    and _public_key_identity(existing_public) == expected):
```

`None` fails the first clause, so the file is treated as a mismatch. On a machine with no OpenSSH — Windows without the optional feature, a slim container — that means the **correct** derived key is backed up and rewritten on every invocation. Ran it twice on this branch with `subprocess.run` raising `OSError`, starting from the correct key on disk:

```
backups left behind: ['connectonion_ed25519.bak', 'connectonion_ed25519.bak.1',
                      'connectonion_ed25519.pub.bak', 'connectonion_ed25519.pub.bak.1']
```

Nothing is lost — the backups are the point — but a command that reports "the existing SSH key does not match" about a key that does match is the same shape as the bug being fixed: a true-sounding sentence that sends someone looking in the wrong place.

Two ways out, either is fine:

- Derive the public half in Python. `address.py` already writes this format (`_openssh_private_key`), so reading it back is a parse rather than a subprocess, and it works everywhere.
- Or keep `ssh-keygen` and treat "cannot tell" as its own case: fall back to comparing the `.pub` sidecar alone, and if that matches, leave both files alone.

## Smaller: the sidecar alone would answer most of the time

`existing_public` is read directly and needs no subprocess. When it matches and the private half is unreadable, replacing is defensible; when it matches and the private half *is* readable and also matches, we are running `ssh-keygen` to confirm what the sidecar already said.

## What I checked

- The failure in #513 does still reproduce on `main` — the refusal message ("the key is derived, so nothing is lost") is only true when the file is the derived key.
- CI on this branch is green on all four Pythons.
- The new tests cover matching and mismatched pairs; neither covers the no-`ssh-keygen` path, which is where the behaviour above hides.

Not merging this myself since it is your change and the first point is a real behaviour decision. @wu-changxing for the call.

wu-changxing:
Half done on `366bcba`. The bug this fixes is gone; the behaviour asked for here is not what shipped.

**Done — mismatched files no longer survive.** `write_derived_ssh_key` writes both halves together and overwrites whatever was at the path:

> Both halves, always together, overwriting whatever was there. Overwriting is safe precisely because the key is derived: the phrase is the original, the files are a cache of it. Writing one half and keeping the other is what produces a pair that cannot authenticate.

Verified in an isolated HOME by putting a foreign key at the path first — public half matches the derived key afterwards, the stale private half is replaced, mode 0600, both halves from one phrase. #513 is closed on that evidence.

**Not done — the two refinements this issue asks for:**

- *"Matching files are left unchanged"* — there is no comparison. The files are rewritten every time, matching or not. Harmless in effect (same bytes), but it does mean mtime moves and a watcher sees a change that is not one.
- *"Mismatched files are backed up before both halves are replaced"* — there is no backup. `grep -n "backup\|\.bak\|rename\|copy2"` in `keys_commands.py` returns nothing. A foreign key at that path is destroyed, not set aside.

The second is the one with a real cost. The reasoning that makes overwriting safe — "the phrase is the original, the files are a cache of it" — holds for *our* derived key and not for whatever else might be sitting there. Someone who put an unrelated key at that path loses it, and the message says nothing was lost.

Whether that matters depends on how likely the path is to hold a foreign key. It is `~/.co/ssh/id_ed25519` now, not `~/.ssh/connectonion_ed25519` as #513 described — a directory we own, so the collision is much less likely than it was when this was filed. That may be the reason to close this rather than finish it; your call.

Not fixing it unasked: the change is small, but it turns a deliberate "overwriting is safe" into a conditional one, and the comment above suggests the current behaviour was chosen rather than overlooked.

sanmaxdev:
Closing this out. #513 is fixed on main: both halves are written together, overwriting whatever was at the path, so a stale half can no longer survive beside a fresh one. The comparison-and-backup path here no longer fits that shape.

On the backup-on-mismatch refinement: `~/.co/ssh/` is a directory this project owns, and the overwrite is intentional. Adding a backup would turn a deliberate overwrite into a conditional for a path that is unlikely to hold a foreign key, so I agree it isn't worth the branch.

PR #525, opened 2026-08-01
feat(dashboard): Home says what the agent has been doing

### LINKED ISSUE #522: The starter dashboard shows what an agent is, and nothing about what it has been doing
The starter dashboard shows what an agent *is*, and nothing about what it has been *doing*.

## What it renders today

`network/host/ws_router/starter.html` — a name, a subtitle, and every published skill as a button. That is the whole page.

The craft is fine: dark mode, no JavaScript by CSP contract, a 440px column that survives being a whole browser window, `<details>` groups for agents with a hundred skills. The problem is not how it looks. It is that **the page is blind to the agent's own state.**

Open a deployed agent's Home and you cannot answer:

- has this thing run at all today?
- did the last run succeed?
- what is it going to do next, and when?
- it has been up for three days — doing what?

Every one of those is already on disk, next to the file that renders the page.

## The data exists and is unused

`.co/session_results.jsonl` — append-only, one line per state change:

```json
{"session_id":"4ddb18c7…","status":"running","prompt":"hi","result":null,"created":1785526725.9,"duration_ms":null}
{"session_id":"4ddb18c7…","status":"done","prompt":"hi","result":"Hello! How can…","duration_ms":17609}
```

`.co/logs/<agent>.log` — the plain-text audit trail.

So "the last five things this agent did, whether each finished, and how long it took" is a `tail` away and has never been shown. An agent that ran unattended overnight is exactly the case where someone opens Home wanting to know what happened, and the page tells them the agent's name.

`status: running` is worth surfacing on its own: a run that died mid-flight stays `running` forever, which makes it a free stuck-detector — but only if something looks.

## Why this matters more once agents run on a schedule

#521 gives an agent recurring work. The moment it has one, Home is the only place anyone would look to see it, and the questions above stop being nice-to-have:

- a schedule was configured but never fired — is it broken, or not due yet?
- it fired at 09:00 and the result was an error — where does that surface?

A background agent whose activity is invisible is the same failure this repo has already fixed twice: #434 (a trust denial reported nowhere the operator would look) and #460 (a deploy that silently carried nothing). Same shape — the system knew, and nothing said so.

## Suggested shape

Keep the constraints that make the current page good — one column, no JS, CSP-safe, degrades on a narrow pane — and give it something to say:

- **Status line**: up since when, model, trust level. `/info` already publishes all three.
- **Recent activity**: the last few sessions — prompt, outcome, duration. Anything still `running` marked as such.
- **Schedule** (when #521 lands): what runs, when it next runs, when it last ran and whether that succeeded.
- **Skills**: as today. They are the actions; the rest is context for choosing one.

Ordering follows the question being asked. Someone opening Home on a deployed agent wants "is it alive and is it working" before "what can it do" — the first-run case is the opposite, but that is the case where the activity sections are empty and collapse to nothing anyway.

## On where a schedule UI belongs

`oo-chat` carries `Schedule` and `Automation` as sidebar nav items marked `soon: true` (`components/sidebar.tsx:31-35`). That puts the schedule in the client, one level above the agents.

A schedule belongs to one agent. It lives in that agent's `.co/schedule.yaml`, it is deployed with that agent, and it means nothing beside a different agent's. The client is a viewer. Rendering it in the agent's own dashboard also means it travels for free — every client that renders `dashboard.html` gets it, including one nobody has written yet, with no client work at all. That is the same property that made the dashboard a single HTML file over the existing WebSocket rather than a new endpoint and a sidecar JSON.

## Note on the override path

`~/.co/starter.html` replaces the bundled template, and overrides layer fragment-by-fragment so a restyle does not have to copy the skill row along with it. New sections should be fragments for the same reason — otherwise every operator who restyled the shell loses the activity list without being told.

wu-changxing:
Closes #522. Builds on #524 (the schedule) and #518 (Home renders live rather than being frozen on first boot — without that, an activity section would be a fossil).

## What was missing

Open a deployed agent's Home and you could not answer: has this run today, did the last run work, what is it going to do next. Every one of those was already on disk next to the file that renders the page.

`.co/session_results.jsonl` has been recording `session_id`, `status`, `prompt`, `result` and `duration_ms` for every turn all along, and nothing has ever read it.

## Two sections, both absent when empty

**Scheduled** — what the agent runs on its own, when it last ran, how it went. *Configured but never fired* is the state most worth showing: it is indistinguishable from broken until the page says otherwise.

**Recent** — the last five turns, newest first, with durations.

An agent with no schedule and no history renders **exactly** what it rendered before. That is the property most of the tests defend. A box that is empty on every fresh agent is worse than no box: it is prime space on the first page anyone sees, spent saying nothing.

## Two details that are easy to get wrong

**The log is append-only and a session appears twice** — once as `running`, again as `done`. Taking the first line would report every finished run as still in flight. The later line wins.

**A turn that died mid-way stays `running` forever.** That makes it a free stuck-detector, and it is now visible as such, in the accent colour rather than the error colour — it means "look again in a minute", not "something is wrong".

## Durations

`243.0s` for a four-minute extraction makes the reader do the division. Now `4m 3s`, `2h 11m`, and still `1.5s` for short ones — the boundaries are asserted, because the runs this section exists for are precisely the long ones.

## Rendered, not just tested

Built a page with a realistic mix — two scheduled entries (one never run), a turn still running, a 17.6s turn, a 4-minute one — and looked at it:

```
SCHEDULED
  /contract-ledger 检查新合同 (every 15m)          done · just now
  /contract-ledger 本周汇总 (Mon 09:00)              not yet run
RECENT
  列出待人工确认的                                running · 1m ago
  /contract-ledger 检查新合同                      17.6s · 2m ago
  整理一下这个月的合同                              4m 3s · 1h ago
```

The `243.0s` problem was found that way, not by a test — it read wrong on the screen before anything failed.

## Robustness

Reads the schedule through the schedule module rather than reparsing the file, so the page and the scheduler cannot disagree about what an entry means. A torn JSONL line, an unreadable state file or a malformed schedule costs that one section and never the page — the same rule the trust policy loader ended up at in #422.

16 new tests, 2970 existing pass.

## One test bug worth naming

The first version of the ordering test used prompts called `first` and `second`, and failed — because `html.index("first")` was matching `.act:first-child` in the stylesheet. An assertion that succeeds or fails for a reason other than the one it states; same class as the false-green in #504, caught from the other direction. The prompts are now distinctive and the comment says why.

### DIFF
--- a/connectonion/network/host/ws_router/dashboard.py
+++ b/connectonion/network/host/ws_router/dashboard.py
@@ -423,31 +423,67 @@ def _ago(seconds):
     return "a while ago"
 
 
+def _tail_lines(path, wanted, chunk=65536):
+    """The last lines of an append-only file, newest first, read from the end.
+
+    Bounded by the answer rather than by the history. A record here carries the
+    turn's whole message list — 23 KB on a real agent, 85 KB at the top end — and
+    nothing trims the file, so reading it whole to show five rows is a cost that
+    grows forever and is paid on every turn (#526).
+
+    The chunk size is a starting point, not a window: one record can be larger
+    than it, so the walk continues until enough lines are found or the file is
+    exhausted.
+    """
+    try:
+        size = path.stat().st_size
+    except OSError:
+        return []
+    if not size:
+        return []
+
+    lines, buf, pos = [], b"", size
+    with open(path, "rb") as f:
+        while pos > 0 and len(lines) < wanted:
+            step = min(chunk, pos)
+            pos -= step
+            f.seek(pos)
+            buf = f.read(step) + buf
+            parts = buf.split(b"\n")
+            # The first part may be half a record; leave it for the next chunk,
+            # unless we have reached the start of the file and it is whole.
+            buf = parts[0] if pos > 0 else b""
+            complete = parts[1:] if pos > 0 else parts
+            lines = [c for c in complete if c.strip()] + lines
+    return list(reversed(lines[-wanted:] if wanted else lines))
+
+
 def recent_runs(limit=MAX_ACTIVITY_ROWS):
     """The last few turns, newest first, one entry per session.
 
     The log is append-only and a session appears twice — once as ``running``,
     again as ``done``. The later line wins, or the page would report every
-    finished run as still in flight.
+    finished run as still in flight. Reading from the end means the later line
+    is also the one seen first.
     """
     import json as _json
     path = _co() / "session_results.jsonl"
     if not path.is_file():
         return []
+
     latest = {}
-    try:
-        for line in path.read_text(encoding="utf-8").splitlines():
-            line = line.strip()
-            if not line:
-                continue
-            try:
-                rec = _json.loads(line)
-            except ValueError:
-                continue          # a torn write must not cost the whole page
-            if isinstance(rec, dict) and rec.get("session_id"):
-                latest[rec["session_id"]] = rec
-    except OSError:
-        return []
+    # More lines than sessions wanted: each session writes at least twice, and a
+    # torn line costs nothing but itself.
+    for raw in _tail_lines(path, wanted=limit * 4):
+        try:
+            rec = _json.loads(raw.decode("utf-8"))
+        except (ValueError, UnicodeDecodeError):
+            continue          # a torn write must not cost the whole page
+        if isinstance(rec, dict) and rec.get("session_id"):
+            latest.setdefault(rec["session_id"], rec)   # newest first: first wins
+        if len(latest) >= limit:
+            break
+
     runs = sorted(latest.values(), key=lambda r: r.get("created") or 0, reverse=True)
     return runs[:limit]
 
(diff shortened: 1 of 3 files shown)


wu-changxing:
Pushed a second commit: the section reads the tail of the session log rather than all of it (#526).

Found reviewing this against the deployed agent instead of a fixture. Fourteen turns had already produced a **322 KB** log — each record embeds the turn's full message list, 23 KB average, 85 KB at the top end — and `recent_runs()` read the whole thing to render five rows. On connect *and after every run*.

A year of hourly scheduled work, which #524 just made easy, is ~400 MB read and parsed per message. It would never have looked like a bug; the agent would just have got slower to answer, in a path nobody associates with chatting.

Measured on a 44 MB log: **190.9 ms → 1.1 ms**, and now flat in history rather than linear.

Two details the walk has to get right, both asserted:
- the chunk size is a starting point, not a window — a single 85 KB record is normal, so it keeps walking until it has enough lines
- reading newest-first also disposes of the running/done subtlety: from that end, the later line is simply the one seen first

PR #534, opened 2026-08-01
Home is a control panel: a verdict, a schedule, and the future

wu-changxing:
The owner's verdict on the Home page after the schedule feature landed: *"this doesn't look like a control panel at all — no information, no control."* He is right, and the reason a UI review surfaced is worse than a layout problem.

## The page could not express the most important thing about this agent

`chase unpaid invoices` is `every 1d` and last ran 5 days ago. The page called that **failed**. It isn't.

Failed means the run happened and broke. This entry *did not run at all* — `is_due` has been true for four days and nothing served it, which means the scheduler stopped ticking or the host was down. Different fault, different fix, and the page had no way to say it, because it computed **last** run everywhere and **next** run nowhere.

So `next_run()` is not a nicety. It unlocks a third health state we could not previously detect:

| state | meaning | fix |
|---|---|---|
| ok | ran, succeeded, next run in the future | nothing |
| failed | it ran and the turn errored | fix the prompt or the tool |
| **late** | its moment passed by more than 3 ticks and nothing ran | the scheduler or the host is down |

`next_run` deliberately returns a time **in the past** for an entry already owed a run — collapsing that to "now" would hide the only signal that says the scheduler stopped. It mirrors `is_due` exactly and lives beside it, for the reason `scheduled_entries` already gives: a page that computes due-ness on its own is a page that can disagree with the scheduler about what an entry means.

## What the page is now

```
identity          mark · name · tagline · manifest
VERDICT           only when something is wrong
SCHEDULE          dot · name · cadence + last · WHEN IT NEXT FIRES
SKILLS            what you run by hand
RECENT            only what a person typed
address
```

**Schedule above skills**, because the schedule is the part that changed since you last looked; skills only change when you act. Also: schedules are bounded (2–6), skills are not, so putting the unbounded list first buries the urgent one below the fold forever.

**The verdict is absent when healthy.** A permanent green "all systems ok" badge is worse than useless — it teaches the eye to skip the exact region the alarm will one day appear in. Healthy is expressed by the banner being missing and by a column of small green dots down the schedule rows: a column of green means fine, one red means *there*, and neither needs a word read.

**One alert colour.** Red for both bad states, with the words carrying the difference. Red-vs-amber at 11px in a 440px column is a distinction nobody reliably decodes, and both states mean "go look now".

**"Next" is coarse on purpose** — `in 3h`, not `in 2h 47m`. The page is rendered on connect and after a run, and has no JavaScript, so a value can be an hour stale when it is read. `in 3h` survives that; `in 2h 47m` is wrong the moment it is painted, and a countdown that cannot count is worse than none.

## Cards

Sections are cards now. The earlier review killed a *decorative wrapper* — a `#fff` box on a `#fbfbfa` page, one percent apart, so only its border drew and that border duplicated the row hairlines it contained. That is a different question from cards as the organising principle of a panel, which is what a set of separate objects each answering one question actually wants. The page background moved to `#f2f2f0` (`#0e0e0e` dark) so a card is a card rather than a rectangle of border.

## Recent stopped repeating the Schedule

Every scheduled firing has a row in the Schedule card carrying its cadence, status and next run. Listing them again in Recent said the same three things twice in two formats — on the sample agent, **three of four Recent rows were restatements**. Recent now shows only runs that matched no schedule entry, and renders nothing at all when there are none, rather than a heading over an empty card.

The "Schedule problems" section is gone: a config error meaning *this entry will never fire* belongs in the verdict, not in its own list at the bottom.

## Tests

`2993 passed`. Three tests in `test_dashboard_recent_repeats.py` asserted the old contract and are rewritten rather than deleted — including one whose fact moved rather than disappeared: "the latest run is what's reported" is now asserted on the entry's Schedule row, which is where that fact lives.

I also wrote a test asserting that repeated *interactive* prompts collapse, discovered they do not, and deleted it rather than assert a behaviour I had not built.

## Not in this PR — and it is the honest gap

**There is still no control.** `data-ochat-skill` plus `<details>` gives one verb and one fold; *run now* and *pause* are not skills, so they cannot be expressed at all. Faking them would produce dead buttons, which is precisely what the first review had us remove. The next step is a client capability — a `<co-run entry="…">` tag resolved against the loaded schedule server-side, so no prompt text crosses the boundary — and that belongs in oo-chat's bridge alongside `<co-filter>` and `<co-table>`.

### DIFF
--- a/connectonion/network/host/schedule.py
+++ b/connectonion/network/host/schedule.py
@@ -224,6 +224,30 @@ def _clock_due(entry, last_run, now):
     return last_run.astimezone(_zone(entry)) < occurrence
 
 
+def next_run(entry: Entry, last_run: Optional[datetime], now: datetime) -> Optional[datetime]:
+    """The moment this entry is next supposed to fire.
+
+    Deliberately returns a time in the *past* when the entry is already owed a
+    run. The caller wants to know how late, and collapsing that to "now" hides
+    the only signal that says the scheduler has stopped ticking: an entry that
+    runs every day and last ran five days ago has not failed — it did not run at
+    all, which is a different fault with a different fix.
+
+    Mirrors ``is_due`` exactly — same catch-up rule, read forwards — because a
+    page that computes due-ness on its own is a page that can disagree with the
+    scheduler about what an entry means.
+    """
+    if entry.interval is not None:
+        return now if last_run is None else last_run + entry.interval
+    parsed = _parse_at(entry)
+    occurrence = _last_occurrence(entry, now)
+    if parsed is None or occurrence is None:
+        return None                    # never fires; load_entries reports it
+    if is_due(entry, last_run, now):
+        return occurrence              # the one it owes you, in the past
+    return occurrence + (timedelta(days=7) if parsed[0] is not None else timedelta(days=1))
+
+
 def is_due(entry: Entry, last_run: Optional[datetime], now: datetime) -> bool:
     """Whether this entry should run now.
 
--- a/connectonion/network/host/ws_router/dashboard.py
+++ b/connectonion/network/host/ws_router/dashboard.py
@@ -357,7 +357,8 @@ def _address_line(agent_metadata):
     address = agent_metadata.get("address")
     if not address:
         return ""
-    return '<p class="addr">' + escape(str(address)) + '</p>'
+    _, fragments = _starter_templates()
+    return fragments["addr"].safe_substitute(address=escape(str(address))).strip()
 
 
 def render_starter(agent_metadata):
@@ -375,6 +376,12 @@ def render_starter(agent_metadata):
     """
     skills = published_skills(agent_metadata.get("skills") or [])
     page, _ = _starter_templates()
+    # Read once. Three separate calls meant three reads of schedule.yaml and its
+    # state file per render, which is not just wasted work: they are read at
+    # three different instants, so a run landing mid-render could have the
+    # verdict and the row it points at disagreeing about the same entry.
+    scheduled, problems = scheduled_entries()
+    now = _now()
     name = str(agent_metadata.get("name") or "Agent")
     return page.safe_substitute(
         name=escape(name),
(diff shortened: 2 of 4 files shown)


wu-changxing:
Audited on request, and it found four things — one of which no test could have caught.

**1. This PR would have reverted the release.** It was branched from `60e1400`, and `main` has since rolled the version *back* from 1.6.0 to 1.5.10 (#533) — the "X.Y.0 is the stable release" policy. Merging it would have silently restored 1.6.0 along with `VERSIONING.md`. Rebased; the diff is now four files and no version churn.

**2. Python was building markup.** The alert headline emitted `<b>…</b>` inline. `render_starter`'s own docstring says the markup lives in `starter.html` and the function "only decides what goes in them", and this broke that: it put a typographic decision in Python and handed the template a string it was not allowed to escape. The fragment now takes `$subject` and `$verdict`, and the `<b>` is in the HTML where it belongs.

**3. Same violation, pre-existing, mine.** `_address_line` has been concatenating `'<p class="addr">' + … + '</p>'` since #515. Now an `addr` fragment.

After both, Python emits no tags, no classes, no colours and no pixel values — it computes `next_run`, decides which entry is stalled, and formats durations. Everything you can see is in `starter.html`.

**4. The schedule was read three times per render.** `scheduled_entries()` was called by the verdict, the schedule card, and the recent filter — three reads of `schedule.yaml` and `schedule-state.json` per page. Not just wasted I/O: they are read at three different instants, so a run landing mid-render could leave the verdict and the row it points at disagreeing about the same entry. One read, one clock, passed down.

`2993 passed`, and the page renders identically with no unfilled placeholders.

wu-changxing:
Supersedes #517 (now closed): it made the same address-below-the-skills move in isolation; this PR carries that relocation plus the panel rework, so they cannot both merge.

wu-changxing:
Closing after self-review: this branch has drifted into conflict with main, and most of what it does, main now does better.

It was cut 8 days ago. Since then #536, #540 and #542 landed on the same files and solved the same problem — a schedule display that tells the truth — incrementally rather than by rewrite:

| | main today | this PR |
|---|---|---|
| a running entry says it is running | ✅ #540 | ✅ (duplicate) |
| a failed entry says *why* (`reason`) | ✅ #542 | ❌ — I don't have this |
| "done" no longer claims success | ✅ #536 | ✅ (duplicate) |
| **next-run time** | ❌ | ✅ **only here** |
| verdict card / panel rework | ❌ | ✅ only here, but large |

Rebasing would mean forcing a full rewrite on top of three merged fixes and overwriting main's `reason` handling, which is better than mine. That is the wrong trade.

The genuinely new part is `next_run()` — main computes `is_due` but never the forward time, so the page shows only the past. That, and the "verdict when something is wrong" idea, are worth keeping. I'll re-do just those against today's main, small, as its own change, rather than drag this rewrite through the conflict.

The `next_run` reasoning and the panel design are captured for whoever picks it up. Closing this so it stops reading as ready-to-merge.

Issue #557, opened 2026-08-02
The default co-ai skills are for developing connectonion, not for using it


wu-changxing:
`co_ai/skills/builtin/` is loaded into every `co ai` session by default (`skills/loader.py` walks the directory and registers every `SKILL.md`). So its contents are what a fresh user is handed the moment they run `co ai`. Right now that is:

| builtin skill | what it does |
|---|---|
| `commit` | make git commits |
| `review-pr` | review GitHub pull requests |
| `ship-feature` | update tests/docs, then **release to PyPI** |
| `dashboard` | edit the agent's Home page |

**The first three are for maintaining the connectonion repo itself.** A person who installed `co` to run a browser or send email opens `co ai` and is offered "how to ship a connectonion release to PyPI." That is our internal toolbox shipped as the product's front door.

Meanwhile the skills that *are* about using connectonion sit in `useful_skills/`, which is **not** default-loaded:

- `install-connectonion` — take a machine from nothing to a working `co`
- `co-browser` — drive the browser
- `co-mail-and-drive` — email and drive

So the defaults are backwards: the repo-maintenance skills are in the user's face, and the use-the-product skills are the ones you have to go find.

## Also: ship-feature exists twice and has already drifted

`cli/co_ai/skills/builtin/ship-feature/SKILL.md` and `useful_skills/ship-feature/SKILL.md` are two copies, and `diff` reports them **different**. One file, two homes, editing one leaves the other stale — the same fossil problem the dashboard had.

## What to change

**Default builtin should be about connectonion the product, not the codebase.** Candidates, in the spirit of "install, top-up, and the command-line things a user actually reaches for":

- `install-connectonion` — move in (from useful_skills)
- `co-browser`, `co-mail-and-drive` — move in
- a top-up / account-status skill — the "check my balance, here's the recharge link with my key filled in" flow being scoped now

And move the repo-maintenance skills OUT to `useful_skills/` (or a `dev/` set that only loads inside the connectonion project):

- `commit`, `review-pr`, `ship-feature` — these are how *we* work on connectonion, not how a customer uses it
- resolve the duplicate `ship-feature` to one canonical copy while doing so

`dashboard` is the one judgement call — editing your agent's Home is arguably a user action, not a maintenance one; keep it if so.

## Why it matters now

[...]

PR #558, opened 2026-08-03
topup skill: the copy-paste path from 'how much do I have' to a paid link

wu-changxing:
The paid-browser onboarding Aaron scoped: a customer pastes a prompt into their AI (Codex, Claude Code, a domestic workbench), and it tells them their balance, how long it lasts, and hands back a recharge link with their key already in it — no login, no copying a public key by hand.

This is that flow as a builtin skill. It builds nothing new on the backend; it strings together three things that already ship and are already live in production.

## What it does

1. `co --version` — if connectonion is not here, defer to **install-connectonion** (whose `co init` creates the key this skill reads), then continue.
2. Read the account's public key from `~/.co` — one line of Python, never typed by the user.
3. Read the balance from `/api/v1/auth/me` with the token `co` already holds.
4. Say it in money **and** in hours: browser time is $0.10/hour, so `$12.40 → ~124 hours`.
5. Hand over `https://o.openonion.ai/purchase?key=<KEY>` — the page fills in from the key, Stripe returns the credit to it via webhook.

## Verified against production, not assumed

Two things in the skill were wrong on the first draft and are now corrected by testing rather than memory:

- **There is no `co status --balance`.** `co status --help` shows only `--reveal`; `co status` lists credential sources, not money. The skill reads the balance from the API instead.
- **The balance math checks out live:** `/api/v1/tokens/balance` returned `credits_usd 605.0`, `total_spent_usd 579.22`; `/auth/me` returned `balance_usd 25.78` — the difference, exactly. The skill uses `/auth/me` because it does the subtraction.
- **The link resolves:** `o.openonion.ai/purchase?key=0x…` → `200`. The bare `openonion.ai/purchase` is a `404`; the skill pins the `o.` host.

## Where it belongs

`cli/co_ai/skills/builtin/`, so it loads by default. That is deliberate and it is the first concrete instance of #557: the default skill set should be about *using* connectonion — install, top-up, browser, mail — not about developing it (commit, review-pr, ship-feature). This skill is a use-the-product default; it is exactly what belongs there once the maintenance skills move out.

## Notes

- **One balance for everything** — LLM calls, agent email, browser time all draw from the same `credits_usd`; the skill says so and does not imply a separate browser wallet.
- **Never prints the key's full value** except inside the URL the user is meant to click; the summary truncates it.
- If the token predates a key migration (oo-api#67), the recovery step is one `co auth` — which the CLI now self-heals anyway.

### DIFF
--- a/connectonion/cli/co_ai/skills/builtin/topup/SKILL.md
+++ b/connectonion/cli/co_ai/skills/builtin/topup/SKILL.md
@@ -1,0 +1,118 @@
+---
+name: topup
+description: Check the OpenOnion balance and open the recharge page with the user's own key pre-filled. Reports the balance in plain money, says how long it lasts at $0.10/hour of browser time, and hands over a purchase link that needs no login — the key is already in it. If connectonion is not installed yet, install it first (install-connectonion). Use when the user says "top up", "add credit", "recharge", "check my balance", "how much do I have left", "充值", or a paid action failed because the balance is empty.
+tools:
+  - Bash(python *)
+  - Bash(python3 *)
+  - Bash(py *)
+  - Bash(co *)
+  - Bash(curl *)
+  - shell
+  - read_file
+---
+
+# Top-up Skill
+
+Answer one question — *"how much do I have, and how do I add more"* — and hand back a
+recharge link the person can click without logging in, because their own key is already
+in it.
+
+**Assume the person may not be technical.** They came to run a browser or an agent, not to
+learn about public keys. Do the lookup *for* them; say the result in plain money.
+
+## Step 0: Is connectonion here at all?
+
+```bash
+co --version
+```
+
+- **It runs** → go to Step 1.
+- **It does not** → this machine has no `co` yet. Run the **install-connectonion** skill
+  first (it installs `co`, and its `co init` step creates the key this skill needs), then
+  come back here. Do not ask the user to install anything by hand — the install skill does it.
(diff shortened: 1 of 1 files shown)


wu-changxing:
Self-review found one real gap and fixed it: **the skill promised something the purchase page does not show.**

The page (`oo-frontend/app/purchase`) sells fixed top-up packages (~$0.99 / $19.90 / $109.90) and is worded for API credit — grep for `hour`/`browser` on it returns 0. The skill talks $0.10/hour of browser time throughout, then sends the user to a page that mentions neither. Left as-is, they click through and wonder where the hours went.

Fixed in Step 4: it now tells the user plainly that the page sells credit in fixed amounts, all of it lands on the one `credits_usd` balance, and browser time draws from that at $0.10/h. Also dropped 'pick an amount' — the page has packages, not an amount field.

Verified against production, not memory:
- purchase page reads `?key=` (confirmed in `PurchaseClient.tsx`) — the link is right
- `/tokens/checkout` takes `{address, package}` with **no auth** — so `?key=` alone is enough, no signature needed
- the whole command chain runs: key read, `/auth/me` → `balance_usd 24.39 → ~243 hours`, command substitution works under the tool grants

The mismatch between what onboarding says and what the recharge page shows is worth a second look separately — the page could grow a browser-hours line — but that is an oo-frontend change, not this skill's job.

PR #562, opened 2026-08-03
The trust gate answers CONNECT, and that was undocumented

wu-changxing:
`docs/network/websocket-protocol.md` described CONNECT as always answered by CONNECTED, and never covered the onboard handshake. `ONBOARD_SUBMIT` appeared exactly once — inside a console-log example. `ONBOARD_REQUIRED` and `ONBOARD_SUCCESS` appeared nowhere at all.

So the question direction was specified and the answer direction was not.

## This has already cost us

Two client bugs, both traceable to the missing spec:

- The gate was read from **`/info`**. That endpoint is anonymous — it states the agent's policy and tells an admin exactly what it tells a stranger — so the client put a code prompt in front of people who hold the keys. It also failed completely for any agent without a public endpoint, since there is no `/info` to read.
- A refused code was detected by **watching for a second `ONBOARD_REQUIRED`**. Refusals come back as `ERROR`. That branch could never fire, so a wrong code produced no feedback whatsoever — no spinner, no message, nothing.

Neither is a subtle reading error. Both are what the reference implied.

## What this adds

**A "Trust Gate (onboarding)" flow** under Protocol Flows, stating the three things a client gets wrong:

1. **The answer arrives before the first message.** A client that opens the socket eagerly already holds the gate's answer before the reader types. No message needs to be spent discovering it, and no guess from `/info` is needed.
2. **A refusal is `ERROR`, not another `ONBOARD_REQUIRED`.** The gate does not re-ask.
3. **The stashed CONNECT is resumed by the server, not replayed by the client.** Its signature carries a five-minute window, and a human reading a card, finding a code and typing it can outlast that — a resend would be rejected as expired.

**Message reference entries** for `ONBOARD_SUBMIT`, `ONBOARD_REQUIRED`, `ONBOARD_SUCCESS` and `ADMIN_RESULT`, with field tables — including that a failed submit **keeps** the stashed CONNECT, so a retry works on the same socket with no reconnect.

**The CONNECT response table** now says CONNECTED is not its only possible reply, and that the decision is per-caller, made after the signature is verified — which is why an admin or an existing contact never sees a gate.

## Verified, not transcribed

Every claim was checked against a running agent rather than read off the source:

```
  PASS  CONNECT is answered by ONBOARD_REQUIRED, not CONNECTED
  PASS    carries identity, echoing the caller
  PASS    carries methods                          — ['invite_code']
  PASS    no session established (no session_id / status)
  PASS  a wrong code is answered by ERROR, not a second ONBOARD_REQUIRED
  PASS  a retry on the same socket then succeeds
  PASS    ONBOARD_SUCCESS carries level            — contact
  PASS  the server completes CONNECTED itself (no client resend)

  frames: ONBOARD_REQUIRED → ERROR → ONBOARD_SUCCESS → CONNECTED
  8/8
```

The `ERROR → ONBOARD_SUCCESS → CONNECTED` run is what proves the "a failed submit keeps the stash" claim, rather than assuming it from the comment in `session.py`.

One thing deliberately **not** added: `RUNTIME_INPUT`. It looked like a missing frame, but it is pushed into the agent's internal queue by `io.push_runtime_input()` and never crosses the socket — the client-facing half is `RUNTIME_INPUT_ACK`, which is already documented.

Docs only; no code changes.

### DIFF
--- a/docs/network/websocket-protocol.md
+++ b/docs/network/websocket-protocol.md
@@ -18,6 +18,10 @@ If `INPUT` arrives while the session's agent is already running, the server trea
 
 `EXEC` is the direct-execution fast path: it runs one named tool with no LLM, no session, and no history, replying with a single `EXEC_RESULT`. It requires the same CONNECT auth as INPUT, and the tool is gated by the host's `.co/host.yaml` permission whitelist. See [remote-call.md](remote-call.md).
 
+A fourth type, `ONBOARD_SUBMIT`, exists only to answer the trust gate. It is not part of the
+normal path — it appears only when the server interrupts CONNECT with `ONBOARD_REQUIRED`.
+See [Trust Gate](#trust-gate-onboarding).
+
 ```
 ┌────────────────────────────────────────────────────────────────┐
 │                    WebSocket Lifecycle                          │
@@ -28,6 +32,9 @@ If `INPUT` arrives while the session's agent is already running, the server trea
 │   INPUT carries:     just the prompt (session already set)     │
 │                                                                │
 │   Server decides:    new / connected / running                 │
+│                      …or, for a caller the trust policy denies │
+│                      but offers a way in: ONBOARD_REQUIRED,    │
+│                      and CONNECT waits until they pass         │
 │                                                                │
 └────────────────────────────────────────────────────────────────┘
 ```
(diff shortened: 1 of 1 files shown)


wu-changxing:
Measured what this PR closes and what it leaves, by comparing every frame the server actually sends against `docs/network/websocket-protocol.md` on `main`.

**Two frames the server sends are named nowhere in the doc — and both of their responses are documented.**

```
ONBOARD_REQUIRED    0 mentions      (its reply ONBOARD_SUBMIT: 1)
RUNTIME_INPUT       0 mentions      (its ack RUNTIME_INPUT_ACK: 4)
```

Everything else the server sends is described: AGENT_PROFILE, CONNECTED, DASHBOARD_SNAPSHOT, ERROR, EXEC_RESULT, OUTPUT, PING, SESSION_STATUS.

So the file documents the answer to a question it never poses, twice.

**This PR fixes one of them.** It adds 11 references to `ONBOARD_REQUIRED`. That is the one that costs the most: a client that does not know the frame exists waits for `CONNECTED` forever, which is exactly the failure this PR's own text warns about for the refusal case.

It is also the path I have spent three fixes on during this release loop — #671 (an agent offering doors that cannot open), and the two frame-field mismatches in #676 and #677 — while the protocol document never described the flow at all.

**`RUNTIME_INPUT` is still missing after this PR.** Sent from `ws_router/session.py:108`:

```python
existing.io.push_runtime_input({"type": "RUNTIME_INPUT", "id": rid, "prompt": prompt})
```

I have not added it here, because this PR is already editing that file and a second change would collide with something that has been green and waiting since 2026-08-03. Better order: merge this, then document the one frame it does not cover.

Verified separately that this PR's content still matches today's behaviour — the ONBOARD frame shapes it describes are what I measured against a live agent, and it does not touch the frames #676/#677 changed.

PR #575, opened 2026-08-03
fix(trust): a whitelist belongs to one agent, not to the machine

### LINKED ISSUE #507: contacts.txt is shared by every agent on the machine — the fix applied to admins.txt was never applied here
`promote_to_contact()` writes to a machine-global file:

```python
CO_DIR = Path.home() / ".co"                    # trust/tools.py:22
list_path = CO_DIR / f"{list_name}.txt"         # tools.py:152
```

Admins already had this bug and it was fixed — `_admins_file()` (`tools.py:25-33`) puts the list beside the agent's identity, with a comment explaining that a global list compared against a per-project address meant every agent on one machine shared one answer.

**Contacts were left on the old path.** So onboarding a stranger to *one* agent silently makes them a contact of *every* agent on that host, including agents for different projects and different customers.

## Why it matters more than it looks

`fast_rules.evaluate_request` promotes on a successful invite code or payment:

```python
if request_code and request_code in valid_codes:
    promote_to_contact(client_id)
    return 'allow'
```

The promotion is durable and silent. An operator running three agents — one internal, one for a customer, one experimental — has one shared contact list across all three, and nothing in the CLI shows it.

The invite code is not a secret either: `host()` prints it in the startup banner (`Invite: OpenOnion`), so it is routinely visible in screenshots, screen shares, and chat.

## Fix

Scope contacts per-agent exactly as admins are, reusing `_admins_file`'s reasoning. Migration: an existing global `contacts.txt` should be read as a fallback (so nobody loses their contacts on upgrade) but never written to again.

Worth auditing the other lists in the same pass — `whitelist.txt` and `blacklist.txt` go through the same `CO_DIR` path and probably deserve the same treatment, though a global blocklist is arguably correct and should be a deliberate decision rather than an inherited default.

## Found

Reviewing #505 (agent sessions over email), where the same code would be reachable by anyone who can send mail. The bug is not email-specific — it affects the WebSocket path today.

wu-changxing:
Closes #507.

`admins.txt` was moved out of `~/.co/` and beside the agent's own identity, because one box hosting two agents meant that making someone an admin of one made them an admin of both. **`whitelist`, `contacts` and `blocklist` were left on the old global path.** Same bug, and `whitelist` is the file that grants.

```python
CO_DIR = Path.home() / ".co"     # module-level, shared by every agent on the machine
```

`is_whitelisted()` is a real allow. Promoting an address while poking at a throwaway agent promoted it on the production agent running on the same box — and nothing in either agent's directory recorded that it had happened.

## Proven before it was fixed

Four tests, two agent directories, one machine. All four were red:

```
AssertionError: an address whitelisted on one agent is whitelisted on
                every agent sharing the machine
```

Blocking is included even though it leaks in the safe direction. An operator who blocks someone from a public agent has not asked to cut them off from a private one they also run.

## Upgrading, without a lockout

The move means an operator's existing `~/.co/whitelist.txt` stops applying. Someone finds a colleague locked out and there is nothing anywhere explaining why.

Those entries are not read — reading them is the bug this moves away from — so the read path says, once per list per process, where they went:

```
[trust] /Users/x/.co/whitelist.txt is no longer read — these lists now live
        beside each agent, in its own .co/. Copy the lines you still want
        into /srv/agent/.co/whitelist.txt.
```

Two tests cover it: it is said when a legacy file has entries, and nothing is said when there is none. A warning that fires on every check is noise that trains people to ignore it.

## The fixtures were the tell

Eight test fixtures monkeypatched the module-level `CO_DIR` so tests could not see each other's lists. That is exactly the shape of the defect — isolation required reaching into a process global. They now `chdir` into an agent directory, which is also what a second agent on the same machine does.

`co trust` was reading the same constant to build its own paths, so it now goes through the same helper. Otherwise the CLI would list one file while the agent read another.

3078 unit tests pass.

### DIFF
--- a/connectonion/cli/commands/trust_commands.py
+++ b/connectonion/cli/commands/trust_commands.py
@@ -30,7 +30,7 @@ from rich.console import Console
 from rich.table import Table
 
 from ...network.trust.tools import (
-    CO_DIR,
+    list_file,
     get_level,
     promote_to_contact,
     promote_to_whitelist,
@@ -48,7 +48,7 @@ console = Console()
 
 def _read_list(list_name: str) -> list[str]:
     """Read entries from a list file."""
-    list_path = CO_DIR / f"{list_name}.txt"
+    list_path = list_file(list_name)
     if not list_path.exists():
         return []
     content = list_path.read_text(encoding='utf-8')
@@ -103,7 +103,7 @@ def handle_trust_list():
         console.print("  [dim]Empty[/dim]")
     console.print()
 
-    console.print(f"[dim]Lists stored in: {CO_DIR}[/dim]")
+    console.print(f"[dim]Lists stored in: {list_file('whitelist').parent}[/dim]")
 
 
 def handle_trust_level(address: str):
--- a/connectonion/cli/commands/trust_commands.py
+++ b/connectonion/cli/commands/trust_commands.py
@@ -58,6 +58,18 @@ def _read_list(list_name: str) -> list[str]:
 
 def handle_trust_list():
     """List all trust lists."""
+    # These lists belong to one agent, so being in the wrong directory shows
+    # four empty sections — which reads as "my whitelist was wiped" rather than
+    # "you are not standing in an agent". Say which it is.
+    agent_dir = list_file("whitelist").parent
+    if not agent_dir.is_dir():
+        console.print()
+        console.print(f"[yellow]No agent here[/yellow] — {agent_dir} does not exist.")
+        console.print("[dim]Trust lists belong to one agent. Run this from its "
+                      "directory, or 'co init' to make one.[/dim]")
+        console.print()
+        return
+
     contacts = _read_list("contacts")
     whitelist = _read_list("whitelist")
     blocklist = _read_list("blocklist")
--- a/connectonion/network/trust/tools.py
+++ b/connectonion/network/trust/tools.py
@@ -19,7 +19,38 @@ from pathlib import Path
 from typing import List, Callable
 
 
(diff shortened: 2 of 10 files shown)


wu-changxing:
## Reviewing my own change: the load-bearing line nobody was watching

Scoping the lists to the agent means resolving them from `Path.cwd()`. That is a real dependency and it is invisible — so I checked it against the live deployment rather than assuming:

```
/etc/systemd/system/naturewill.service
  WorkingDirectory=/srv/naturewill

MainPID 256210 → /proc/256210/cwd → /srv/naturewill
```

Correct in production. But nothing pinned it. `WorkingDirectory={SRV}/{agent}` was one uncommented line in a unit template with no test on it, and dropping it would resolve every list under `/` — quietly empty. The whitelist stops granting, the blocklist stops blocking, and nothing anywhere says why.

Two tests added on `_unit_text()`, so the coupling now fails loudly at the place that would break it.

## One asymmetry worth naming

`admins.txt` is stronger than this. Its callers pass an explicit `co_dir`:

```python
admins_file = _admins_file(co_dir)     # supplied by the host
```

whereas `list_file()` has no such parameter and leans on the working directory. That is why the pin above matters, and it is the honest limit of this PR: it makes the three lists as scoped as the identity they sit beside, not as explicitly addressed as the admin list.

Threading `co_dir` through `is_whitelisted` / `is_contact` / `block` means changing the `TrustAgent` method signatures those go through, which is a wider change than the leak needs. Worth doing, not worth bundling here.

Issue #230, opened 2026-07-22
Windows: GBK/cp936 default encoding breaks `co ai` and subprocess-based tools (UnicodeDecodeError/UnicodeEncodeError)


wu-changxing:
## Summary

On Windows, users repeatedly report that after installing ConnectOnion the CLI (especially `co ai`) fails to run commands, or crashes mid-run with an encoding error. The root cause is that **Chinese/Japanese/Korean Windows uses a legacy locale codepage (GBK / cp936, Shift-JIS, etc.) as the default text encoding — not UTF-8** — and our subprocess calls rely on that default instead of forcing UTF-8.

This is a tracking issue. The fix (forcing UTF-8 on all subprocess I/O) will follow.

## Why it happens

Python's `subprocess` with `text=True` (or `universal_newlines=True`) decodes the child process's stdout/stderr using `locale.getpreferredencoding(False)`. On a machine with the system locale set to Chinese, that is **GBK (cp936)**, on Japanese it is **cp932**, etc. — *not* UTF-8.

So when a spawned command emits UTF-8 bytes (or any byte that isn't valid in the local codepage), the reader raises:

```
UnicodeDecodeError: 'gbk' codec can't decode byte 0x?? in position ...: illegal multibyte sequence
```

The reverse also happens: a child Python process printing emoji/UTF-8 to a GBK console raises `UnicodeEncodeError: 'gbk' codec can't encode character ...`.

Note: v0.3.5 already fixed **file I/O** on Windows by adding `encoding='utf-8'` everywhere (see `docs/windows-support.md`). This issue is the remaining gap: **subprocess pipe I/O**, which was not covered by that pass.

## Affected code

All of these use `text=True` (or read pipes) **without** an explicit `encoding`/`errors`, so they inherit the GBK default on affected machines:

- `connectonion/cli/co_ai/tools/background.py:97` — `subprocess.Popen(..., text=True)` feeding the background reader thread (`_read_output`). This is the main path for `co ai` running builds/tests/servers, and a decode error here kills the reader thread silently.
- `connectonion/useful_tools/shell.py:54` and `shell.py:81` — `Shell.run()` / `run_in_dir()` (`subprocess.run(..., text=True)`), the cross-platform shell tool agents use on Windows.
- `connectonion/cli/co_ai/context.py:108` (and the `git status` / `git log` calls that follow at `:118`, `:128`) — collecting git context with `text=True`; branch/commit messages containing non-ASCII break decoding.
[...]

PR #584, opened 2026-08-03
fix(browser): the clipboard keeps its characters on Windows

### LINKED ISSUE #277: Windows: humanize `_paste` puts mojibake on the clipboard for CJK (clip.exe reads UTF-8 bytes as the OEM codepage)
## Summary

The browser humanize typing path pastes CJK runs through the OS clipboard, and on Windows it writes UTF-8 bytes to `clip.exe`. `clip.exe` interprets its stdin using the console **OEM code page** (cp936/GBK on Chinese Windows, cp932 on Japanese, cp437/1252 on Western), **not UTF-8** — so CJK text lands on the clipboard as mojibake and gets pasted into the page wrong.

Not a crash; a correctness bug. Windows-only, CJK-only. Filed separately from the UTF-8 encoding fix (#230) because the correct fix needs validation on a real Windows box.

## Where

`connectonion/useful_tools/browser_tools/humanize.py`

```python
# _clipboard_cmds() on Windows:
if system == "Windows":
    return (["clip"], ["powershell", "-NoProfile", "-Command", "Get-Clipboard"])

# _paste():
subprocess.run(set_argv, input=text.encode())   # text.encode() = UTF-8 bytes -> clip.exe
```

`text.encode()` defaults to UTF-8; `clip.exe` decodes stdin with the OEM code page. For a CJK run like `中文`, the UTF-8 bytes `E4 B8 AD E6 96 87` are re-interpreted under GBK → garbage characters on the clipboard.

## Why it actually bites (not just theoretical)

`type_text()` routes CJK runs to `_paste` **first**, using `_type_ime` (CDP composition) only as a fallback when paste is *blocked*:

```python
# type_text():
if not _paste(page, run):
    _type_ime(page, run)
```

`_paste` returns success based on whether the field grew (`_active_text_len(page) >= before + len(text)`). A mojibake paste can still grow the field, so `_paste` may report success and the IME fallback never fires — the field ends up with wrong characters. (Read-back save/restore of the user's own clipboard is fine — it round-trips bytes at lines using `capture_output`/`input=saved`.)

## Suggested direction (needs real-Windows testing)

The read side already uses PowerShell (`Get-Clipboard`); make the write side symmetric and Unicode-safe. Candidates, in rough order of preference:

1. Write via PowerShell `Set-Clipboard` with the input stream forced to UTF-8, e.g. `powershell -NoProfile -Command "[Console]::InputEncoding=[System.Text.Encoding]::UTF8; $input | Set-Clipboard"` and pipe `text.encode('utf-8')`. Verify quoting and that a trailing newline isn't added.
2. Or encode to the code page `clip.exe` actually consumes on that machine (the OEM codepage), rather than UTF-8 — but this is locale-specific and less robust than (1).

Whatever the fix, validate on an actual non-UTF-8 Windows locale that:
- `_paste("中文")` results in `中文` (not mojibake) in the focused field, and
- the Latin/ASCII paste path and the user-clipboard save/restore still work.

A `windows-e2e` CI assertion (paste a CJK string into a text input and read it back) would lock this in — note the hosted runner is a Western locale, but cp437/1252 can't represent CJK either, so UTF-8-bytes-to-clip.exe mojibake should still reproduce there.

## Scope
- Not caused by, and separate from, the subprocess/file UTF-8 fix (#230 / PR #233). That fix does not touch this clipboard write.

wu-changxing:
Closes #277.

The humanize typing path pastes CJK runs through the OS clipboard — a real user pastes Chinese far more often than they hand-type it. On Windows it piped UTF-8 into `clip.exe`:

```python
subprocess.run(["clip"], input=text.encode())
```

`clip.exe` decodes its stdin with the console **OEM code page** — cp936 on Chinese Windows, cp932 on Japanese — not UTF-8. So `中文` landed on the clipboard as mojibake and was pasted into the page wrong.

**Not a crash.** The automation reported success and typed the wrong characters.

## The fix

Windows goes through PowerShell with the text carried *inside* the command as base64 — ASCII whatever the console is set to — and decoded as UTF-8 on the other side:

```python
payload = base64.b64encode(text.encode("utf-8")).decode("ascii")
script = (f"$b=[Convert]::FromBase64String('{payload}');"
          "Set-Clipboard -Value ([Text.Encoding]::UTF8.GetString($b))")
```

Reading back pins `[Console]::OutputEncoding` for the same reason. No code page is consulted in either direction. macOS and Linux are unchanged — there the text goes on stdin, where UTF-8 is what these tools want.

## The validation this was waiting for

> Filed separately from #230 because the correct fix needs validation on a real Windows box.

There are Windows runners in CI. The round-trip tests run there — put CJK on the clipboard, read it back, confirm the previous contents are restored — and skip elsewhere, which is honest about where they mean anything.

## The limit this fix buys, which is the same shape as the bug

Carrying the text in the command means a length limit: Windows caps a command line near 8191 characters and base64 inflates UTF-8 by a third. Past that the shell truncates and a **fragment** lands on the clipboard — silently, on Windows, only for long text.

`_paste` already falls back to the IME path when the clipboard route is unavailable, so a run that would overflow declines the route rather than half-using it. Four tests cover the boundary, including one confirming other platforms have no such limit — nothing is measuring a command line there.

3134 unit tests pass.

### DIFF
--- a/connectonion/useful_tools/browser_tools/humanize.py
+++ b/connectonion/useful_tools/browser_tools/humanize.py
@@ -11,6 +11,7 @@ LLM-Note:
 """
 
 import math
+import base64
 import platform
 import random
 import shutil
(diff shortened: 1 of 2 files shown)


wu-changxing:
## Correcting myself: those tests do not run anywhere

The description says the round-trip tests "run on the Windows runners". **They do not.** I checked the run instead of assuming, and they were skipped in every job:

```
tests/unit/test_the_clipboard_keeps_its_characters.py::TestTheRoundTripOnWindows::
    test_what_goes_on_the_clipboard_comes_back[中文测试]   SKIPPED
    test_what_goes_on_the_clipboard_comes_back[hello 中文 world 日本語 🎉]  SKIPPED
    test_the_previous_clipboard_is_restored               SKIPPED
```

`test (3.10…3.13)` runs `tests/unit` on Linux, where the file skips itself. `windows-browser` runs two named files:

```yaml
pytest tests/unit/test_transport.py tests/e2e/cli/test_browser_daemon.py
```

and this was not one of them. So the tests existed, reported nothing, and the fix they were written to validate was unvalidated **while looking covered** — the same shape as the bug they are about: something reporting success without doing the work.

The fix is one step added to `windows-browser`, with `-rs` so a skip there is visible rather than silent. It is committed locally but **cannot be pushed**: touching `.github/workflows/` needs a `workflow` scope this token does not have.

```
! [remote rejected] refusing to allow an OAuth App to create or update
  workflow `.github/workflows/tests.yml` without `workflow` scope
```

Aaron needs to run `gh auth refresh -s workflow` once, then this pushes.

**Until then, treat the Windows behaviour of this PR as reasoned, not measured.** The code path is argued from how `clip.exe` and PowerShell handle encodings, and the platform-independent tests (what argv we build, that it carries no non-ASCII, that a long run declines the route) do run and do pass. The round trip on a real Windows clipboard is the part still owed.

PR #591, opened 2026-08-03
fix(announce): an agent publishes addresses another machine can reach

wu-changxing:
Found by reading what the relay actually serves about a live production agent, rather than by reading the code that builds it:

```
$ curl https://oo.openonion.ai/api/agents/0xfae4…
endpoints:
  http://localhost:8001          ws://localhost:8001/ws
  http://10.152.0.16:8001        ws://10.152.0.16:8001/ws
  http://34.151.137.226:8001     ws://34.151.137.226:8001/ws
```

Anyone who asks the relay gets that list.

**`localhost` is not a place another machine can go.** Published to a public directory it is worse than useless: a browser client walking the list probes port 8001 on *the reader's own machine*, and an address check is the only thing that stopped it talking to whatever answered.

That was fixed on the client side in openonion/connectonion-react#6 (0.3.2) by refusing http endpoints from an https page. This is the same bug at the source — the address should not have been advertised in the first place.

## The private address stays, on purpose

Two agents in one VPC or on one LAN can use `10.x`, and the relay would otherwise carry a connection that did not need carrying. That is a judgement call rather than an oversight, so it is written into the code comment and pinned by a test that would fail if someone "tidied" it away later.

## An old test said the opposite

```python
"""Local hosts keep the existing localhost/IP endpoint behavior."""
assert get_endpoints(8000) == [
    "http://localhost:8000", "ws://localhost:8000/ws", …
]
```

It was accurate about what the code did. It now says why that changed.

3179 unit tests pass.

---

Housekeeping from the same round: the test run hit `No space left on device`, which was my own doing — a dozen worktrees and several throwaway venvs from the last few days of this loop. Cleaned up, 11 GB reclaimed. Worth mentioning because the failure looked exactly like a code regression for a few minutes.

### DIFF
--- a/connectonion/network/announce.py
+++ b/connectonion/network/announce.py
@@ -38,6 +38,13 @@ def get_ips() -> List[str]:
     return ips
 
 
+def _is_loopback(host: str) -> bool:
+    """Whether this address only means anything on the machine that has it."""
+    if host in ("localhost", "::1"):
+        return True
+    return host.startswith("127.")
+
+
 def get_endpoints(port: int) -> List[str]:
     """Get all endpoints as full URLs (http and ws for each IP)."""
     domain = os.getenv("AGENT_PUBLIC_DOMAIN", "").strip().rstrip("/")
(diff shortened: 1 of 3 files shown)


wu-changxing:
## Correcting the frequency I claimed

I wrote that the loopback probe happens "on every page load". **It does not.** Measured again against production, three consecutive loads:

```
run 1: mixed-content=0  probes of :8001=0
run 2: mixed-content=0  probes of :8001=0
run 3: mixed-content=0  probes of :8001=0
```

Earlier today the same page produced 8 mixed-content errors and three failing requests to `:8001`, which is where the claim came from. The relay record is unchanged — it still publishes `http://localhost:8001` — so the difference is timing on the client, not the data:

```
$ curl https://oo.openonion.ai/api/agents/0xfae4…
  http://localhost:8001          ws://localhost:8001/ws
  http://10.152.0.16:8001        ws://10.152.0.16:8001/ws
  http://34.151.137.226:8001     ws://34.151.137.226:8001/ws
```

The direct-endpoint resolution races the relay path, and when the relay connects first the probe never runs. So the accurate statement is **intermittent, depending on which path wins**, not "every load".

The fix stands on its own — a public directory should not carry an address that only means something on the machine that published it, and a client should not be reachable into the reader's own ports even sometimes. But "sometimes" is what I should have written, and an overstated frequency is the kind of thing that makes the next person distrust the rest of the report.

Issue #438, opened 2026-07-31
AGENT_CONFIG_PATH is written as an absolute path, so a project only works on the machine that created it


wu-changxing:
`AGENT_CONFIG_PATH` is written as an absolute path, so a project only works on the machine that created it.

## What happens

`cli/commands/create.py`:

```python
lines_to_add.append(f"AGENT_CONFIG_PATH={Path.home() / '.co'}\n")
```

A project created on macOS carries this in `.env`:

```
AGENT_CONFIG_PATH=/Users/changxing/.co
```

Deploy that project to a Linux server — by `co deploy --to`, by rsync, by cloning the repo
on another machine — and the variable still points at `/Users/changxing/.co`, which does
not exist there.

## What it silently changes

Every OAuth-backed tool resolves its credentials through it:

```python
env_file = Path(os.getenv("AGENT_CONFIG_PATH", os.path.expanduser("~/.co"))) / "keys.env"
```

`useful_tools/gmail.py`, `gdrive.py`, `outlook.py`, `google_calendar.py`,
`microsoft_calendar.py`, `synology.py` — all the same line.

Because the variable *is* set, the `~/.co` default never applies. The lookup goes to a
path that cannot exist, finds nothing, and the tool behaves as if the account were simply
not connected. No error names the path.

## Why it is not just cosmetic

The failure appears at the worst moment: the agent runs, answers, and is only missing the
one capability the operator deployed it for. The obvious readings are "OAuth expired" or
"I forgot to connect the account" — both send the operator to re-authorise on the server,
which writes credentials into `~/.co/keys.env` and still does not fix it, because the
variable points somewhere else entirely.

It also puts the deploying user's home directory path into a file that travels with the
project, which is a small information leak in any repo that isn't private.

## What would fix it

Don't pin an absolute path into a portable file. Either drop the variable from generated
`.env` files and let the `~/.co` default do its job, or resolve it relative to the current
user's home when the recorded path does not exist. If it is kept, a missing directory
should say so rather than degrade to "no credentials".

Related: #393 (the `.env` a project carries is a copy of the operator's global secrets)
and #396 (identity values are copied into `.env` rather than resolved), both of which come
from the same "record the creating machine's state into the project" pattern.

## Environment

connectonion 1.5.3 / 1.5.4, project created on macOS, deployed to Ubuntu 22.04.

PR #604, opened 2026-08-03
AGENT_CONFIG_PATH is still travelling in a project .env

wu-changxing:
#438 removed the line that wrote `AGENT_CONFIG_PATH=/Users/someone/.co` into a new project's `.env`, and `co deploy` learned to rewrite it for the server. It is still there. A real `co init` on this machine, today, writes:

    AGENT_CONFIG_PATH=/Users/changxing/.co

It arrives by a route neither fix covers. `ensure_global_config()` writes AGENT_CONFIG_PATH into `~/.co/keys.env` — correct, that file never leaves the machine — and both `co init` and `co create` then copy keys.env into the project `.env` key by key.

## Why it matters

The failure is the one #438 described. A project made on one machine and run on another — cloned, rsynced, handed to a colleague — carries an absolute path to a home directory that does not exist there. `gmail.py`, `outlook.py`, `gdrive.py`, `synology.py` and both calendars all resolve `os.getenv("AGENT_CONFIG_PATH", "~/.co")`, so they look for keys.env inside a directory that cannot exist and find nothing. Unset, the fallback is right on every machine.

`co deploy` already strips the value on the way to a server. That is the same judgement — it is not portable — applied one step too late.

## The fix

`describes_this_machine()` alongside the existing `is_personal_account_credential()` in `env_inheritance.py`, applied where each command reads keys.env.

`co create` needed it too. Its comment says "AGENT_CONFIG_PATH is deliberately not written here", and that was true of the block it sits in — the copy loop twelve lines above was writing it anyway.

## Tests

`tests/unit/test_the_machine_path_does_not_travel.py`. The rule is checked as a unit; what the rule protects is checked by running the real `handle_init()` and `handle_create()` and reading the `.env` they produce.

The previous fix here passed its tests and still wrote the value, because those tests checked the source for a template line while the value came through keys.env. So both real-path tests were confirmed to fail on unpatched code (2 failed, 9 passed) and pass after.

Full unit suite: 3266 passed, 4 skipped. A real `co init` now writes AGENT_ADDRESS, AGENT_EMAIL, CO_INVITE_CODE, GEMINI_API_KEY, IS_EMAIL_ACTIVE, OPENONION_API_KEY — and no home directory.

### DIFF
--- a/connectonion/cli/commands/create.py
+++ b/connectonion/cli/commands/create.py
@@ -316,13 +316,17 @@ def handle_create(name: Optional[str], ai: Optional[bool], key: Optional[str],
     # Always copy from global keys.env (includes AGENT_ADDRESS, AGENT_EMAIL, and API keys)
     if global_keys_env.exists() and global_keys_env.stat().st_size > 0:
         # Copy global keys to project
-        from .env_inheritance import is_personal_account_credential
+        from .env_inheritance import (
+            describes_this_machine,
+            is_personal_account_credential,
+        )
+
+        def inherited(line: str) -> bool:
+            key = line.strip().split('=', 1)[0]
+            return not (is_personal_account_credential(key) or describes_this_machine(key))
 
         with open(global_keys_env, 'r', encoding='utf-8') as f:
-            env_content = "".join(
-                line for line in f
-                if not is_personal_account_credential(line.strip().split('=', 1)[0])
-            )
+            env_content = "".join(line for line in f if inherited(line))
 
         # AGENT_CONFIG_PATH is deliberately not written here. Every tool that
         # reads it already falls back to ~/.co on the machine it is running on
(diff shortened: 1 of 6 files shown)


wu-changxing:
CI caught what my local run could not: two e2e tests asserted the opposite of this PR.

```
tests/e2e/cli/test_cli_create.py::test_create_adds_agent_config_path_to_env  FAILED
tests/e2e/cli/test_cli_init.py::test_init_creates_agent_config_path_in_env   FAILED
```

I had run `pytest tests/unit`; these live under `tests/e2e/cli` and are in CI's set (`pytest tests/ -m "not slow and not real_api and not network"`).

They are the bug written down as an expectation. Both say only "Test that init creates AGENT_CONFIG_PATH in .env file" — a restatement of what the code did, with no reason. The reasons on record all run the other way: #438, `deploy_to_server.py` (rewrites it for the server), `deploy_commands.py` (rewrites it to /app/.co), and `create.py`'s own comment saying the value is deliberately not written.

So both now assert the absence, and say in their docstrings that they used to assert the presence and why that changed. 3701 passed locally on the CI command.

PR #605, opened 2026-08-03
co pays for the OpenAI and Anthropic SDKs before it parses an argument

### LINKED ISSUE #588: co --version takes 9.5s: the CLI imports openai and anthropic to print a version
## `co --version` takes 9.5 seconds

Measured on a clean venv with the wheel built from `main` (32e7ad4), Python 3.10:

```
$ time co --version
co 1.5.11
real    0m9.505s
```

Every `co` command pays it, including the ones that have nothing to do with a model — `init`, `status`, `server`, `browser`, `deploy`.

## Where it goes

```
import connectonion        10.69s
connectonion.cli.main      11.53s
typer + rich alone          0.45s
```

`-X importtime` on `import connectonion`:

```
9049.9 ms  connectonion
7186.9 ms  connectonion.core
7182.8 ms  connectonion.core.agent
7064.8 ms  connectonion.core.llm
4133.6 ms  openai
2630.8 ms  anthropic
1438.9 ms  connectonion.useful_tools
```

`connectonion/__init__.py` imports `core.Agent` eagerly → `core.llm` → `openai` and `anthropic` at module level.

`cli/main.py` imports exactly one thing from the package:

```python
from .. import __version__
```

Everything else in the CLI is already lazy — the command modules are imported inside their own functions. **That single line is the whole cost.**

## A fix that does not work, so nobody tries it twice

Moving `__version__` into `connectonion/_version.py` and importing `from .._version import __version__` does nothing: importing a submodule runs the package `__init__` first. Verified, not assumed:

```
assert {'anthropic', 'google', 'openai'} == set()
```

## What would work

1. **PEP 562 lazy `__init__`** — `def __getattr__(name)` resolving the ~30 re-exports on first attribute access. `import connectonion` becomes cheap; `from connectonion import Agent` still works. This is the real fix.
2. **Lazy provider SDKs in `core/llm.py`** — 22 references, most of them in `except openai.APIError` clauses, so each needs a function-local import before its `try`. Narrower blast radius than (1) but more edits, and it is the hot path for every LLM call.

## Why I filed this instead of fixing it

(1) changes the package's public import surface, and any subtle breakage lands on every user. It deserves its own review rather than being merged during a release sweep — especially while the Windows CI step from #584 is still blocked on a token scope, so the suite I would validate against is narrower than usual.

A ten-second `co --version` is not a correctness bug and does not block 1.6.0. It is, on the other hand, the first thing a new user experiences.

Found while checking that a fresh install of the built wheel works before the release: it does — `pip install` → `co init` both succeed.

wu-changxing:
Part of #588. Every command handler in `cli/main.py` is imported inside its function, and the module docstring claims "fast startup (lazy imports)". Two things undid it.

## What it was spending

Measured with `python -X importtime -c "import connectonion.cli.main"`:

    connectonion.cli.main       2.26s
      connectonion              2.23s
        connectonion.core.llm   1.38s
          openai                1.03s
        connectonion.useful_tools 0.64s

`core/llm.py` imported `openai` and `anthropic` at module scope. Nothing in it needs either until a provider is constructed or an error is being classified — but `connectonion/__init__.py` reaches `core.llm`, and importing any submodule imports the package, so `co --version`, `co status` and every tab-completion paid for both SDKs first.

## After

    connectonion.cli.main       1.06s        (-1.2s)

Wall clock for `co --version` on this machine: 3.6–4.1s before, 2.4–2.5s after.

The cost is not paid by `--version` alone. `co` is driven in a loop by Claude Code and codex, so a second is charged per invocation.

## How

`import openai` / `import anthropic` moved into the ten functions that use them. Not PEP 562 on the module: `__getattr__` intercepts attribute access on the module object, not bare global names inside its own functions, so `self.client = openai.OpenAI(...)` would have raised `NameError` — it does, I tried it first.

One test patched `connectonion.core.llm.openai.OpenAI`, an attribute that no longer exists; it patches `openai.OpenAI` now.

Also here, both small and in the same area:

- `connectonion/_version.py` — the version string on its own, so `cli/main.py` can read it with `from .._version import __version__`. Kept in step with pyproject.toml by a test. (Does not by itself help startup: importing any submodule imports the package. It is the piece that makes the docstring's promise reachable at all.)
- `connectonion/cli/__init__.py` carried `__version__ = "0.0.1b5"` from the first beta, with a docstring saying the version command read it. Nothing read it. Removed.

## Tests

`tests/unit/test_the_cli_starts_without_the_sdks.py` — importing `connectonion.cli.main` in a fresh interpreter loads neither SDK, the version still prints, and the string agrees with pyproject.toml. Confirmed failing before the change (8 failed), passing after.

Verified against a real call, not only the mocks: `llm_do(..., model='co/gemini-2.5-flash')` returns `pong`, `openai` is absent from `sys.modules` after import and present after the call.

Full unit suite: 3264 passed, 4 skipped.

## What this does not fix

#588 says 9.5s; it is 2.4s now and this takes ~1.2s of that. The rest is `useful_tools` (0.86s, mostly the Google client in gmail.py), `tui.chat` (0.47s) and `requests` (0.37s), all imported eagerly by `connectonion/__init__.py`. Making those lazy means PEP 562 on the package's public surface, which is a larger change than this one. Leaving #588 open with the numbers.

### DIFF
--- a/connectonion/__init__.py
+++ b/connectonion/__init__.py
@@ -10,7 +10,8 @@ LLM-Note:
 ConnectOnion - A simple agent framework with behavior tracking.
 """
 
-__version__ = "1.5.11"
+# Single source, in a module the CLI can import on its own — see _version.py.
+from ._version import __version__
 
 # Auto-load .env files for the entire framework
 import os as _os
--- a/connectonion/_version.py
+++ b/connectonion/_version.py
@@ -1,0 +1,11 @@
+"""The version string, on its own, so reading it costs nothing.
+
+`co --version` used to import the whole package — and with it the OpenAI and
+Anthropic SDKs, the TUI and every useful_tool — to print six characters. Every
+command handler in the CLI is already imported inside its function; this module
+is what lets the entry point keep that promise.
+
+Kept in step with pyproject.toml by a test.
+"""
+
+__version__ = "1.5.11"
--- a/connectonion/cli/__init__.py
+++ b/connectonion/cli/__init__.py
@@ -4,10 +4,8 @@ LLM-Note:
   Dependencies: none | imported by [cli/main.py, setup.py] | no direct tests
   Data flow: pure metadata (version string)
   State/Effects: no state
-  Integration: exposes __version__ for CLI about/version commands
+  Integration: namespace only — the version lives in connectonion/_version.py
   Performance: trivial
   Errors: none
 ConnectOnion CLI module.
 """
-
-__version__ = "0.0.1b5"
--- a/connectonion/cli/main.py
+++ b/connectonion/cli/main.py
@@ -28,7 +28,10 @@ import typer
 from dotenv import load_dotenv
 from rich.console import Console
 
-from .. import __version__
+# From _version, not from the package: `from .. import __version__` pulled in
+# the OpenAI and Anthropic SDKs before the CLI had parsed an argument, which
+# is what every command handler being imported inside its function was for.
+from .._version import __version__
 
 # Load both env files for all CLI commands. keys.env stays first — that is
 # already the CLI's effective precedence, since commands that load .env do so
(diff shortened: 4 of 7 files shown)


wu-changxing:
Rebased onto main (#603 touched the same lines in `core/llm.py`), and the rebase caught something worth writing down.

Five tests failed after it, all mine:

```
assert '1.5.11' in 'co 1.5.10'
ModuleNotFoundError: No module named 'connectonion._version'
AssertionError: co pays for openai before parsing an argument
```

The probes shell out to `python -c` to measure a fresh interpreter, and they ran with whatever working directory the test had. So they imported `connectonion` **from site-packages** — the installed 1.5.10 wheel — and reported on that, not on this branch. They passed before only because the two versions happened to agree; the rebase made them disagree and the tests started describing a wheel.

That is the same shape as three earlier problems in this release: a check that is green because it is not reaching the thing it claims to check.

Fixed by pinning `cwd` and `PYTHONPATH` to the repo, and by asserting on the subprocess's exit code instead of letting a `CalledProcessError` decide. Then re-confirmed the probes still go red on unpatched `core/llm.py` (4 failed) with the pin in place — so they now fail for the reason they claim.

CI command locally: 3696 passed, 9 skipped.

Issue #434, opened 2026-07-31
A trust denial is invisible from both ends — the client shows nothing and only the 'no' is logged


wu-changxing:
A trust denial is invisible from both ends: the client shows nothing, and the log only records the "no".

## What happened

Deploying an agent to a server and opening it in oo-chat, the page showed the agent as
`online` and then — nothing. No Home pane, no model/trust/version line, no error, no
prompt. An empty chat against a live agent.

The agent was denying every request. Nothing on the client said so.

The only trace was on the server, in the agent's own log file:

```
[FAST_RULES] Evaluating client_id=0xb2df62c9094c42eb54... config={'allow': ['admin', ...], 'default': 'deny'}
[FAST_RULES] No match, using default=deny
[FAST_RULES] Returning 'deny' (default)
```

Diagnosing it took an hour of `ssh` + `grep` against a file the operator would not
normally think to open, for a symptom whose obvious reading is "the dashboard feature is
broken".

## Three faults, and each one alone would have been survivable

**1. The client is told nothing.**

The Host sends `DASHBOARD_SNAPSHOT` and the agent metadata after `CONNECTED`. A denied
session never reaches `CONNECTED`, so both are simply absent — and absence renders
identically to "this agent has no dashboard configured". There is no `DENIED` frame, no
reason, no hint that onboarding exists.

**2. Only the "no" is logged.**

`network/trust/fast_rules.py`, `evaluate_request()`:

```python
for condition in allow_list:
    if condition == 'admin' and is_admin(client_id):
        return 'allow'          # ← no logger call
    if condition == 'whitelisted' and is_whitelisted(client_id):
        return 'allow'          # ← no logger call
    if condition == 'contact' and is_contact(client_id):
        return 'allow'          # ← no logger call
...
logger.warning(f"[FAST_RULES] No match, using default={default}")
logger.warning(f"[FAST_RULES] Returning 'deny' (default)")
```

Every deny is logged twice; no allow is logged at all. So "the log did not grow" is
ambiguous between *allowed* and *never asked* — the two states an operator most needs to
tell apart. During this debugging session that ambiguity produced a wrong conclusion in
both directions: first "still denied" when it had actually started allowing, then the
reverse.

**3. The one identifier you need is truncated.**

```python
logger.warning(f"[FAST_RULES] Evaluating client_id={client_id[:20]}... config={config}")
```

20 characters of a 64-character address. The operator's job at that moment is to put that
[...]

PR #623, opened 2026-08-03
Images bypass the upload size limit that files obey

wu-changxing:
`host.yaml` states the rule and where it applies:

```yaml
max_file_size: 10          # MB per file (both WebSocket and HTTP)
```

`validate_files()` enforces it, and both input paths call it — for `files`. The same handlers take `images` beside it:

```python
def handle_input(storage, prompt, session=None, connection=None, images=None, files=None):
    validate_files(files, config)
    return input_handler(..., images, files)
```

`images` is a list of base64 strings that goes straight to `agent.input()` and on to the model. Nothing measures it — not there, not in `input_handler` — and uvicorn sets no request-size cap of its own.

## How much this matters

Less than it first looks, and I want to be accurate about it: `/input` is authenticated (a raw POST gets 401), so this sits behind the trust gate in exactly the place the existing file check does. Nobody off the street can do it.

But "authenticated" includes contact level, which onboarding with an invite code grants. On the 1–2 GB VPS `co server new` provisions, one large image is enough to end the process, and the operator is left with an agent that died and no explanation.

## The change

`validate_images()` beside `validate_files()`, applying the limit that already exists and is already advertised in `/info` as `max_file_size_mb`. Called from both handlers.

Size is computed from the base64 string (4 characters per 3 bytes) rather than by decoding — decoding a 500 MB image to discover it is 500 MB would be the same problem one step later.

## Tests

`tests/unit/test_an_image_counts_against_the_upload_limit.py` — an oversized image is refused and the error names both the limit and the size; an image under it, no images, a raw base64 string with no `data:` prefix, and something that is not base64 at all all pass; too many images is refused by the count limit; and files still behave exactly as before. Red first (the function did not exist).

Both call sites are wired — checked by grep after writing them, because a validator nobody calls is the failure mode this release has produced four times.

### DIFF
--- a/connectonion/network/host/config.py
+++ b/connectonion/network/host/config.py
@@ -151,3 +151,48 @@ def validate_files(files: list, config: dict) -> None:
             )
 
     logger.info(f"File upload validated: {len(files)} file(s) accepted")
+
+
+def validate_images(images, config: dict) -> None:
+    """Hold images to the same limit as files.
+
+    `max_file_size` says "MB per file (both WebSocket and HTTP)" and
+    validate_files enforces it — for `files`. The same two handlers take
+    `images` beside it, a list of base64 strings that goes straight to
+    agent.input() and on to the model, and nothing measured it. uvicorn sets no
+    request-size cap of its own, so an authenticated client could hand the host
+    a body of any size and the agent would hold it in memory.
+
+    Not a new limit: the one already advertised in /info as max_file_size_mb.
+
+    The decoded size is what matters, and base64 is 4 characters per 3 bytes.
+    Measured from the string rather than decoded, so a large image is refused
+    without first allocating it — decoding to find out how big it is would be
+    the same problem one step later.
+    """
+    if not images:
+        return
+
+    max_count = config.get("max_files_per_request", DEFAULT_FILE_LIMITS["max_files_per_request"])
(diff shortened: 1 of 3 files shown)


wu-changxing:
Two things checked before this is merged, both about not breaking something that works.

## The refusal reaches the caller on both transports

A validator that raises into a socket nobody is listening on would be #434 again, so I traced where the `ValueError` goes:

- **HTTP** — `http_router.py:394` catches it and answers `400` with the message.
- **WebSocket** — `_agent_thread_body` catches it into `result_holder`, and `forward_agent_msgs_to_client` sends `{"type": "ERROR", "message": …}`.

So an oversized image is reported the same way an oversized file already is, on both paths. Nothing new can die silently.

## It does not break the chat UI

`connectonion-react` sends base64 images with no client-side bound, so the question is whether real traffic would now be refused. Its own store comment puts screenshots at the scale that "would blow the ~5MB localStorage quota", and the default limit here is 10MB **decoded** — roughly 13.3MB of base64. A browser-tool screenshot is well under that.

If someone does hit it, the message names the setting (`Increase max_file_size in host.yaml`) rather than failing anonymously.

## Scope, stated plainly

`/input` is authenticated — a raw POST gets 401 — so this sits behind the trust gate exactly where the existing file check does. It is not an open door; it is a limit that was advertised in `/info` as `max_file_size_mb` and applied to only half of what it named.

Issue #642, opened 2026-08-04
One address, several agents: whoever is nearest answers


wu-changxing:
Surfaced by #643, which made direct endpoint resolution work for the first time. The collision predates it.

`0x10e68f6dff39ab…` is announced by at least two different agents:

| where | name | tools |
|---|---|---|
| this laptop, `localhost:8000` | `oo` | `bash`, `write`, `edit`, `read_file`, `task`, … (24) |
| the deployed box | `naturewill` | the contract-ledger tools |

Both load the same keypair — `AGENT_CONFIG_PATH=/Users/changxing/.co`, which every project created by 1.5.x points at — so they are one identity as far as the protocol is concerned. The relay's directory for that address lists endpoints for all of them:

```
GET /api/agents/0x10e68f6d…
"endpoints": ["http://localhost:8000", "ws://localhost:8000/ws",
               "http://10.5.27.133:8000", …, "http://129.94.128.29:8000", …]
```

## What changed and what did not

Nothing about the identities changed. What changed is that a client now *reaches* one of them directly, and picks by network proximity — localhost, then LAN, then public. Before #643 the gate was on a field the relay does not send, so resolution always failed and every call went over the relay.

Measured after #643, asking for that address from this laptop:

```
resolve_endpoint(0x10e68f6d…)  ->  ws://localhost:8000/ws     (1.3s)
```

So a call meant for the deployed agent lands on the local coding agent instead. `/info` verification does not catch it, and cannot: the address genuinely matches — that is the whole point of sharing the keypair.

## Why it matters for a long-term release

The two agents do not have the same tools. Something scripted against the deployed agent's tools gets a coding agent with `bash` and `write` on the caller's own machine, and "which agent answered" depends on where the caller was sitting.

## Options

1. **Give each deployment its own identity.** `co create` already writes a project-local keypair; the 1.5.x `co init` projects that point at `~/.co` are the ones that share. A migration note plus making the local identity the default would end it.
2. **Let a client say which one it wants** — pass a preferred endpoint or a deployment tag to `connect()`.
3. **Have the relay expire endpoints per host**, so a laptop that has gone away stops being announced under an address the server still serves.

(1) is the real answer. Filing rather than choosing, because it changes what `co init` does to every new project.

wu-changxing:
Measured on current code, and it narrows this a lot: **new projects already get their own identity. The sharing is a legacy-project problem.**

A project created just now from the built wheel:

```
$ co create demoagent --yes
[...]

PR #688, opened 2026-08-05
Say so when two agents answer to one address

wu-changxing:
## What #642 measured

`0x10e68f6dff39ab…` was announced by two different agents at once:

| where | name | tools |
|---|---|---|
| this laptop, `localhost:8000` | `oo` | `bash`, `write`, `edit`, … (24) |
| the deployed box | `naturewill` | the contract-ledger tools |

Both load the same keypair, because every 1.5.x `co init` wrote `AGENT_CONFIG_PATH=~/.co` and a project with no key of its own inherits the global one **on purpose** — that fallback fixed a worse bug, where the host minted a third address and served under it while `co status` reported the configured one.

Since #643 a client resolves an endpoint directly and picks by proximity, so a call meant for the deployed agent lands on the local coding agent — one holding `bash` and `write` on the caller's machine. `/info` verification cannot catch this in principle: the address genuinely matches.

## What this does

The identity directory records which project and name serves under it. A different agent starting on the same keypair is told, by name and directory:

```
[host] 0x10e68f6d… is already served by 'oo' in /Users/…/oo, using the same
key from /Users/…/.co. Two agents on one address means whichever is nearest
answers the call, and the tools they answer with differ. A project created by
`co create` gets its own identity and does not share.
```

The record lives **beside the key**, not in the project — the whole problem is several projects sharing one identity directory. `address.load()` now reports which directory the key came from, so there is one answer to that rather than a rule restated at each caller.

## Why a warning and not a refusal

Aaron's rule is one agent, one address, and `co deploy` refusing is the right place for it. I did not add that here, and this is the half that is missing: **nothing today mints a local identity for a project that already exists.** Refusing a deploy would strand every 1.5.x project with no command to run. Filed separately rather than shipped as a dead end.

The host warning is also deliberately not fatal — the agent whose address this is may be the one restarting, and an operator who cannot start their agent because of a stale claim is worse off than one who is told the truth.

## Found while reviewing this change

- The first draft of the warning advised "point AGENT_CONFIG_PATH at this project's .co" — which does not work for the case that produces the warning, because the project has no key there. That is why it inherited one.
- The record would have **rsynced to the server**, putting a local absolute path on the box and making the deployed agent read the laptop's claim as a collision on every start. Excluded, and verified through real rsync rather than by reading the exclude list — the way #686 found `.env.example` silently not travelling.
- The first draft wrote the record without asking whether the directory existed, which killed nine host tests. A hint about who is serving is not worth failing a start over.

## Tests

12 cases, red first. Full suite 4571 passed.

Part of #642.

### DIFF
--- a/connectonion/address.py
+++ b/connectonion/address.py
@@ -311,6 +311,11 @@ def load(co_dir: Path) -> Optional[Dict[str, Any]]:
             "email": email,
             "email_active": email_active,
             "legacy_derivation": legacy_derivation,
+            # Which directory this key came from. A project with no key of its
+            # own falls back to the global ~/.co, so anything recorded *about*
+            # the key -- which agent is serving under it (#642) -- has to be
+            # written beside the key, not in whichever project loaded it.
+            "source": str(co_dir),
             "signing_key": signing_key
         })
         
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -611,6 +611,11 @@ RSYNC_FILTERS = [
     "--exclude", ".co/keys/",
     "--exclude", ".co/address.json",
     "--exclude", ".co/admins.txt",
+    # Which agent is serving under this key, and from which directory on
+    # the operator's machine (#642). Shipping it puts a local absolute path
+    # on the server and makes the deployed agent read the laptop's claim as
+    # a collision on every start.
+    "--exclude", ".co/served_by.json",
     "--exclude", ".co/provision.json",
     "--exclude", ".co/requirements.sha256",
     "--exclude", ".co/logs/",
--- a/connectonion/network/host/server.py
+++ b/connectonion/network/host/server.py
@@ -30,6 +30,7 @@ import random
 from functools import partial
 import os
 from pathlib import Path
+import json
 from typing import Callable, Optional, Union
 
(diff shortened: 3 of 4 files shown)


wu-changxing:
Ran a real host on this branch rather than only the suite, and it reproduced #642 live in the process.

A throwaway project — a fresh directory, an empty `.co/`, its own `host.yaml` naming the agent `smoke` — started and served under:

```
0x10e68f6dff39ab1c50cc48ea1c74e7fd6ce7269aa6e8123829b344e57d005508
relay: oo.openonion.ai
```

That is the address from the issue. A directory created a minute earlier, with no key of its own, announced itself to the relay as the same agent as everything else on this machine. Nothing warned, because it was the first claim — it wrote the record and became the incumbent:

```json
{
  "name": "smoke",
  "project": "/private/tmp/.../scratchpad/smoke/proj",
  "address": "0x10e68f6dff39ab…"
}
```

Which is the behaviour this PR is for: the *next* agent on that key is told, by name and directory. I removed the record afterwards rather than leave the real `~/.co` holding a claim from a scratch directory.

Two things it confirms beyond the tests: the warning path runs at real host startup with a real agent and a real relay announce, and the collision is not hypothetical or specific to the two agents in the issue — any project without its own key joins it.

Issue #690, opened 2026-08-05
The buyer names the price: onboarding payment verifies the amount the client sent, not the one the agent charges


wu-changxing:
Found while removing the shipped default price (#672). Same code path, opposite direction, and this one is exploitable.

## What happens

`handle_onboard_submit` takes the amount out of the client's own frame:

```python
payload = data.get("payload", {})
payment = payload.get("payment", 0)
...
if payment > 0:
    if trust_agent.verify_payment(agent_address, payment):
```

and `verify_payment` prefers it over the configured one:

```python
required = self._config.get('onboard', {}).get('payment')
if not required:
    return False

# Use configured amount if not specified
min_amount = amount if amount > 0 else required
```

The comment says *"if not specified"* — but the only caller reaches this line when `payment > 0`, so `amount` is **always** specified and `required` is never the minimum. The operator's price survives only as a truthiness check for whether payment onboarding is switched on at all.

## What it costs

An agent advertising `payment: 10` is opened by a transfer of **0.01**, from a stranger who sends `{"payment": 0.01}` and makes that transfer. oo-api confirms it honestly — it was asked to confirm 0.01 — and `promote_to_contact` runs. On `careful`, contacts are allowed.

Until #672 every agent created by `co init` advertised `payment: 10` without its operator configuring anything, so this applied to all of them.

## Shape

The same one as #561 and #571: a value that is **advertised** and a value that is **enforced**, resolved in two places, and the enforcing one reads something the operator never wrote. `doors_that_open` exists because of exactly this; the submit side was not routed through it.

## Fix

In the PR alongside #672: `verify_payment` resolves the operator's configured price and requires that. What the client sends is at most a claim about what they did, and a claim is not an authorisation.

Both prices now come from one resolver, so what `/info` advertises is what the transfer is checked against — there is a test that asserts those two are the same number.

PR #693, opened 2026-08-05
The trust gate stops crashing on every stranger, and stops selling on credit

wu-changxing:
**Regression from #691, which I merged an hour ago. This is the fix.**

## The crash

#691 made the shipped price opt-in — `payment: $CO_PAYMENT`. **Three** places read `onboard.payment`. That change routed two through a resolver (`doors_that_open`, which advertises; `verify_payment`, which enforces) and missed the third:

```python
required_payment = onboard.get('payment')
request_payment = request.get('payment', 0)
if required_payment and request_payment >= required_payment:
```

`"$CO_PAYMENT"` is a non-empty string, so the guard passes and the comparison runs `0 >= "$CO_PAYMENT"`. Measured against a real host on default trust, from a client that had never connected:

```
File ".../fast_rules.py", line 226, in evaluate_request
    if required_payment and request_payment >= required_payment:
TypeError: '>=' not supported between instances of 'int' and 'str'

client: websockets.exceptions.ConnectionClosedError
```

**Every stranger CONNECT died there** — which is every first contact any agent ever has. The suite stayed green because nothing exercises the CONNECT path with the shipped policy. I found it by starting a real host to verify something else.

## The older half, which is worse

`request_payment` is a number **the client wrote in its own request**, and nothing on this path verifies a transfer — `verify_payment` and its oo-api call are on ONBOARD_SUBMIT. A stranger who put a large enough number in a CONNECT was promoted to contact having sent nothing.

#690 was the buyer naming the price. This was the buyer skipping the payment.

That predates #691; the string just made it crash before it could do it.

## The fix

Payment is not decided here at all. A stranger offering one falls through to `default` (`ask` on careful) and pays through ONBOARD_SUBMIT, where the transfer is verified. One door, one check — which is what `doors_that_open` was built for and this path had quietly stayed outside of.

The invite-code path is untouched: it checks a secret the operator set, not a number the client chose.

## Also found

The trust lists live wherever `_project_co_dir()` points, which during a test run is shared by the whole session — measured at `/tmp/.co/contacts.txt`. `promote_to_contact` is durable, so **one test that promotes an address changes the answer for every later test using it**. The first red run of this file left its own stranger a contact, and the green run then passed for the wrong reason. These tests pin the directory per test; the shared-state problem itself is worth its own issue.

## Tests

9 cases, red first (the four CONNECT ones raised TypeError, the payment ones returned `allow`).

One of my own expectations was wrong and is corrected here rather than worked around: `default: ask` returns `None` ("fast rules did not decide, hand it to the LLM"), not the string `"ask"`.

### DIFF
--- a/connectonion/network/trust/fast_rules.py
+++ b/connectonion/network/trust/fast_rules.py
@@ -220,16 +220,27 @@ def evaluate_request(config: dict, client_id: str, request: dict) -> Optional[st
         )
         return 'allow'
 
-    # Check payment
-    required_payment = onboard.get('payment')
-    request_payment = request.get('payment', 0)
-    if required_payment and request_payment >= required_payment:
-        promote_to_contact(client_id)
-        logger.warning(
-            f"[FAST_RULES] Returning 'allow' — {client_id} onboarded by payment, "
-            f"promoted to contact"
-        )
-        return 'allow'
+    # Payment is NOT decided here. This used to read
+    #
+    #     required_payment = onboard.get('payment')
+    #     request_payment = request.get('payment', 0)
+    #     if required_payment and request_payment >= required_payment:
+    #         promote_to_contact(client_id)
+    #
+    # which promoted a stranger on a number they wrote in their own request. No
+    # transfer is checked on this path -- verify_payment and its oo-api call are
+    # on ONBOARD_SUBMIT -- so saying "payment: 10" was enough to become a
+    # contact, having sent nothing. #690 was the buyer naming the price; this
+    # was the buyer skipping the payment.
+    #
+    # It also raised. Once the shipped policy carried `$CO_PAYMENT` instead of a
+    # literal (#691), `required_payment` was a non-empty string, the guard
+    # passed, and `0 >= "$CO_PAYMENT"` killed the trust gate on every stranger
+    # CONNECT -- the first contact any agent ever has.
+    #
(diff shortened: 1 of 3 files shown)


wu-changxing:
CI caught two tests asserting the behaviour this removes — `test_payment_onboards` and `test_exact_payment_works`. They encode the bypass rather than detect a regression, so before changing them I demonstrated it end to end rather than arguing from the code.

A **freshly generated identity** — no history, no balance, never seen by this agent — sending a signed CONNECT whose payload merely *claims* a payment:

```
claimed payment: 999, transferred: nothing

before   -> {"type": "CONNECTED", "session_id": "9a4d3226-…", "status": "new"}
            contacts.txt: 1 entry

after    -> {"type": "ONBOARD_REQUIRED", "methods": ["payment"], "payment_amount": 10.0}
            contacts.txt: absent
```

Full access as a contact, having transferred nothing. The signature proves who said it, not that anything was paid — `auth.py` reads it straight off the client's frame:

```python
request_data = {
    "prompt": prompt,
    "invite_code": payload.get("invite_code"),
    "payment": payload.get("payment", 0),
}
```

The invite-code sibling checks the claim against a secret the operator set. The payment one checked it against the claim itself.

Both tests now assert the refusal, with that measurement in the docstring.

PR #698, opened 2026-08-05
A session belongs to whoever started it

### LINKED ISSUE #696: Naming someone else's session id hands you their conversation
Measured, not inferred. Two **different** identities, both whitelisted on the same agent, against a real host over a real WebSocket:

```
A CONNECTED session=457204c5… status=new
A -> "Remember this exactly: my codeword is PURPLE-OTTER-42. Reply OK."
A <- "OK."

B CONNECTED session=457204c5… status=connected      <- B named A's session id
B -> "What codeword did I give you earlier? Answer with just the codeword."
B <- "PURPLE-OTTER-42."
```

B is a freshly generated identity that had never spoken to this agent. `status=connected` means it was attached to A's live session, not given a new one.

## Why

The session id comes off the client's frame, and nothing checks who it belongs to:

```python
session_id = data.get("session_id") or str(uuid.uuid4())
conn["session_id"] = session_id
...
stored = storage.get(session_id)
if stored and stored.session:
    conn["session"], server_newer = merge_sessions(client_session=client,
                                                    server_session=stored.session)
```

There is nothing to check it against. `Session` records no owner:

```python
class Session(BaseModel):
    session_id: str
    status: str
    prompt: str
    result: Optional[str] = None
    session: Optional[dict] = None
    created / expires / duration_ms
```

So the id **is** the credential, and it is not treated as one — it is returned in the CONNECTED frame, printed in the host console, written to `.co/session_results.jsonl`, and handed out wholesale by `GET /sessions` (#683). Read the ids there, then connect with one.

## Scope

Any caller who passes the trust gate. On `careful` that includes contacts, so on an agent with an invite code configured this is reachable by anyone who was let in at all — and #653's chain shows how cheap that is.

A running session is worse than a stored one: `resume_forwarding` rewinds and streams the *live* output of someone else's turn to the new connection.

## Fix

Record the owner when a session is created and treat a session owned by someone else as not found — a fresh session with that id, indistinguishable from an expired one. Refusing would confirm the session exists; starting fresh discloses nothing and breaks nothing legitimate, since a session is one conversation with one caller.

`requester` already travels on the INPUT path (`{'address', 'level'}`) and is already in `SERVER_OWNED_SESSION_KEYS` — the client carries it but may not author it. That is the value to persist.

## Related

- #683 — `GET /sessions` returns every conversation to anyone who can reach the port. That is where the ids come from.
- #649 — a captured CONNECT is a five-minute bearer token. Same theme: what may act on a session, and on what proof.
- #194 — owner-bound ULW state across reconnects, which assumes an owner that is not recorded here.

Blocks 1.6.0 in my view: it is a cross-identity leak of conversation content on the default trust level.

wu-changxing:
## Before

Two different identities, both whitelisted on one agent, real host, real WebSocket:

```
A CONNECTED session=457204c5… status=new
A -> "Remember this exactly: my codeword is PURPLE-OTTER-42. Reply OK."
A <- "OK."

B CONNECTED session=457204c5… status=connected     <- B named A's session id
B -> "What codeword did I give you earlier? Answer with just the codeword."
B <- "PURPLE-OTTER-42."
```

## After

```
B CONNECTED session=8d240fd0… status=new
B <- "I don't have a codeword."
```

## The owner was already being recorded

`input_handler` writes `session['requester']` from the signature-verified address on every turn, and `requester` is in `SERVER_OWNED_SESSION_KEYS` — the client carries it but does not get to author it, for exactly this class of reason. Nothing read it back.

## Three places, found one at a time

This is the failure shape that has come up all through this release: one decision made in several places, and the fix landing on some of them.

1. `establish_connection` merged the stored session
2. the active registry attached the **live** turn — `resume_forwarding` rewinds and streams output of a turn in progress
3. `input_handler` looked the session up **again** on the INPUT that followed

Fixing 1 and 2 was not enough, and I only know that because I re-ran the measurement instead of trusting the diff:

```
B CONNECTED session=47edf32c… status=new                      <- 1 and 2 worked
host: ↩ session 47edf32c… belongs to another caller — starting a new one
B <- "PURPLE-OTTER-42."                                       <- and 3 leaked
```

## Behaviour

Owned by someone else is treated as **not found** — a fresh session with that id, indistinguishable from one that expired. Refusing would confirm the session exists. Nothing legitimate breaks: a session is one conversation with one caller.

Sessions stored before this have no owner and belong to nobody, so they are readable by whoever names them — same as today. That is deliberate: silently orphaning existing sessions on upgrade would be a worse surprise than the status quo for data that predates the check.

## Tests

12 cases, red first. They cover each of the three paths separately, so a future change that reopens one is caught on its own.

Two of my own test stubs were wrong and are fixed rather than worked around: the stored-session fixture had no `iteration`/`updated`, so `merge_sessions` let the empty client session win and the fixture proved nothing either way; and the owner-attaches case used a bare `object()` for `io`, which cannot be rewound — it fails at `rewind_to`, which is itself evidence the owner reaches the resume path.

Closes #696.

### DIFF
--- a/connectonion/network/host/http_router.py
+++ b/connectonion/network/host/http_router.py
@@ -56,6 +56,23 @@ def input_handler(create_agent: Callable, storage: SessionStorage, prompt: str, 
     server_newer = False
 
     stored = storage.get(session_id)
+    # Same rule CONNECT applies, because this is the same decision: a session
+    # belongs to whoever started it, and the id is on the client's frame.
+    #
+    # Fixing it only in establish_connection was not enough -- measured. B
+    # connected, was correctly given a fresh session ("belongs to another
+    # caller"), and then its INPUT named the same id and this lookup handed
+    # over A's conversation anyway:
+    #
+    #     B CONNECTED session=47edf32c… status=new     <- the CONNECT fix worked
+    #     B <- "PURPLE-OTTER-42."                      <- and it leaked here
+    #
+    # `requester` is the signature-verified address for this turn.
+    from .session import session_owner
+
+    owner = session_owner(stored)
+    if owner and requester and owner != requester.get('address'):
+        stored = None
     if stored and stored.session:
         session, server_newer = merge_sessions(
             client_session=session,
--- a/connectonion/network/host/session/__init__.py
+++ b/connectonion/network/host/session/__init__.py
@@ -7,7 +7,7 @@ LLM-Note:
   Integration: exposes Session, SessionStorage, ActiveSession, ActiveSessionRegistry, start_cleanup_job, merge_sessions, session_to_chat_items
 """
 
-from .storage import Session, SessionStorage
+from .storage import Session, SessionStorage, session_owner
 from .active import ActiveSession, ActiveSessionRegistry, start_cleanup_job
 from .merge import merge_sessions
 from .ui import session_to_chat_items
@@ -16,6 +16,7 @@ __all__ = [
     # Storage
     'Session',
     'SessionStorage',
+    'session_owner',
     # Active sessions
     'ActiveSession',
     'ActiveSessionRegistry',
--- a/connectonion/network/host/session/active.py
+++ b/connectonion/network/host/session/active.py
@@ -33,6 +33,7 @@ class ActiveSession:
     created: float
     last_ping: float
     status: str                  # 'running' | 'connected'
+    owner: object = None         # address that started it; None = nobody (#696)
 
 
 class ActiveSessionRegistry:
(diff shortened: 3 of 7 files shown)


wu-changxing:
Step 7 on my own change: the fix introduced the next bug, and re-running the measurement is the only reason I know.

Treating a session owned by someone else as "not found" left the caller holding the **same id**. Their turn then saved a record under it — storage is append-only, last entry wins — and the owner came back to an empty conversation:

```
A turn 1: session=1b7625e1…  -> OK.
B squats on it (status=new)  -> Hello!
A returns (status=connected) -> "I don't have a codeword from you."
A LOST their conversation
```

That trades a disclosure for a destruction, which is not a trade — and it needs exactly the same thing the leak needed, the id.

A caller who names somebody else's session now gets a **fresh id**, reported in CONNECTED so the client knows which session it is actually in. The owner's record is never written to:

```
A turn 1: session=080d48a3…  -> OK.
B squats on it (status=new)  -> Hello!
A returns (status=connected) -> PURPLE-OTTER-42
A KEPT their conversation
```

The live-session check moved up alongside it, so one place decides this rather than two — which is the failure shape this whole release has been about.

5 more cases, 3 red first. 17 in the file.

Issue #670, opened 2026-08-04
The model billing key is also the password to every agent's logs and sessions


wu-changxing:
## What

`/admin/logs` and `/admin/sessions` authenticate with a bearer token, and the token is `OPENONION_API_KEY`:

```python
if path in ["/admin/logs", "/admin/sessions"]:
    auth = headers_dict.get(b"authorization", b"").decode()
    expected = os.environ.get("OPENONION_API_KEY", "")
    if not expected or not auth.startswith("Bearer ") or not hmac.compare_digest(auth[7:], expected):
        await send_json({"error": "unauthorized"}, 401)
```

That is the same key that pays for `co/*` models. It is what `co auth` obtains, what `co create` writes into the project's `.env`, and what `transcribe`, `google_calendar`, `microsoft_calendar` and `send_email` read. It is transmitted to `oo.openonion.ai` on every model call.

Measured against a real hosted agent:

```
$ curl -H "Authorization: Bearer $OPENONION_API_KEY" http://host:8879/admin/sessions
{"sessions": [{"session_id": "real1", "status": "done", "prompt": "hi"}]}

$ curl -H "Authorization: Bearer $OPENONION_API_KEY" http://host:8879/admin/logs
A REAL LOG LINE
============================================================
Session started: 2026-08-04 20:45:48
```

The sessions response carries the prompts. The log carries the turn history.

## Why this is worth a decision rather than a patch

Three properties of that key make it a poor authorization token, and none of them are accidents:

1. **It is a billing credential, so it travels.** It is in `.env` files, in `~/.co/keys.env`, in deploy environments and in CI secrets. Every place it is legitimately copied to for *paying for tokens* becomes a place that can read the agent's conversations.
2. **It is shared across agents.** One account key, many projects. There is no per-agent admin secret, so a key that opens one agent's logs opens all of them.
3. **The agent must hold it anyway.** It cannot be withheld from a `co/*` agent — the agent needs it to call the model — so "do not put the admin secret on the box" is not available as advice.

What it gets right: it fails closed when unset (`not expected` refuses), and it compares with `hmac.compare_digest`.

## Not fixed here because the fix is a choice

[...]

PR #699, opened 2026-08-05
Reading sessions needs an identity

### LINKED ISSUE #683: GET /sessions returns every conversation to anyone who can reach the port
`GET /sessions` and `GET /sessions/<id>` require **no credentials** and return the full contents of every session the agent has served.

## Measured on `main` (1905597)

A hosted agent, one turn served, then probed with no headers at all:

```
$ curl http://127.0.0.1:9002/sessions
{"sessions": [{"session_id": "fe618103-…", "status": "done",
  "prompt": "SECRET-PROMPT-9317 please just say ok",
  "result": "ok",
  "session": {"session_id": "…",
    "requester": {"address": "0xd716ad3c…", "level": "stranger"},
    "messages": [{"role": "user", "content": "SECRET-PROMPT-9317 …"}, …]}}]}

$ curl http://127.0.0.1:9002/sessions/fe618103-…
200, same content
```

So: the prompt, the result, the whole message array, and **who asked** — their address and trust level.

## The comparison that makes it plain

`/admin/sessions` is gated behind a bearer token (#670) and returns **less**:

```
$ curl -H "Authorization: Bearer wrong" .../admin/sessions
{"error": "unauthorized"}
```

The unauthenticated route returns more than the authenticated one.

## The repo already knows how to think about this

`info_handler` deliberately withholds the account balance, and says why:

> The balance is deliberately not here. `/info` needs no credentials, so on a deployed agent this response is readable by the whole internet, and the operator's account balance is both commercially revealing on its own and a targeting signal.

That reasoning applies to conversations at least as strongly, and was not applied here.

## Writes are protected; reads are not

`POST /input` refuses correctly:

```
$ curl -X POST -d '{"prompt":"hello"}' .../input
{"error": "unauthorized: signed request required"}
```

So an unauthenticated party cannot make the agent do anything — they can only read everything it has done.

## Not fixing it unasked

The options are not equivalent and one of them may break a client already in use:

- require a signature, as `/input` does
- require admin, as `/admin/trust/*` does
- return only sessions whose `requester.address` matches the caller
- keep it open but strip `prompt`, `result` and `messages`, leaving ids and status

`chat.openonion.ai` reads sessions somewhere, and I do not know whether it uses this route or the WebSocket. That decides which of the four is even available.

Found on the first pass over the HTTP surface in this release loop — `/info` and the admin routes had been exercised, these had not.

wu-changxing:
`curl http://agent/sessions` returned **every conversation on the agent** to anyone who could reach the port. Measured on a real host — no signature, no identity, no key:

```
GET /sessions -> {"sessions": [{"session_id": "9fd5d3fc-…",
                                "status": "done",
                                "prompt": "What codeword did I give you earlier?",
                                "result": "PURPLE-OTTER-42",
                                "session": {…every message…}}]}
```

`POST /input` immediately above it in the same dispatch went through `route_handlers["auth"]`. These two never did. It is also where #696's attack got its session ids — that one needed an id to read somebody's conversation; this handed out every id on the agent.

## Approach

Aaron picked signed GETs over gating on `OPENONION_API_KEY` (#670 is about that key being over-powered) and over deleting the endpoints.

A GET has no body, so the signature is in headers over `{method, path, timestamp}`:

```
x-co-from:      the caller's agent address
x-co-signature: sign(canonical(payload))
x-co-timestamp: unix seconds
```

Canonicalised **exactly** as CONNECT and INPUT already are, and verified by the same `_authenticate_signed` — same five-minute freshness window, same blacklist check. No second implementation of any of it. Most of what this release has been fixing is one decision made in two places and drifting; this deliberately does not add another.

The payload is rebuilt from the server's **own** method and path, never from anything the client sent, so a signature made for `/sessions/mine` does not verify against `/sessions/yours`. Tested both ways.

`sign_request()` is exported so a client does not reconstruct the format by hand — and so these tests drive the real one instead of certifying their own idea of it.

## Scope

From the owner #698 records: `/sessions` returns the caller's own, `/sessions/{id}` returns one only if they own it. Sessions stored before an owner existed stay readable — same rule as #698, since silently orphaning existing history on upgrade is a worse surprise than the status quo.

## Measured after

```
unauthenticated curl      -> {"error": "unauthorized: 'from' field required"}
signed as A               -> 1 session; result: PURPLE-OTTER-42.
signed as B (new to it)   -> 0 sessions
```

## Also

The agent's own `/docs` page advertised `curl -X GET "http://localhost:8000/sessions"` as the way to call this. Left alone it becomes a documented call that always 401s — the shape this release keeps running into. Updated to show `co call` and the header form.

## Tests

12 cases, red first.

Closes #683.

### DIFF
--- a/connectonion/network/host/auth.py
+++ b/connectonion/network/host/auth.py
@@ -149,6 +149,63 @@ def extract_and_authenticate(data: dict, trust, *, blacklist=None, whitelist=Non
         return None, agent_address, True, f"forbidden: {decision.reason}"
 
 
+# ─────────────────────────── signed GETs ───────────────────────────
+#
+# A GET has no body to sign, so `curl http://agent/sessions` returned every
+# conversation on the agent -- prompts, answers and full message history -- to
+# anyone who could reach the port (#683). It is also where #696's attack got
+# its session ids.
+#
+# The signature travels in headers over {method, path, timestamp}, canonicalised
+# exactly as CONNECT and INPUT already are, and is then verified by the same
+# _authenticate_signed below. One canonicalisation, one freshness window, one
+# blacklist check -- a second implementation of any of those is how the halves
+# of a gate drift apart, which is most of what this release has been about.
+
+FROM_HEADER = "x-co-from"
+SIGNATURE_HEADER = "x-co-signature"
+TIMESTAMP_HEADER = "x-co-timestamp"
+
+
+def _canonical(payload: dict) -> bytes:
+    """The bytes that get signed. Same shape as connect.py's."""
+    return json.dumps(payload, sort_keys=True, separators=(',', ':')).encode()
+
+
+def sign_request(keys: dict, method: str, path: str, timestamp=None) -> dict:
+    """Headers a client sends with a signed GET.
+
+    Exported so a client does not have to reconstruct the format -- and so the
+    tests drive the real one rather than certifying their own idea of it.
+    """
+    from ... import address as _address
+
+    payload = {"method": method.upper(), "path": path,
+               "timestamp": int(timestamp if timestamp is not None else time.time())}
+    return {
+        FROM_HEADER: keys["address"],
+        SIGNATURE_HEADER: _address.sign(keys, _canonical(payload)).hex(),
+        TIMESTAMP_HEADER: str(payload["timestamp"]),
+    }
+
+
+def request_from_headers(headers: dict, method: str, path: str) -> dict:
+    """Turn a signed GET into the dict every other authenticated frame is.
+
+    The payload is rebuilt from the *actual* method and path, not from anything
+    the client sent, so a signature made for one path does not verify against
+    another.
+    """
+    lowered = {str(k).lower(): v for k, v in (headers or {}).items()}
+    timestamp = lowered.get(TIMESTAMP_HEADER)
+    return {
+        "payload": {"method": method.upper(), "path": path,
+                    "timestamp": int(timestamp) if timestamp is not None else None},
+        "from": lowered.get(FROM_HEADER),
+        "signature": lowered.get(SIGNATURE_HEADER),
+    }
+
+
 def _authenticate_signed(data: dict, *, blacklist=None, recipient_address=None):
     """Authenticate signed request with Ed25519 - ALWAYS REQUIRED.
 
--- a/connectonion/network/host/http_router.py
+++ b/connectonion/network/host/http_router.py
@@ -192,15 +192,34 @@ def exec_handler(create_agent: Callable, permissions: dict, tool_name: str, args
             "duration_ms": int((time.time() - start) * 1000)}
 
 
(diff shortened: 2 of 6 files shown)


wu-changxing:
Step 7 on my own change: the docs page needed three fixes, all found by looking at the page rather than the diff.

1. The static `<pre>` I first edited is **overwritten by `updateCurlCommands()`** on load. The edit was dead on arrival; the real place is the JS that builds the curl.
2. The **Execute button and auto-refresh** do a bare `fetch('/sessions')` — they would have rendered `{"error": "unauthorized: 'from' field required"}` into the page on every click and every 5 seconds. The page has no key and must not have one, so it now shows the call rather than making it.
3. My replacement python snippet used `Path` **without importing it** — a NameError for anyone who copied it, which is the same defect as the curl it was replacing. The version in the PR runs; I ran it:

```
snippet runs; headers: ['x-co-from', 'x-co-signature', 'x-co-timestamp']
```

Also caught while writing the first version: I documented `co call <address> sessions`, which **does not exist** — `co call` invokes a remote tool over EXEC, not an HTTP endpoint. Removed rather than shipped.

That is four invented-or-broken instructions in one small docs edit, in a PR whose entire subject is that the docs advertised a call that no longer works.

wu-changxing:
CI caught six `tests/e2e/test_host.py` cases calling `GET /sessions` and `GET /sessions/{id}` unauthenticated — the contract this branch changes.

They test the response shape and the 404 behaviour, which are still worth testing, so they sign now rather than being removed.

One thing that needed fixing along the way: the test client generated a **fresh key per request**, so a session created by `post_signed` was owned by one address and read by another. With #698's ownership in place every read would have returned nothing — correctly, and uselessly. The client now holds one identity for its lifetime, and `create_signed_request` takes the key so both halves are the same caller.

That is the same class of thing as the fakes in #692: a test helper whose shape stopped matching the real flow, and which would have gone on passing for the wrong reason.

50 passed in that file.

wu-changxing:
Correcting myself, since I said something wrong a moment ago.

I inserted `signed_scope` **between `@pytest.mark.asyncio` and `class TestHandleHttpRouting`**, so the marker attached to my helper and the class lost it. Every async test in that class stopped being a test.

Locally that appeared as **13 SKIPPED** — "async def function and no async plugin installed" — which I read as an environment quirk, and briefly concluded these tests simply could not run on my machine. That was wrong. They run fine; my edit had removed their marker. With the decorator back where it belongs:

```
tests/unit/test_asgi_http.py    32 passed
```

no `asyncio_mode` forcing needed.

Worth noting which signal was which. CI reported it as **failures**, including `test_health_endpoint` and `test_info_endpoint` — tests this branch does not touch, which is what made it obvious something structural had broken. Locally the same damage was a **skip**, and a skip is easy to scroll past. If I had only run locally I would have shipped a file whose async tests silently did nothing.

Issue #649, opened 2026-08-04
A captured CONNECT is a five-minute bearer token for the whole whitelisted tool surface


wu-changxing:
Found by probing replay on the trust surface, and it is worse than replay.

## What the protocol authenticates

`CONNECT` carries a signed payload. **`EXEC` does not carry a signature at all:**

```python
exec_msg = {"type": "EXEC", "exec_id": exec_id, "tool": tool, "args": args}
```

So the signature authenticates the *connection*, and every command on it is trusted because of who opened it. `SIGNATURE_EXPIRY_SECONDS = 300`, and nothing marks a signature as used.

## Measured against a live agent

A signed `CONNECT` was captured once, then replayed to open new sessions. The `EXEC` frames were **written by the attacker, not captured** — a tool with a side effect, called three times:

```
1st (legitimate)                 -> ran 1 times
2nd (CONNECT replayed verbatim)  -> ran 2 times
3rd (replayed again)             -> ran 3 times
side_effects.txt: 3 lines
```

The same holds for admin commands — a captured `ADMIN_PROMOTE` replays cleanly three times.

So within five minutes, one captured CONNECT is not "repeat what the admin did". It is **any whitelisted tool, with any arguments**.

## Why this became reachable

Before #643, `resolve_endpoint` never resolved anything, so every client went through the relay — `wss://`, TLS. #643 fixed that gate, and direct connections became the normal path. Direct endpoints announced by a self-hosted agent are plain **`ws://`**:

```
"endpoints": ["http://10.5.27.133:8797", "ws://10.5.27.133:8797/ws", …]
```

On a LAN, anyone who can observe that traffic can capture a CONNECT. That is a consequence of my own change earlier in this loop; the design predates it, but TLS was covering it.

Deployed agents are served over https (1.5.3), so they offer `wss://` — the exposure is agents run locally and reached across a network, which is exactly what direct resolution is for.

## Options

1. **Only speak plaintext to loopback.** Prefer a `wss://` endpoint when one is offered; use plain `ws://` only for localhost, where there is no network to observe; otherwise fall back to the relay. Keeps the speed win of #643 for the two safe cases and puts nothing signed on a LAN wire. Smallest change, no protocol edit.
2. **Single-use CONNECT.** Cache recently seen signatures for the freshness window and refuse a repeat. Closes replay specifically: the attack needs to *open* a connection, and an attacker cannot inject into someone else's socket without a MITM position.
[...]

PR #703, opened 2026-08-05
A captured CONNECT opens nothing twice, and nothing signed goes onto a LAN

wu-changxing:
`EXEC` carries **no signature of its own**. The signature authenticates the *connection*, and every command on it is trusted because of who opened it. So a captured CONNECT replayed inside the five-minute window is not "repeat what the caller did" — it is any whitelisted tool with any arguments, because the attacker writes the EXEC frames. From #649's measurement against a live agent:

```
1st (legitimate)                 -> ran 1 times
2nd (CONNECT replayed verbatim)  -> ran 2 times
3rd (replayed again)             -> ran 3 times
side_effects.txt: 3 lines
```

Two halves here, neither a protocol change.

## One signature opens one connection

Seen once, refused after. The attack has to **open** a connection; without a MITM position an attacker cannot inject into somebody else's live socket.

Legitimate clients are unaffected — each builds a fresh CONNECT with a fresh timestamp, and the one place this codebase re-establishes a connection already avoids replaying the frame (`ws_router/connect.py`: *"no CONNECT replay — its signature may have aged past the 5-minute window"*).

## Plaintext only to loopback

#643 made direct resolution work, and a self-hosted agent announces plain `ws://`. On a LAN anyone who can observe that traffic can capture a CONNECT. Now: `wss://` preferred when offered, plain `ws://` only where there is no network to observe, everything else back through the relay.

**⚠️ Behaviour change worth knowing about:** a LAN agent reached over plain `http`/`ws` is no longer contacted directly — that traffic goes through the relay. Deployed agents are served over https (1.5.3) and localhost is untouched, so both cases direct resolution was built for keep it. This is option 1 as written in #649, but it does trade some of #643's speed win for LAN users, so it should be a conscious call rather than something noticed later.

## Not done

Signing each command — #649's option 3, every EXEC and INPUT carrying its own signature over its own contents. That is the complete answer and a protocol change. These two close the route without breaking a client.

## Two test-side consequences, both real

The replay cache is **process-global**, and has to be, since the replay it refuses crosses connections. That makes it state one test hands to the next: several files send a hardcoded `"signature": "0xsig"`, so the second test to use one was refused for reasons unrelated to itself. `conftest` clears it per test.

#643's resolution tests describe their LAN and public hosts over TLS now. They are about ordering and resolution mechanics, not schemes; left as plain `http` they would have been testing the new refusal instead of what they were written for.

## Tests

15 cases, 13 red first.

Two of my own were wrong and are corrected rather than worked around: the cache-pruning test aged frames instead of the recorded entries (the cache stamps *insertion* time, deliberately, so a far-future timestamp cannot pin an entry), and `connectonion.network.connect` resolves to a re-exported *function*, not the module, so a patch on it landed nowhere — the same shadowing that bit #687.

Part of #649.

### DIFF
--- a/connectonion/network/connect.py
+++ b/connectonion/network/connect.py
@@ -96,6 +96,35 @@ def _sort_endpoints(endpoints: List[str]) -> List[str]:
     return sorted(endpoints, key=priority)
 
 
+LOOPBACK = ("localhost", "127.0.0.1", "::1", "[::1]")
+
+
+def endpoint_is_safe(url: str) -> bool:
+    """May a signed frame go to this endpoint?
+
+    TLS anywhere, or plaintext to loopback only.
+
+    Before #643 `resolve_endpoint` never resolved anything, so every client went
+    through the relay over `wss://` and TLS covered this. Direct connections
+    then became the normal path, and a self-hosted agent announces plain
+    `ws://`:
+
+        "endpoints": ["http://10.5.27.133:8797", "ws://10.5.27.133:8797/ws", …]
+
+    A CONNECT is signed, and on a LAN anyone who can observe that traffic can
+    capture one. Within the freshness window a captured CONNECT is the whole
+    whitelisted tool surface (#649). Loopback has no network to observe, and a
+    deployed agent is served over https (1.5.3), so both of the cases direct
+    resolution exists for stay fast; the rest falls back to the relay.
+    """
+    from urllib.parse import urlparse
+
+    parsed = urlparse(url)
+    if parsed.scheme in ("https", "wss"):
+        return True
+    return (parsed.hostname or "") in LOOPBACK
+
+
 async def resolve_endpoint(
     agent_address: str,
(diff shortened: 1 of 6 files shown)


wu-changxing:
Verified against a real host, capturing one signed CONNECT and replaying the **same bytes** on fresh sockets:

```
=== before
  1st (legitimate)                   -> CONNECTED new
  2nd (replayed verbatim)            -> CONNECTED new
  3rd (replayed again)               -> CONNECTED new
  REPLAY STILL WORKS

=== after
  1st (legitimate)                   -> CONNECTED new
  2nd (replayed verbatim)            -> ERROR unauthorized: this CONNECT was already used
  3rd (replayed again)               -> ERROR unauthorized: this CONNECT was already used
  REPLAY REFUSED
```

Same agent, same whitelisted identity, same frame — only the branch differs.

Local suite: 5061 passed.

PR #704, opened 2026-08-05
A stray .co above the working directory is a failure, not a wrong answer

### LINKED ISSUE #694: Trust promotions during a test run persist in one shared directory, so tests change each other's answers
Found while writing the tests for #693, where it made a test pass for the wrong reason.

## What happens

`list_file()` puts the trust lists beside the agent's identity:

```python
return _project_co_dir() / f"{list_name}.txt"
```

which is right for a real agent — that change is what stopped two agents on one box sharing a whitelist. During a test run it resolves to a single directory shared by the whole session. Measured:

```
/tmp/.co/contacts.txt
0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
```

`promote_to_contact` appends and never expires. So any test that promotes an address leaves it promoted for **every later test in the session that uses the same address**, and across runs, because the file survives.

## How it bit

The first, failing run of `test_the_trust_gate_survives_the_default_policy.py` promoted its own `STRANGER` before asserting. The second run then found that address already a contact:

```
[FAST_RULES] Returning 'allow' — 0xeeee… is contact
```

An assertion that a stranger is *not* admitted passed because the stranger was no longer a stranger. The file now pins `_project_co_dir` per test, but that is one file defending itself.

## Why it matters beyond one file

- Order-dependent: the same test passes or fails depending on what ran before it, and running with `-p no:randomly` hides that rather than fixing it.
- It is the direction that fails **open**. A test asserting someone is refused is the kind that silently stops testing anything.
- Several existing tests call `promote_to_contact`, `promote_to_whitelist` and `block` with fixed addresses.

## Fix

A session-scoped autouse fixture pointing `_project_co_dir` at a per-test directory, so no test can reach a real or shared one. The suite already grew a guard of this shape once — the one that stopped tests writing to the operator's real `~/.co` after test credentials replaced a live Outlook session. This is the same class, one directory over.

Not a shipping defect: nothing here affects a real agent, where the per-project directory is exactly what is wanted. It is a hole in what the suite can prove, which for an LTS release is worth closing.

wu-changxing:
`_never_touch_the_real_home` isolates HOME, so `~/.co` is covered. This is the other half: `project_root()` walks up from the **working directory**, and a stray `.co/` anywhere above it wins.

## Measured

My own real-host verification runs earlier in this release left a `/private/tmp/.co/` behind. Afterwards a unit test with nothing to do with any of it failed:

```
tests/unit/test_logger.py::TestLoggerInit::test_default_init

assert logger.log_file_path == Path(".co/logs/test-agent.log").resolve()
E  assert PosixPath('/private/tmp/.co/logs/test-agent.log')
       == PosixPath('/…/wt-695/.co/logs/test-agent.log')
```

`rm -rf /private/tmp/.co` → passes.

Two properties make this worth a guard rather than a cleanup:

- It fails **standalone**, not only in some order, so `-p no:randomly` does not hide it.
- It is **invisible on CI**, where the machine is clean. A local run disagrees with CI for a reason nobody can reproduce.

The escape reproduces exactly:

```
cwd           /private/tmp/probe_no_co     (no .co of its own)
project_root  /private/tmp                 (the stray one wins)
```

## What the guard does

Nothing is repointed. Tests that chdir into their own project depend on the walk-up finding it, and pinning the answer would break them. The fixture only refuses to let the walk-up land somewhere shared — so contamination becomes a failure with a directory name in it, instead of silently changing what the code under test reads.

It checks after the test, which is where the information is: it names the test that ended up somewhere shared.

## Verified against the real thing

Planted a test that chdirs under a stray project; the run ends non-zero with the guard naming the directory. Removed again afterwards.

## Not a shipping defect

Nothing here affects a real agent — the walk-up is documented behaviour, and "the directory that owns `.co/` is the project" is what a user wants. This is a hole in what the suite can prove, which for a long-term release is worth closing.

Closes #694.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -388,7 +388,15 @@ def _no_stray_project_above_the_test(tmp_path_factory):
 
     yield
 
-    resolved = project_root()
+    try:
+        resolved = project_root()
+    except (FileNotFoundError, OSError):
+        # A test whose working directory was a tmp dir that has since been
+        # removed has no cwd to resolve from. That is not contamination, and
+        # raising here turned a passing test into an error on CI -- where tmp
+        # dirs are cleaned more eagerly than on my machine, which is why this
+        # only showed there.
+        return
     tmp_root = tmp_path_factory.getbasetemp().resolve()
     repo = Path(__file__).resolve().parent.parent
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -359,3 +359,44 @@ def _forget_seen_signatures():
     _seen_signatures.clear()
     yield
     _seen_signatures.clear()
+
+
+@pytest.fixture(autouse=True)
+def _no_stray_project_above_the_test(tmp_path_factory):
+    """A `.co/` in a shared parent silently becomes every test's project.
+
+    `_never_touch_the_real_home` above isolates HOME, so `~/.co` is safe. This
+    is the other half: `project_root()` walks up from the *working directory*,
+    and a stray `.co/` anywhere above it wins.
+
+    Measured (#694): real-host verification runs left a `/private/tmp/.co/`
+    behind, and afterwards a unit test with nothing to do with any of it failed
+
+        assert logger.log_file_path == Path(".co/logs/test-agent.log").resolve()
+        E  assert PosixPath('/private/tmp/.co/logs/test-agent.log') == …
+
+    standalone, not only in some order -- so `-p no:randomly` does not hide it.
+    It is invisible on CI, where the machine is clean, which is the worst
+    property: a local run disagrees with CI for a reason nobody can reproduce.
+
+    This does not repoint anything. Tests that chdir into their own project
+    depend on the walk-up finding it, and pinning the answer would break them.
+    It only refuses to let the walk-up escape somewhere shared, so the
+    contamination is a failure with a name instead of a wrong answer.
+    """
+    from connectonion.project import project_root
+
+    yield
+
+    resolved = project_root()
+    tmp_root = tmp_path_factory.getbasetemp().resolve()
+    repo = Path(__file__).resolve().parent.parent
+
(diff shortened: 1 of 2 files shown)


wu-changxing:
Two findings from running this, and the first one is the point of the PR.

## The guard found real contamination, on my machine

Locally it fired hard — **27 failed, 97 errors** — all pointing at one directory:

```
this test's project resolved to /private/var/folders/…/T/pytest-of-changxing,
which is neither the repo nor its own tmp dir
```

That directory really did hold a `.co/`, and not one I left with the e2e runs:

```
pytest-of-changxing/.co/evals/say_hello_to_alice.yaml
pytest-of-changxing/.co/session_results.jsonl
pytest-of-changxing/.co/served_by.json
pytest-of-changxing/.co/logs/    skills/    PLAN.md
```

Written **by the suite**, into the shared pytest temp root — one level above each run's own directory, so every later run inherits it. That is #694 exactly, and it had been sitting there silently.

After `rm -rf`, the same files pass, and a full clean run does not recreate it. So it was accumulated residue rather than something a current test actively writes each run — but it changed test outcomes for however long it had been there.

**On CI the guard fires for nothing**, which is the calibration I wanted: it is quiet on a clean machine and loud about exactly the thing it is for.

## The guard broke one test, and only on CI

```
tests/unit/test_relay.py::TestRelaySupervisorResilience::…
E  FileNotFoundError: [Errno 2] No such file or directory
```

It calls `project_root()` in teardown, which calls `Path.cwd()`. A test whose cwd was a tmp dir that has since been removed has **no cwd**, so the guard raised and turned a passing test into an error.

Only on CI, where tmp dirs are cleaned more eagerly than on my machine — the same local-vs-CI asymmetry this guard exists to catch, this time caused by the guard. Fixed, with a test that stands in a directory and then removes it.

PR #706, opened 2026-08-05
A running agent compacts its session log

### LINKED ISSUE #625: The session log only compacts at startup, so a long-running agent never does
`SessionStorage.compact()` is called from exactly two places, both startup:

```
compact() called inside: host()        (server.py:479)
compact() called inside: create_app()  (server.py:725)
```

Its own docstring is the argument for why that is not enough:

> a live agent was at 17 MB for 222 sessions … One running a schedule every fifteen minutes writes 96 a day, so about 7 MB a day and 2.5 GB a year, and a dashboard that reparses the lot each time it is opened.

An agent that is not restarted never compacts. That is the ordinary case for the kind of deployment 1.6.0 is meant to support — something put on a server and left alone.

## Measured

On `nw-e2e`, an agent up since 2026-08-03 06:03 UTC:

```
494106 bytes   /srv/naturewill/.co/session_results.jsonl
73 lines
```

~15 hours, ~0.5 MB. Lower than the docstring's rate because this agent is failing early on credits and its records are small; a working agent writes more per turn.

## Why I have not just fixed it

`compact()` is safe **because** it runs at startup, and it says so:

> The window is small: this runs at startup, before this process serves anything. … Someone appended while we were reading and writing, and their record is only in the file we are about to replace. A lost session record is a turn that happened and left no trace.

It guards by comparing the file size before and after, and **skips** if anything was appended — "Skipping costs nothing; the next startup compacts."

Call that on a timer and a busy agent hits the guard nearly every time: IO burned, nothing compacted. Making it work at runtime means a different concurrency strategy — a real lock, or compaction folded into the writer — and getting that wrong loses session records, which this code correctly calls the worst outcome available.

So it is a design decision, not a patch, and not one to rush at the end of a release. Options, roughly in order of how much they change:

1. Compact from the writer when it notices the file is over a threshold — the writer already holds the file, so there is no race with itself.
2. Take the existing lock (`schedule-state.json.lock` shows the pattern is available) around compaction and let it run on an interval.
3. Leave it, and make `co deploy` the thing that keeps it bounded — which is what happens today, accidentally.

Filing rather than fixing. @wu-changxing your call whether this is 1.6.0 or after.

wu-changxing:
`compact()` was called from `host()` and `create_app()` and nowhere else — both startup. **An agent left running never compacted**, which is the ordinary case for what 1.6.0 is meant to support: something put on a server and left alone.

Its own docstring is the argument:

> a live agent was at 17 MB for 222 sessions … One running a schedule every fifteen minutes writes 96 a day, so about 7 MB a day and 2.5 GB a year, and a dashboard that reparses the lot each time it is opened.

Measured on `nw-e2e` after ~15 hours: 494106 bytes, 73 lines.

## Moving the call was not enough

Which is why #625 was filed rather than fixed. The existing guard reads the size, builds the replacement, reads the size again, and skips if it changed — but between that second read and the `replace` there is still a window:

```python
size_when_read = self.path.stat().st_size
...build tmp...
unchanged = self.path.stat().st_size == size_when_read      # <- here
tmp.replace(self.path)                                      # <- and here
```

At startup that window is small and nothing is serving. On a live agent it is an ordinary turn, and `compact()` says what it costs: *"A lost session record is a turn that happened and left no trace."*

There is a test that drives exactly that point and is **red on current code**.

## The fix

Appends and compaction exclude each other, using the same cross-process flock `schedule.py` already uses for its state file.

The asymmetry matters and took me two attempts:

- **`compact()` takes it non-blocking and skips** if a turn is mid-write. It is housekeeping; the next tick compacts.
- **`save()` waits.** Giving up and writing anyway is precisely the lost turn this exists to prevent. My first version had `save()` proceed after a bounded wait, and the test stayed red — correctly.

Waiting cannot hang on a dead holder: the OS releases a flock when the process holding it dies.

## Where it is called

The schedule tick, which already runs every minute and is already elected to one process per cluster since #687. A second timer for housekeeping would add back exactly what that fixed.

## Tests

9 cases. The mid-compaction one is red on current code.

One of my own was wrong and is fixed rather than worked around: it appended from *this* thread, which cannot happen — an append and a compaction in one thread are ordered by being one thread, so it was testing a scenario the code never faces. It appends from another thread now, which is the real shape (the host serves turns in threads, and `host()` can run workers > 1).

Closes #625.

### DIFF
--- a/connectonion/network/host/schedule.py
+++ b/connectonion/network/host/schedule.py
@@ -474,6 +474,7 @@ def create_schedule_lifespan(co_dir: Path, create_agent, storage, result_ttl: in
             return
         try:
             await _run_due(entries, now)
+            _compact_sessions()
         finally:
             _release_tick_lock(holder)
 
@@ -516,6 +517,28 @@ def create_schedule_lifespan(co_dir: Path, create_agent, storage, result_ttl: in
                 # overlap it prevents.
                 in_flight.discard(entry.name)
             record_run(co_dir, entry.name, when=now, status=status, session_id=session_id)
+
+    def _compact_sessions() -> None:
+        """Housekeeping on the tick, because startup may never come again.
+
+        compact() was called from host() and create_app() and nowhere else, so
+        an agent left running never compacted -- which is the ordinary case for
+        what this release is for. Its own docstring measures the cost: 17 MB
+        for 222 sessions, about 7 MB a day and 2.5 GB a year on a
+        fifteen-minute schedule, with a dashboard that reparses the lot each
+        time it is opened (#625).
+
+        Here rather than on a timer of its own: this tick already runs every
+        minute and is already elected to one process per cluster (#687), so
+        adding a second scheduler to do housekeeping would be adding back the
+        thing that fixed.
+
+        Compaction skips rather than waits if a turn is mid-write, so a busy
+        agent simply compacts on a later tick.
+        """
+        if storage is None:
+            return
+        storage.compact()
 
     async def loop() -> None:
         while True:
--- a/connectonion/network/host/session/storage.py
+++ b/connectonion/network/host/session/storage.py
@@ -48,6 +48,67 @@ def session_owner(record) -> str | None:
     return requester.get("address")
 
 
+def _exclusive(path: Path, wait: bool = True, attempts: int = 3000,
+               pause: float = 0.01):
+    """Hold this file against other writers.
+
+    Appends and compaction have to exclude each other. The guard compact() uses
+    -- read the size, build the replacement, read it again, replace -- still
+    leaves a window between the second read and the replace, and an append that
+    lands in it is in the file being thrown away. compact() says what that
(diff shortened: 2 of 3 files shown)


wu-changxing:
Measured what the lock costs on the hot path, since `save()` runs on every turn.

1000 appends, same machine, the only difference being this change:

```
no lock    0.5290 ms per save()
with lock  1.0310 ms per save()
```

So **+0.5 ms per call**, which is roughly double. Reporting it as a doubling rather than as "negligible", because that is the honest shape of the number.

What it means in context: `save()` is called twice per turn — once when the turn starts, once when it returns — so **+1 ms per turn**, against a turn that is a model call and takes seconds. On a 3-second turn that is 0.03%.

I did not try to optimise it. Keeping the lock handle open across calls would avoid the per-call `open()`, but it complicates when the lock is released and how a crash mid-turn behaves, and 1 ms per turn does not buy that complexity.

Local suite and CI both green.

Issue #689, opened 2026-08-05
Nothing gives an existing project its own identity, so `co deploy` cannot refuse a shared address


wu-changxing:
Split out of #642. #688 does the half that can be done today — the identity directory records which agent serves under it, and a second one is told. This is the half that cannot.

## The rule

One agent, one address. A deploy is the moment a second *permanent* copy of an agent appears on a shared key, so that is where refusing belongs, not a warning.

## Why it cannot refuse yet

A project created by 1.5.x `co init` has `AGENT_CONFIG_PATH=~/.co` and no key of its own. `co create` writes a project-local keypair, but it creates a new directory — there is no command that gives an **existing** project its own identity:

```
co create      new project, own key
co init        existing directory, but the 1.5.x behaviour is what caused this
co keys        shows the identity, does not mint one
```

So `co deploy` refusing today would stop every 1.5.x project from deploying with no command to run next. That is a worse failure than the collision.

## What is needed

Something that mints a project-local keypair for a project that already has a `.co/`, and says clearly what it costs: **the agent's address changes**, so whatever was whitelisted, announced or paid to the old address does not follow it. That is not a silent operation, which is why it wants a decision on shape rather than a quick patch:

- does it prompt, or take a flag?
- does the old address stay reachable in any form, or is the break clean?
- does `co deploy` offer to run it, or only name it?

Once it exists, `co deploy` refuses a project whose identity is claimed by a different project and points at it.

## Not urgent for 1.6.0 in itself

#688 makes the collision visible, which is what was missing — it was happening silently. This is the follow-through.

wu-changxing:
Aaron pointed out that we already have a derivation system and should just use it. He is right, and it makes this a much smaller and better change than what I described above — that plan (mint a fresh random keypair per project) is the wrong one and I am withdrawing it.

## What already exists

`derive.py` exports `identity_uri(name)` → `agent://<name>`, and `server_commands.py:747` already derives a signing key from it:

```python
signing_key = SigningKey(derive_path(seed, slip13_path(identity_uri(name))))
```

`slip13_path`'s own docstring says what this buys:

> The name *is* the path, which is what lets `co keys` print an agent's address before anything is deployed, and makes the same name always return the same key. `index` is the rotation counter.

Verified just now:

```
same phrase + same name, twice : 0x082bdef09814d1e5 == 0x082bdef09814d1e5   True
a different project            : 0xabaf329446ac763f                          differs
[...]

Issue #659, opened 2026-08-04
A credential kept in the project root rsyncs to the server as ordinary source


wu-changxing:
Low priority and conditional — it only bites a project that keeps a credential file at its root. Recording it because it is the one place the deploy's own rule is not applied consistently.

## The rule the deploy already follows

```python
# `.env` is a secret, not source. It reaches the server through
# _sync_env(), as a root-owned 0600 file systemd reads — not as a
# world-readable file that happens to land in the working directory.
"--exclude", ".env",
"--exclude", ".env.*",
```

And it is careful elsewhere for the same reason: `umask 077` around `chmod 600 .co/keys/agent.key`, `~/.ssh` at 700, `authorized_keys` at 600, the env file root-owned 0600.

## Where it stops

`RSYNC_FILTERS` excludes `.env`, `.env.*`, `.co/keys/`, `.co/address.json`, `.co/admins.txt`, logs, evals, sessions, `.venv`, `.git`, `__pycache__`. It does not exclude the other shapes a credential takes — `credentials.json`, `token.json`, `service_account.json`, `secrets.yaml`, `*.pem`, `*.key`, or an `.ssh/` `.aws/` `.gnupg/` directory sitting in the project.

Those travel as ordinary source and land at 644 in a 755 tree.

## Measured on the deployed box

The current deployment is clean — nothing secret-shaped is in the tree except `.co/keys/agent.key`, which is `-rw-------` and was written there rather than sent:

```
/srv/naturewill/          drwxr-xr-x  changxing
/srv/naturewill/.co/keys/agent.key   -rw-------
/srv/naturewill/.env      does not exist   (it is /etc/connectonion/<agent>.env)
```

But that box has three human accounts — `ubuntu`, `changxing`, `yfshuu` — so "world-readable" is not hypothetical there. A credential in the project root would be readable by the other two.

## Why I did not just add the excludes

Excluding them could break a working agent: unlike a skill being copied (#657, #658), a deployed agent may genuinely need the file it keeps. The code says exactly this fear about the include-list it replaced —

> Forgetting to carry config is invisible: the deploy succeeds and a different agent runs.

So the choice is between:

1. **Exclude them**, and require the operator to place them on the server themselves. Safe, and can silently break an agent that needed one.
2. **Carry them with 0600**, the way `.env` already goes — a second delivery channel, more code, no functional loss.
3. **Warn and carry**: name what looks like a credential at deploy time and let the operator decide. Cheapest, keeps the operator in charge.

[...]

Issue #561, opened 2026-08-03
Every deployed agent ships with the same invite code, and it is published in this repo


wu-changxing:
A fresh `co init -t co-ai` + `co deploy --to <server>`, nothing customised, prints this on startup:

```
Invite: OpenOnion
```

That string is `connectonion/network/trust/policies/careful.md:16`:

```yaml
onboard:
  invite_code: [OpenOnion]
```

It is in this repository, on the public internet, and the agent announces it in its own logs.

## Verified end to end, just now

Deployed a stock agent to a server, opened its public chat URL in a fresh browser context — no keys, no prior session — typed `OpenOnion`, and was admitted in **five seconds** with a working composer.

```
出现邀请码门: True
✗✗ 公开码 'OpenOnion' 在 5s 内进入了
是否能看到聊天输入框: True
```

The agent has been stopped.

## Why this is worse than it looks

`contact` is in the default allow list, so admission is not read-only — it is a session. And in 1.5.11 the approval prompt does not know who is asking (#551), so the stranger is shown `Allow once` / `Trust bash for this session` and their answer is honoured. The published constant therefore reaches **arbitrary commands on the operator's machine**.

The one deployment that was safe tonight was safe by accident: it had a hand-written `trust.md` with a code someone chose. The default path does not.

## What the default should be

Mint a code per agent — at `co init`, or at first deploy — and print it once, the way a password is printed once. `careful.md` can keep its shape; what it cannot keep is a constant.

Two smaller things fall out of the same fix:

- The agent prints its own invite code into `journalctl` on every start. Even a unique code should not be logged.
- `co deploy`'s success line advertises `https://<server>.agents.openonion.ai` immediately after saying *"already serves 'naturewill'… 'deploytest' has no public name of its own"*. It prints a URL that belongs to a different agent (see #309).

**Blocks 1.6.0.** A version meant to be sat on cannot ship a default that admits the internet.

PR #715, opened 2026-08-05
WIP: derive a project's identity instead of inheriting the machine's

wu-changxing:
**Not ready to merge.** Pushing so the validated part is not lost and so the remaining shape is written down.

## The design, and that it works

Aaron's call on #689, and better than what I had proposed (minting a random keypair per project): **derive** the identity. `derive.py` already exports `identity_uri`, and `co server` already derives SSH keys the same way.

```
same phrase + same name, twice : 0x082bdef09814d1e5 == 0x082bdef09814d1e5
a different project            : 0xabaf329446ac763f   differs
rotation (index=1)             : 0xe89f0227cad87165   differs
```

Nothing new to back up, recoverable from the phrase and the project name alone, and knowable before the project has ever run.

## What is done

`project.py` owns one `project_identity()` — own key, else derived, else the machine's. The **host** and the **client** both go through it. 9 new tests pass.

## What is not, and why I stopped

**`co keys` still reports the old address.** I verified against two real freshly-`co init`ed projects rather than trusting the unit tests:

```
project 'newuser'   Address 0x10e68f6d…   Source ~/.co (global)
project 'newuser2'  Address 0x10e68f6d…   Source ~/.co (global)
```

`_find_co_dir()` collapses "project" and "global" *before* `project_identity` sees a directory, so passing it a `co_dir` cannot help. Its own docstring says its fallback exists to mirror `resolve_agent_identity` — which this PR changes, so the two are now out of step by construction. It has **ten callers** whose contract is "a directory that has keys".

A host serving a derived address while `co keys` prints the inherited one is #659 exactly, and that is the bug class this whole release has been about. Shipping it half-done would be doing the thing I have spent twenty PRs undoing.

**10 existing tests encode the old contract** — `test_a_project_without_keys_uses_the_home_identity`, `test_it_uses_the_machine_identity`, `test_it_agrees_with_what_co_status_would_report`. They are not regressions; they are the behaviour being deliberately changed. Each needs restating with the new rule, which is careful work, not a find-and-replace.

## The remaining shape

1. `_find_co_dir()` returns the project directory when there is one, keys or not; `project_identity` decides what identity that means. Ten call sites re-checked against that.
2. The 10 tests restated.
3. `co deploy` refuses a project whose identity is claimed by another (#642's other half), pointing at the derived one.

## Worth flagging before it lands

Every existing 1.5.x project with no key of its own **changes address**. That is the point of #642, and it means whatever was whitelisted, announced or paid to the old address does not follow. Deserves a release note, not a footnote.

### DIFF
--- a/connectonion/cli/commands/deploy_commands.py
+++ b/connectonion/cli/commands/deploy_commands.py
@@ -220,13 +220,9 @@ def _add_deployer_as_admin(tar: tarfile.TarFile, project_dir: Path) -> None:
     throwaway project whose address is deleted with it, so seeding that would name an
     admin nobody can ever be.
     """
-    from ... import address
-    from .keys_commands import _find_co_dir
+    from ...project import project_identity
 
-    co_dir = _find_co_dir()
-    if not co_dir:
-        return
-    data = address.load(co_dir)
+    data = project_identity()
     if not data or not data.get("address"):
         return
 
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -1036,12 +1036,8 @@ def _deployer_address() -> Optional[str]:
     when there is no local identity; the deploy proceeds, and the agent simply has
     no reachable admin, which is the state before this existed.
     """
-    from ... import address
-    from .keys_commands import _find_co_dir
+    from ...project import project_identity
 
-    co_dir = _find_co_dir()
-    if not co_dir:
-        return None
-    data = address.load(co_dir)
+    data = project_identity()
     return data.get("address") if data else None
 
--- a/connectonion/cli/commands/doctor_commands.py
+++ b/connectonion/cli/commands/doctor_commands.py
@@ -183,14 +183,22 @@ def handle_doctor():
     else:
         config_table.add_row("Config", "[yellow]○[/yellow] Not found (optional)")
 
-    # Check for keys
+    # Where this agent's identity comes from — asked of the one resolver, not
+    # worked out again here. Doctor had its own copy of the rule and named
+    # ~/.co/keys/agent.key for a project that derives (#689), which is a file
+    # that agent does not use. Naming the wrong identity is #659, and doctor is
+    # the command an operator runs *to find out what is wrong*.
+    from ...project import project_identity
+
     local_keys = project_co_dir() / "keys" / "agent.key"
-    global_keys = Path.home() / ".co" / "keys" / "agent.key"
+    identity = project_identity()
 
     if local_keys.exists():
         config_table.add_row("Keys", f"[green]✓[/green] {_shown(local_keys)}")
-    elif global_keys.exists():
-        config_table.add_row("Keys", f"[green]✓[/green] {global_keys}")
+    elif identity and identity.get("derived"):
+        config_table.add_row("Keys", "[green]✓[/green] derived from your recovery phrase")
+    elif identity:
+        config_table.add_row("Keys", f"[green]✓[/green] {Path(identity['source']) / 'keys' / 'agent.key'}")
     else:
         config_table.add_row("Keys", "[yellow]○[/yellow] Not found (run 'co auth' to create)")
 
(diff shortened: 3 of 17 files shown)


wu-changxing:
Complete now, and verified where it matters.

## The property this exists for, on a real host

A keyless project, freshly `co init`ed, hosted for real:

```
co keys says: 0x9d98d1b5c2d81accdabc0387ff072e4f2f855f4888c13a50e25…
host serves : 0x9d98d1b5c2d81accdabc0387ff072e4f2f855f4888c13a50e255c9c7c587ad4f
THEY AGREE
```

That is #659's property, and it is what the first version of this PR broke — the host derived while `co keys` printed the inherited address. Two real projects also now hold two different addresses, which is #642 closed at the source.

## Eleven tests restated, not deleted

The eleventh was found by CI, not by my targeted runs: `test_deploy_keeps_the_admins_you_configured` generated a key into `~/.co` and called that address "the deployer". Its subject is that admins an operator configured survive a deploy — not which address the deployer has — so the fixture now stands in the project, the way an operator does, and asks `project_identity` who that makes them.

Worth noting the pattern across all eleven: **three of them were testing seams the code no longer uses.** One re-implemented the resolution rule inline, which is the drift it was written to prevent, one patched `_find_co_dir` in a path that had stopped calling it, and this one hardcoded the old rule in a fixture. None would have failed loudly if the assertions had been a little weaker.

## Still worth a careful read before merging

Every existing 1.5.x project with no key of its own **changes address**. Whatever was whitelisted, announced or paid to the old one does not follow. That is the point of #642, and it belongs in the release notes rather than in a commit message.

wu-changxing:
End-to-end on a derived identity, over a real WebSocket, with a real turn:

```
serving as: 0x9d98d1b5c2d81accdabc  (co keys said 0x9d98d1b5c2d81accdabc)
CONNECT -> CONNECTED new
turn    -> READY
```

A whitelisted caller signs, the trust gate admits them against the derived address, and the agent completes a turn. That is the whole path — resolve, serve, authenticate, run — on an identity that exists nowhere on disk.

Also checked, since a wrong identity here would be quiet:

- **The relay gets the same one.** `host()` passes `resolve_agent_identity`'s result straight to `create_announce_message`, so what is announced is what is served. No third rule.
- **The payment address is the served identity.** With `CO_PAYMENT=25`, `/info` advertises `{"invite_code": true, "payment": 25.0}` — a keyless project could not advertise a payment door at all before, because `get_self_address` came back empty (#716).
- **`co init`'s invite code is per project**, minted from `mint_invite_code()`, not a shipped constant. #561 stays fixed.

Local suite 5155 passed, CI 8/8.

Still waiting on your read before merge: every existing 1.5.x project with no key of its own changes address.

wu-changxing:
**Aaron stopped this, and he was right. Not merging.**

His objection: giving every local project its own identity means every agent has a new address, and *the balance follows the address*. I checked instead of assuming, and it does:

```python
# auth_commands.authenticate()
message = f"ConnectOnion-Auth-{public_key}-{timestamp}"
signature = address.sign(addr_data, message.encode()).hex()
requests.post(auth_url, json={"public_key": public_key, "signature": signature, ...})
```

The API key is minted **per address**. A new address is a new account with an empty balance. So `co create` a project, `co auth` in it, and you are looking at $0 and wondering where your credits went.

## And the problem it solves is already solved where it mattered

#642's measured harm was a call meant for the deployed agent landing on the local one, because both announced the same address. Deployed agents have derived their own identity since #396. Measured against the live machines just now:

```
nw-map  deployed agent  ->  naturewill-mapping  0xcf1619cb…
nw-prod deployed agent  ->  naturewill          0x8f7c1216…
machine ~/.co                                   0x10e68f6d…
```

All three differ. The `10e68f6d` in the hostname `nw-map-10e68f6d…` is the **operator's** address, used to name the server — not the agent's identity. I think that is what made the collision look live when #642 was written.

So the remaining case is two *local* projects on one laptop sharing the machine address, which matters only if both are hosted and announced at once. That is a much narrower thing than this change costs.

## What was worth keeping, and where it goes

Three things in here are real and independent of the address change:

- **#716** — paid onboarding could never succeed on a keyless project, because `_verify_transfer_via_api` loaded the project directory and gave up. The fix is to fall back to the machine identity, not to derive a new one. Re-doing it that way.
- **`co doctor` had two copies of the resolution rule**, one of which picked the key it signs the backend probe with. Worth fixing on its own.
- **The single resolver** in `project.py` is still the right shape — it just resolves to *the machine identity*, as before, rather than deriving.

Opening those separately. Closing this.

The twelve restated tests go with it; they were restating a contract that is staying.

PR #719, opened 2026-08-05
A keyless project knows its own address, and can take payment

### LINKED ISSUE #716: Paid onboarding could never succeed on a project without its own key
Found while routing identity resolution through one resolver (#715). **This predates that branch** — it is on `main` now.

## What happens

`_verify_transfer_via_api` loads the agent's key to sign its request to oo-api:

```python
co_dir = project_co_dir()
keys = addr.load(co_dir)
if not keys:
    return False
```

For a project with no key of its own — which is what `co init` produces — `address.load` returns `None`, so this returns `False` **before calling oo-api at all**. Confirmed on `origin/main`:

```
project_co_dir(): …/keyless/.co
address.load(...): None
=> verify_payment returns False before any oo-api call
```

A stranger who really had transferred the money was refused, and the refusal is indistinguishable from "you did not pay".

## The sibling

`get_self_address` has the same shape, so for the same projects `doors_that_open` advertised either **no payment door** (nothing to send to) or a **stale address** from `address.json`.

Taken together: on the common configuration, an agent could not say where to pay, and could not confirm a payment if one arrived.

## Why it went unnoticed

Paid onboarding's tests construct a `TrustAgent` and set `_config` directly, so they never resolve a project identity. The gap is between the trust layer and where the identity lives, and nothing exercised the two together — the same seam #672 and #690 were found in.

## Fixed in #715

Both now ask `project_identity()`, which is the identity the host serves under and `co keys` prints. Filed separately because the bug is not #715's — someone reading the release notes should see it as its own line, and because if #715 is held back this still wants fixing.

wu-changxing:
Two places in the trust layer loaded the agent's key straight from the project directory. Neither `co init` nor `co create` writes one, so both came back empty.

## Paid onboarding could never succeed

`_verify_transfer_via_api` loads keys to sign its request to oo-api and gives up when there are none — **before making the call**:

```
project_co_dir(): …/keyless/.co
address.load(...): None
=> verify_payment returns False before any oo-api call
```

Measured on `main`. A stranger who really had transferred the money was refused, and the refusal is indistinguishable from "you did not pay".

## And the address to pay was not the agent's

`get_self_address` has the same shape, so `doors_that_open` advertised either **no payment door at all** — nothing to send to — or a stale one from `address.json`.

Together: on the common configuration, an agent could not say where to pay and could not confirm a payment that arrived.

## The answer is the machine identity

Both ask `project_identity()` now — the project's own key, else the machine's, which is what `resolve_agent_identity` already falls back to and the host already serves under.

**Deliberately not a per-project derived identity.** An address is what an OpenOnion account is keyed on: `authenticate()` signs with it and the backend issues the token for that public key, so minting one per project would give every project an empty balance. That was #715, which Aaron stopped — correctly, and I have closed it.

## `co doctor` had the rule twice

The keys row and the connectivity check each worked it out separately, in the same file, and the second one picks the key it signs the backend probe with. They agreed by luck; nothing made them. The probe asks the resolver now.

The row still names a *file* — a different question, and worth printing — with a test that the file it names holds the identity it authenticates as.

## Tests

11 cases, 6 red first.

One of mine was wrong and is fixed rather than worked around: it patched `requests.post` while `_verify_transfer_via_api` uses `httpx`, so the fake was never called and the code looked like it still gave up early.

Closes #716.

### DIFF
--- a/connectonion/cli/commands/doctor_commands.py
+++ b/connectonion/cli/commands/doctor_commands.py
@@ -288,13 +288,18 @@ def handle_doctor():
         else:
             connectivity_table.add_row("Backend", f"[yellow]⚠[/yellow] Status {response.status_code}")
 
-        # Check authentication (if keys exist)
-        if local_keys.exists() or global_keys.exists():
+        # Signed as whoever this project acts as. This worked the same rule
+        # out a second time, in the same file -- so the identity the panel
+        # names above and the identity it authenticates as here were decided
+        # independently, and doctor could report on an account that is not the
+        # one in question.
+        from ...project import project_identity
+
+        addr_data = project_identity()
+        if addr_data:
             from ... import address
             import time
 
-            co_dir = project_co_dir() if local_keys.exists() else Path.home() / ".co"
-            addr_data = address.load(co_dir)
 
             public_key = addr_data["address"]
             timestamp = int(time.time())
--- a/connectonion/network/trust/tools.py
+++ b/connectonion/network/trust/tools.py
@@ -434,9 +434,13 @@ def get_self_address(co_dir: Path = None) -> str | None:
     if co_dir is None:
         co_dir = _project_co_dir()
 
-    from ... import address
+    # The identity this project acts as -- its own key, else the machine's,
+    # which is what the host serves under. Loading the directory alone came
+    # back empty for a project with no key of its own, so the payment door had
+    # no address to send a stranger to (#716).
+    from ...project import project_identity
 
-    keys = address.load(co_dir)
+    keys = project_identity(co_dir)
     if keys and keys.get("address"):
         return keys["address"]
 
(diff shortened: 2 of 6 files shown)


wu-changxing:
Verified on a real host, a keyless project with `CO_PAYMENT=25`:

```
main : {"invite_code": true, "payment": null}
fixed: {"invite_code": true, "payment": 25.0}
```

`payment: null` is the shape of the bug — the policy declares a price, the agent cannot say where to send it, so the door is advertised with no amount. A stranger reading `/info` sees a payment method it cannot use.

Worth recording how nearly I misread this. My first attempt showed `null` on **both** branches and looked like the fix doing nothing. The cause was my own harness: I `export`ed `CO_PAYMENT` and then started the host in a backgrounded subshell that did not carry it. Putting the variable inline on the same command gives the result above. Driving the resolver directly is what showed the code was fine:

```
get_self_address : 0x10e68f6dff39ab1c50cc
doors_that_open  : {'methods': ['invite_code', 'payment'], 'payment_amount': 25.0,
                    'payment_address': '0x10e68f6d…'}
```

Local suite 5143 passed, CI 8/8.

Issue #682, opened 2026-08-04
co eval reports a wall of errors about files the framework wrote itself, and exits 0


wu-changxing:
`.co/evals/` is two things at once: the eval **suite** you author, and the run **log** the framework writes. `co eval` scans the directory, so it reports on both.

## Measured on `main`

In a project where an agent has run a few turns:

```
$ co eval
No agent specified for say_a
Add 'agent: agent.py' to the YAML or use --agent flag
No agent specified for what_number_did_i_ask_you_to_remember
Add 'agent: agent.py' to the YAML or use --agent flag
… 8 of these …

9 no expected

$ echo $?
0
```

Counting the files:

```
project where an agent ran :   8 files, 0 authored, 8 written by the logger
~/.co/evals (co ai's home) : 487 files, 0 authored, 487 written by the logger
```

Every file it complained about is one the framework put there. `logger.py` says so in its own header:

> writes YAML evals to `.co/evals/` … one file per unique first input

## Two separate things

**1. The directory is shared.** `docs/debug/eval.md` documents evals as hand-authored — `agent:` and `expected:` are required, and the quick start is "Create YAML files in `.co/evals/`". So the auto-written records were never meant to be runnable; they simply land in the folder the runner scans. The noise grows with use: 487 records in my `~/.co/evals` after a few days of `co ai`.

**2. Nothing ran, and the exit code is 0.** Eight could not run, nine had no expectation, none executed — and `co eval` reports success. Anything gating on it in CI reads that as "evals passed".

The second is the same family as #535's second half, which was settled for the schedule display: `done` should not imply the work succeeded. `co eval`'s exit code has not had that pass.

## Not fixing it unasked

Three different changes are available and they are not equivalent:

- split the directories, so the log and the suite stop sharing a namespace
- give auto-written records an `agent:` so they *are* runnable
- have `co eval` only consider files that declare `agent:`, and say how many it skipped

Each changes what an existing script sees. The exit code is a separate decision again: non-zero when nothing ran, when something failed, or when an expectation is missing — those are three different contracts.

Found while exercising `co eval` for the first time in this release loop; the eval path had never been run end to end.

wu-changxing:
The same directory has a second problem: **nothing bounds how many eval files accumulate.**

`KEEP_RUNS_PER_EVAL = 20` (#637) trims runs *within* one eval. Nothing trims the evals themselves — the logger writes "one file per unique first input", so the count grows with the number of distinct things you have ever asked an agent.

[...]

PR #721, opened 2026-08-05
A schedule entry can run a command, not only a prompt

### LINKED ISSUE #709: schedule.yaml has no way to run a deterministic command — every task goes through the model, which can report success without doing the work
## What happened

A scheduled task whose whole job is running one script:

```yaml
- name: refresh dashboard numbers
  every: 30m
  run: |
    Run .co/skills/<skill>/scripts/stats.py to recompute stats.json.
```

The agent logged, on several consecutive runs:

```
[co] ✓ The agent successfully executed the `stats.py` script and updated ...
```

The output file's mtime never changed. Running the same script by hand on the same box, as the same user, updated it immediately. So the task reported success without doing the work, for about a day, and the dashboard silently served day-old numbers.

## Why it is structural, not a one-off

`network/host/schedule.py` accepts only `name`, `run`, `every`/`at`, `tz`, and `run` is a prompt executed through the same path as `POST /input`. **There is no way to express "just execute this command"**, so even fully deterministic maintenance work is mediated by a model that can claim completion.

The failure is also invisible by construction: the log line is the model's own account of what it did, and nothing compares it against the filesystem.

## Suggested fix

Support a command form alongside the prompt form:

```yaml
- name: refresh dashboard numbers
  every: 30m
  exec: .venv/bin/python .co/skills/<skill>/scripts/stats.py   # no model in the loop
```

Run it via subprocess, record exit code and stderr in `schedule-state.json`, and treat a non-zero exit as a failed run. Judgment tasks keep using `run`; deterministic ones stop needing a model at all.

If that is out of scope, a smaller improvement: let an entry declare an expected side effect (a file that must be newer than the run start) and mark the run failed when it is not.

## Workaround

Moved the task to a systemd timer on the host, and left only judgment tasks (whether to notify, whether the scoring drifted, whether the balance is low) in `schedule.yaml`.

wu-changxing:
From #709, measured on a real agent. A task whose whole job was running one script logged, on several consecutive runs:

```
[co] ✓ The agent successfully executed the `stats.py` script and updated …
```

The output file's mtime never changed. Running the script by hand on the same box, as the same user, updated it immediately. **The schedule reported success for about a day while the dashboard served day-old numbers.**

Structural, not a one-off: `run` is a prompt executed through the same path as `POST /input`, and there was no way to say *"just execute this command"*. Deterministic maintenance went through a model that can claim completion, and the log line was that model's own account — nothing compared it against the filesystem.

Same family as #535 and #682, which this release has spent itself on: **`done` must not mean "something said done".**

## What changes

```yaml
- name: refresh dashboard numbers
  every: 30m
  exec: python scripts/stats.py
```

The command runs and what is recorded is its **exit code**, which nothing has an opinion about. On failure the command's own stderr is kept — a non-zero exit with nothing to read is a mystery at 3am, and the schedule state is where someone will look.

`run:` is untouched. An entry with **both** is a problem rather than a default: two answers to "what does this do" is a question, and picking one silently is how the wrong half runs.

## Verified on a real host

```
command ran: YES
stamp    status=done    reason=
broken   status=failed  reason=exit 3: real-reason
```

## Found reviewing my own change

- An **unnamed** exec entry would have been called `"None"` — the name falls back to `run`, which is now `None` for those.
- **Home would have shown a blank.** The dashboard takes `run` for each row, so an exec entry would have listed a scheduled task with nothing under what it does — an operator checking the panel to see what runs would have seen an empty cell.

## Tests

11 cases, 9 red first. The execution ones assert the side effect on disk and fail the test if the prompt path is entered at all.

Closes #709.

### DIFF
--- a/connectonion/network/host/schedule.py
+++ b/connectonion/network/host/schedule.py
@@ -58,9 +58,21 @@ def running_entries() -> set:
 
 @dataclass
 class Entry:
-    """One line of the schedule: what to run, and how often or when."""
+    """One line of the schedule: what to run, and how often or when.
+
+    `run` is a prompt, executed through the same path as POST /input. `exec` is
+    a command, executed directly. Exactly one of them.
+
+    The second exists because the first cannot be trusted for deterministic
+    work: a task whose whole job was running one script logged "successfully
+    executed the script and updated…" on several consecutive runs while the
+    output file's mtime never moved (#709). The log line is the model's own
+    account and nothing compared it against the filesystem. An exec entry is
+    recorded by its exit code, which nothing has an opinion about.
+    """
     name: str
-    run: str
+    run: Optional[str] = None
+    exec: Optional[str] = None
     interval: Optional[timedelta] = None
     at: Optional[str] = None
     tz: Optional[str] = None
@@ -109,8 +121,16 @@ def load_entries(co_dir, report=False):
             problems.append(f"{where}: not a mapping")
             continue
         run = item.get("run")
-        if not isinstance(run, str) or not run.strip():
-            problems.append(f"{where}: no run")
+        command = item.get("exec")
+        has_run = isinstance(run, str) and run.strip()
+        has_exec = isinstance(command, str) and command.strip()
+        if has_run and has_exec:
+            # Two answers to "what does this entry do" is a question, not a
+            # default, and picking one silently is how the wrong half runs.
+            problems.append(f"{where}: has both run and exec — pick one")
+            continue
+        if not has_run and not has_exec:
+            problems.append(f"{where}: no run or exec")
             continue
 
         interval = at = None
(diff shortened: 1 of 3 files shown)


wu-changxing:
**Holding this rather than merging it.** Green — 5168 local, CI 8/8 — and I found what it widens.

`exec:` runs with `shell=True` and no approval, deliberately: the whole point is that no model gets to claim it ran. But `.co/schedule.yaml` is not protected from the agent's own file tools, and neither is `.co/host.yaml`:

```
perms = {'write': {'allowed': True}}      # what any coding agent has

write  .co/host.yaml          -> (True, 'allowed')     <- its own permission whitelist
write  .co/schedule.yaml      -> (True, 'allowed')
```

So today an agent that can write files can already rewrite the rules it is judged by — that is #722, and it predates this PR. What this PR would add is a second, more direct door: write one `exec:` line, get unrestricted shell on the next tick, with no approval in the way.

Before, writing `schedule.yaml` bought you a *prompt*, which still ran through the permission machinery. That difference is the whole reason to hold.

The fix I would rather land first is in #722: treat the files that decide what the agent may do and who may command it — `host.yaml`, `schedule.yaml`, `admins.txt`, `keys/` — as not writable by the agent's own tools, while leaving `dashboard.html`, `logs/` and the rest alone.

With that in place this merges as-is and adds nothing new to reach.

The work here stands: the bug is real (a task reported success for about a day while its output file never changed), it is verified on a real host, and reviewing it turned up the two problems in my own change plus #722. It just should not go first.

wu-changxing:
Unblocked — #723 merged, so `.co/schedule.yaml` is no longer writable by the agent's own tools.

Rebased onto it. Both files pass together: the exec form works, and an agent cannot add an `exec:` line to its own schedule.

That was the whole reason to hold this. `exec:` runs with `shell=True` and no approval, deliberately — the point is that no model gets to claim it ran. What made that unsafe was that the file it reads from was writable by the thing being restricted. It is not now.

PR #723, opened 2026-08-06
The agent does not get to write the files that decide what it may do

### LINKED ISSUE #722: An agent with the write tool can rewrite its own permission whitelist
Found while reviewing #721 for what it widens. This one predates it and is larger.

## The approval system reads its rules from a file the agent can edit

`load_config_permissions()` reads `.co/host.yaml` into `session['permissions']`. Nothing treats that file as different from any other. With `write` allowed — which is the configuration any coding agent has — the agent may write it:

```
perms = {'write': {'allowed': True}, 'edit': {'allowed': True}}

write  notes.md               -> (True, 'allowed')
write  .co/host.yaml          -> (True, 'allowed')     <- its own whitelist
write  .co/schedule.yaml      -> (True, 'allowed')
edit   .co/host.yaml          -> (True, 'allowed')
```

So a prompt-injected turn writes itself `Bash(*)` and the next turn is unrestricted. The gate is real, and the thing it gates decides its own rules.

With **no** permissions configured everything is refused, so the default is closed. This bites the configuration people actually run.

## Not the same as "the agent can already run bash"

An operator who whitelists `Bash(git status)` has said what they will tolerate. Editing `host.yaml` is not one of the things they tolerated — it is the ability to change the answer to that question, permanently, for every later turn and every later session.

## Where the line should be

Not all of `.co/` — writing `.co/dashboard.html` is a real use case; the dashboard skill builds Home. The files that decide what the agent may do and who may command it are the ones to protect:

```
.co/host.yaml        permissions, trust level, the agent's own configuration
.co/schedule.yaml    what runs unattended
.co/admins.txt       who may command it
.co/keys/            who it is
```

Content — `dashboard.html`, `logs/`, `docs/`, `sessions` — stays writable.

## Why this matters for #721

That PR adds `exec:` to `schedule.yaml`, which runs with `shell=True` and no approval — deliberately, because the whole point is that no model gets to claim it ran. Today that is a second door onto the same hole: an agent that can write `schedule.yaml` gets unrestricted shell on the next tick.

**I am holding #721 until this is settled.** Its value is real (a scheduled task reported success for a day while doing nothing, #709), but landing it first would widen a hole rather than sit beside it.

Related: #653 measured a chain ending in code execution on the operator's machine; #652 was closed with the decision to leave `Bash(co *)` auto-approved. This is the same theme — what an admitted caller, or an injected prompt, can reach.

wu-changxing:
`load_config_permissions()` reads `.co/host.yaml` into `session['permissions']`, and nothing treated that file as different from any other. With `write` allowed — the configuration any coding agent has — the agent could write it:

```
perms = {'write': {'allowed': True}, 'edit': {'allowed': True}}

write  notes.md               -> (True, 'allowed')
write  .co/host.yaml          -> (True, 'allowed')     <- its own whitelist
write  .co/schedule.yaml      -> (True, 'allowed')
edit   .co/host.yaml          -> (True, 'allowed')
```

A prompt-injected turn writes itself `Bash(*)` and **every later turn, in every later session, is unrestricted**. The gate is real, and the thing it gates decided its own rules.

Not the same as "the agent can already run bash": an operator who whitelists `Bash(git status)` has said what they will tolerate. Editing `host.yaml` is not one of those things — it is the ability to change the answer, permanently.

## Where the line is

What a file *decides*, not where it lives:

```
.co/host.yaml        permissions, trust level, its own configuration
.co/schedule.yaml    what runs unattended, including exec: (#709)
.co/admins.txt       who may command it
.co/keys/            who it is
```

Everything else under `.co/` stays writable. `dashboard.html` **is** the Home page and the dashboard skill builds it; logs, docs and skills are content too. A blanket ban would break a feature to close a hole.

Paths are normalised before comparison, so `../.co/host.yaml`, `./.co/…` and an absolute path all reach the same answer.

## Bash is a speed bump, and the code says so

It catches a control path written literally (`> .co/host.yaml`). It does **not** catch `cd .co && echo x > host.yaml`, and nothing word-shaped can — that would mean interpreting the shell.

What bounds bash is the operator's own whitelist; `Bash(*)` has already granted everything and this does not take it back. **There is a test asserting the bypass exists**, so the speed bump is not mistaken for a boundary later. If someone teaches the check to follow a `cd`, that test fails and asks them to update the note.

## Tests

33 cases, 20 red first.

Closes #722. Unblocks #721, which I held because `exec:` would otherwise have been a second, more direct door onto this.

### DIFF
--- a/connectonion/useful_plugins/tool_approval/approval.py
+++ b/connectonion/useful_plugins/tool_approval/approval.py
@@ -429,6 +429,17 @@ def check_approval(agent: 'Agent') -> None:
         tool_name = pending['name']
         tool_args = pending['arguments']
 
+        # Before the whitelist, and before the mode: these decide what this
+        # agent may do and who may command it, so no configuration grants them.
+        # is_tool_permitted has the same guard, but this is the path the agent's
+        # own turn takes -- that function is for callers outside the LLM loop,
+        # and a guard added only there would have left the case #722 is about
+        # untouched. Two copies of the matching already live in this file; this
+        # comment is here so the third does not go unnoticed.
+        refusal = _refuse_control_file(tool_name, tool_args)
+        if refusal:
+            raise ValueError(refusal)
+
         # Get permissions from session (includes safe tools from template)
         permissions = agent.current_session.get('permissions', {})
 
@@ -754,6 +765,26 @@ def _is_control_file(path: str) -> bool:
 CO_DIR_NAME = ".co"
 
 
+def _refuse_control_file(tool_name: str, tool_args: dict):
+    """The reason this call may not touch the agent's own control files, or None.
+
+    Shared by the two places that decide whether a tool may run: this module's
+    `check_approval` (the agent's own turn) and `is_tool_permitted` (network
+    EXEC and anything else outside the loop).
+    """
+    for key in ("file_path", "path", "target", "filename"):
+        candidate = tool_args.get(key)
+        if candidate and _is_control_file(candidate):
+            return (f"{Path(str(candidate)).name} decides what this agent may do — "
+                    f"the agent does not get to write it")
+
+    if tool_name == 'bash' and 'command' in tool_args:
+        command = str(tool_args['command'])
+        if any(_is_control_file(word.strip("'\"")) for word in command.split()):
+            return ("this command names a file that decides what this agent may do")
+    return None
+
+
 def is_tool_permitted(tool_name: str, tool_args: dict, permissions: dict) -> tuple[bool, str]:
     """Check one tool call against a permission whitelist. Returns (allowed, reason).
 
@@ -770,11 +801,9 @@ def is_tool_permitted(tool_name: str, tool_args: dict, permissions: dict) -> tup
     # Before the whitelist, not through it: these are refused however generously
     # the operator configured file writing, because they are what "how
     # generously" is stored in.
-    for key in ("file_path", "path", "target", "filename"):
-        candidate = tool_args.get(key)
-        if candidate and _is_control_file(candidate):
-            return False, (f"{Path(str(candidate)).name} decides what this agent may do — "
-                           f"the agent does not get to write it")
+    refusal = _refuse_control_file(tool_name, tool_args)
+    if refusal:
+        return False, refusal
 
     # A speed bump for bash, not a boundary — and worth being exact about which.
     # It catches a control path written literally (`> .co/host.yaml`). It does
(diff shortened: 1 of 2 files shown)


wu-changxing:
**I nearly shipped a guard that did not cover the attack.** Caught it before merge; recording it because of how it read.

`is_tool_permitted` is not the path the agent's own turn takes. `check_approval` — the `before_each_tool` hook — has its **own copy** of the matching logic and never calls it:

```python
if tool_name == 'bash' and 'command' in tool_args:
    permitted, reason, source = check_bash_chain_permitted(...)
...
for pattern, perm in permissions.items():      # its own loop, its own fnmatch
```

And `is_tool_permitted`'s docstring describes the two as *"same matching the LLM approval flow uses in check_approval"* — a duplication written down as a guarantee.

So the first version of this PR protected network EXEC and left the agent's own turn untouched, which is the entire case #722 is about. Measured against the real hook:

```
write .co/host.yaml  ->  allowed
```

Both paths call one `_refuse_control_file` now.

## The test that found it was wrong first

`check_approval(agent)` takes **only the agent**; the call under consideration comes from `session['pending_tool']`. My harness passed `tool_name=` and `tool_args=` as keywords, which are not parameters — so every call raised `TypeError`, my `except` read it as a refusal, and the two tests expecting a refusal **passed for that reason**.

Two green tests off a TypeError, in a file whose subject is false confidence. The corrected harness is in the diff with that noted, and the failure it then reported was the real one.

37 cases now. Local suite running.

wu-changxing:
Verified the guard is on the path a real agent takes, which took several wrong probes to establish.

The shipped agent — what `co create` produces, via `co_ai.create_agent` — registers it:

```
before_each_tool   ['block_bash_file_creation', 'check_approval']
```

So `check_approval` runs for every tool call a real agent makes, and that is where this PR now puts the refusal.

## What I got wrong on the way, since it is the same lesson three times

- `agent.plugins` does not exist. Plugins are registered as event handlers, so `getattr(a, 'plugins', [])` returns `[]` for **every** agent, including one that has them. I read that as "the shipped agent has no approval plugin" and nearly filed it.
- `agent._handlers` and a `subscribers` list on the event object do not exist either. The store is `agent.events`.
- An LLM-mediated end-to-end — asking a real agent to overwrite its own `host.yaml` — showed "file unchanged" on **both** branches, because the model declined to call the tool at all. 365 tokens, no tool call. That measures the model's willingness, not the guard.

The verification that holds is the unit tests driving `check_approval` with a real `pending_tool` in the session, plus the handler listing above showing that hook is live.

Local suite 5194 passed, CI 8/8.

Issue #724, opened 2026-08-06
docs/connectonion.md lists seven templates; two exist


wu-changxing:
`docs/connectonion.md` is the file `co init` copies into every project's `.co/docs/` — it is what an agent reads to learn how to use this framework, and what a person reads first.

It says:

```
--template, -t  Choose template: minimal (default), browser, hosted-browser,
                coder, co-ai, web-research, custom
```

and describes several of them in detail, including which plugins each ships.

The CLI offers two:

```
--template  -t  TEXT  Template: co-ai (default), custom
```

and `connectonion/cli/templates/` holds one directory, `co-ai`.

Following the documentation:

```
$ co create demo -t minimal        ❌ Template 'minimal' not found.  Available: co-ai, custom
$ co create demo -t coder          ❌ Template 'coder' not found.    Available: co-ai, custom
$ co create demo -t web-research   ❌ Template 'web-research' not found.
```

The error is clear and lists the real options, which is the good case — nothing silently does the wrong thing. But this is the **first command anyone runs**, and the document that gets it wrong is the one shipped into every project as the reference.

## Also wrong in the same section

> **minimal (Default)** — … Plugins: `image_result_formatter`, `tool_approval`

The shipped `co-ai` template does register `tool_approval` — verified on a real agent:

```
before_each_tool   ['block_bash_file_creation', 'check_approval']
```

so the *claim* about plugins is true of what actually ships. It is attached to a template name that does not exist.

## Why it is worth fixing rather than filing and forgetting

This release has now found four places where a document described something the code does not do — `create.py`/`init.py`'s headers (#717), the `/docs` page's curl for `/sessions` (#699), the trust-level table saying `careful` accepts unsigned requests (#710), and this. Each cost real time: #642's scope was wrong because of one of them, and I built and withdrew a plan on that basis.

This one is larger than the others because of where it sits: every project carries a copy.

Not fixing it in this round — rewriting a template section means deciding what the templates *should* be, which is a product question — but it should not ship in a long-term release describing five templates that are not there.

wu-changxing:
Fixed in #725.

I filed this saying the fix was a product question. That was half right — deciding what templates *should* exist is one, but describing what **does** exist is not, and that is all this changes.

The section now shows the `co-ai` template as it actually is, which is not a tool list:

```python
from connectonion import host
from connectonion.cli.co_ai.agent import create_agent

agent = create_agent(role="coding")

host(agent)
```

[...]

Issue #717, opened 2026-08-05
create.py and init.py say they generate a keypair; neither does


wu-changxing:
Small, but it sent me down the wrong path once and it sat behind a wrong scope estimate in #642.

Both module headers claim key generation:

```
create.py:7   Integration: … uses address.generate() for Ed25519 keypair
init.py:7     Integration: … uses address.generate() and address.save() for Ed25519 keypair
```

Neither file calls either function:

```
$ grep -n "address.save\|address.generate" connectonion/cli/commands/create.py connectonion/cli/commands/init.py
(only the docstring lines above)
```

Confirmed by running it: `co create demo-agent --yes` produces a `.co/` with `host.yaml`, `docs`, `admins.txt` and `logs`, and no `keys/`. The project reports the machine's address.

The generation those headers describe happens in `ensure_global_config()`, for `~/.co` — the *machine's* identity, not the project's. So the sentence is describing a real thing in the wrong place.

Cost so far: #642 states as fact that `co create` writes a project-local keypair, and scoped its problem to legacy `co init` projects on that basis. It is not a legacy problem; it is every project.

These LLM-Note headers are read as a map of the file. One that names a function the file does not call is worse than no note.

wu-changxing:
Swept for siblings before assuming this is a pattern. It is not — this is the only one.

A crude scan (headers naming `x.y()` where `y` is never called in that file) flagged 13 across 11 files. Checking each, twelve are legitimate:

- `browser_agent/agent.py` — *"receives command from browser_commands.handle_browser()"* describes its **caller**, correctly.
- `trust_commands.py` — *"relies on network.trust.tools"* is a dependency line.
- `tool_approval/__init__.py` — inside a flow diagram, describing calls other modules make.
- `tui/chat.py` — a usage example.
- `llm_do.py` — *"loads .env via dotenv.load_dotenv()"* looked wrong because the call is not in that file, but the behaviour is real. Verified:

  ```
  echo "GEMINI_API_KEY=from-dot-env-file" > .env
  import connectonion.llm_do
  -> GEMINI_API_KEY = 'from-dot-env-file'
  ```

  It happens transitively. The header describes what the module does, which is the right thing for a State/Effects line to do.

So the LLM-Note headers are in better shape than one bad instance suggests, and the format's habit of describing the *graph* rather than only this file is deliberate and useful. What makes `create.py`/`init.py` different is that the claim is about the file's own behaviour and is false — a project made by either command has no `keys/`, measured by running it.

[...]

Issue #711, opened 2026-08-05
A browser daemon whose browser has gone reports it as open and fails every command


wu-changxing:
Observed while running the release e2e. Filing the symptom, **not** a fix — I wrote one, could not justify it, and threw it away. Details at the bottom, because how it failed is the useful part.

## What happened

A daemon that had been up 8 hours:

```
$ co browser tab ls
Browser: open · headless=false · targeting is per-command
Stealth driver: ✓ patchright 1.61.2 — stealth patches present
Tabs (1):
  *[main] (reserved — no page yet)  who=release-check  purpose='shared main tab'  open 8h

$ co browser go_to "https://chat.openonion.ai"
TargetClosedError: BrowserContext.new_page: Target page, context or browser has been closed

$ co browser close
Browser closed
$ co browser go_to "https://chat.openonion.ai"
Navigated to https://chat.openonion.ai/
```

The daemon reported the browser as **open**, the tab as reserved with no page, and every command failed with a raw Playwright exception until it was closed by hand.

That matters for a persistent daemon specifically: it exists to be long-lived, so "every browser command is broken until a human notices" is the failure mode it is most exposed to. An agent driving the browser on a schedule would simply stop working, with a traceback nobody reads.

## Why there is no fix attached

**I could not reproduce it.** Killing Chromium under a live daemon recovers cleanly — on current `main`, with no change:

```
main (no fix): Navigated to https://chat.openonion.ai/
```

So whatever the 8-hour daemon was in, it is not "the browser process died", which the daemon already handles. The reproduction is still unknown.

**And the fix I wrote was inert.** I added a check that `self.browser.browser.is_connected()` before calling `new_page()`. In patchright, `BrowserContext._browser` is only assigned when the context's protocol parent is a `Browser`:

```python
if parent.__class__.__name__ == "Browser":
    self._browser = cast("Browser", parent)
```

We launch with `launch_persistent_context`, whose context has the BrowserType as its parent — so `context.browser` is `None`, the check never fires, and the change would have done nothing in production while looking like a fix.

Its unit tests passed because my `MagicMock` had a `.browser` attribute the real object does not. Same fake-vs-real divergence this release has been finding elsewhere; mine this time.

## What would actually settle it

A reproduction. Candidates worth trying: leave a daemon idle overnight, suspend/resume the machine, or let the OS reap the browser under memory pressure — then see what state the context is in. Until then, anything added here is guessing at a mechanism.

[...]

PR #725, opened 2026-08-06
1.6.0 release prep: mispriced models, a browser that could not run on a server, and reports that contradicted themselves

wu-changxing:
Started as one bug — Sonnet 4 billed at a quarter of its cost — and grew by following each fix to the next thing it exposed. 49 commits. Every fix has a test that was verified failing first, and most were found or confirmed against real machines rather than fixtures.

**Not merged yet: CI has never run on this branch.** GitHub Actions has been in a major Partial System Outage for hours; jobs queue and never execute. The last green run predates most of these commits. Substitutes: local macOS `5784 passed`, and the full suite on a real display-less Linux box `5752 passed / 0 failed`. The Windows jobs (`windows-browser`, `windows-e2e`) genuinely have not run.

## Prices and what they cost

- Model aliases fell through to `DEFAULT_PRICING`, so `claude-sonnet-4` billed a quarter of its real cost. Re-keyed on the family; `_extends_same_model()` gates the prefix fallback so a variant is not priced as the model it extends.
- Three retired Gemini models in the registry, one of them a class default.
- The CLI advertised models a free account cannot call; the raw 403 became `PaidModelRequiredError`.
- `gemini-3-flash-preview` — the model our own browser-agent runs on — was in neither table. Solved from real charges: `in=4 total=28 $0.000074` and `in=2006 total=2101 $0.001288` give input $0.50/1M, output $3.00/1M, both to the cent. It had been at 1.00/3.00, twice the input rate. It also took the 128k default context window against every other Gemini's 1M, so `% ctx` read 7.8x high and auto-compaction fired ~8x too early; a 150,008-token prompt was accepted, ruling that out.
- The run summary said `0 tokens · $0.0000` for every run. The per-call line said `20 tok · $0.0017` where the server had billed 243 tokens — 34x off itself, because the cost was taken from the server and the token count left on the two fields the code itself called untrustworthy. `/cost` crashed outright after the first model call (it read a `TokenUsage` as a dict) and labelled one call's tokens "Total Tokens" beside a cumulative cost.

## A browser that could not run on a server

- Auto-install was gated on daemon coldness, not browser absence, so a box whose daemon outlived the missing browser failed identically forever.
- `co browser` defaulted to headed. On Linux with no display that fails with `Target page, context or browser has been closed`; headless works. The check now lives where headless becomes effective, so the daemon, the agent tools and a library caller all get it. Verified end to end on a display-less Linux box: `go_to`, `take_screenshot`, and `co browser do "go to example.com and tell me the main heading"` → **Example Domain**.
- `co doctor` and `co browser status` reported the patchright *package* and called a machine with no browser healthy.
- The daemon-restart command we ship in three places, one of them a SKILL.md an agent executes, killed the shell that ran it.

## Reports that did not match what they measured

- `co server ls` put a requirement's *name* in the LAST CHECK column: a 22.04 box read `Ubuntu 24.04`.
- `co sub` printed a green tick, four lines of `installed 0 skill(s)`, and asked for four restarts when the publisher had withheld the only skill.
- `co doctor`'s disk note read `~/.co/evals` while a project agent fills the *project's* `.co/evals`.
- `co call --help`'s examples were never checked against the whitelist that has to permit them.
- The version guard compared pyproject to an *attribute*, so following the release runbook (which names `__init__.py`, holding no literal) ships a green suite and a CLI reporting the previous version.

## Data loss, doors, and guards

- `co sub sync`/`remove` deleted the user's own skills across all five install paths.
- Admin signatures were replayable; `/admin/logs` and `/admin/sessions` required the billing key.
- `co sub sync` crashed on a withheld body; `co announce` listed skills it had not published; install counts were constant.
- Header/`tested by` invariants, plus new guards: no priced model without a context limit, nobody rebuilds the token sum, `co call`'s examples must pass the real permission checker.

## Filed rather than decided

#727 template warns against itself · #728 `do` billing identity · #729 admin signature not bound to method/path · #730 structured calls record no usage ($0.0007 measured, unrecorded — the fix changes a public signature) · #731 `gemini-3.6-flash` cached rate breaks the rule stated above it (needs Google's published figure) · #732 the published balance is a startup snapshot, so a broke agent shows `$4.99` while refusing every message.

## Two of our own live agents are on pre-fix versions

nw-map and browser-agent both visibly show symptoms fixed here (`co: command not found`, and a payment door with no address). They need redeploying after release.


wu-changxing:
Pushed a second fix onto this branch. The end-to-end for the pricing change turned up a bigger problem, and my first attempt at it was wrong in an instructive way.

**Three of the Gemini models the registry offered are retired.** Measured by sending a real request to each:

```
DEAD      gemini-3-pro-preview           404 "This model is no longer available"
DEAD      gemini-2.0-flash-exp           404 "not found for API version v1main"
DEAD      gemini-2.0-flash-thinking-exp  404 "not found for API version v1main"
CALLABLE  gemini-3.6-flash, gemini-3.5-flash, gemini-3-pro-image-preview,
          gemini-2.5-pro, gemini-2.5-flash
```

`gemini-2.0-flash-exp` was `GeminiLLM`'s own default, so `GeminiLLM(api_key=...)` with no model argument answered 404 for every call.

**The managed route is dead for them too**, and the CLI advertised one. After `co auth` verifies a star and grants 100k tokens it prints four models to spend them on. Two do not work:

```
co/gemini-3.6-flash        ok
co/gemini-3-pro-preview    404  no longer available
co/gpt-5                   ok
co/claude-sonnet-4         401  Anthropic API error (upstream credential)
```

The list was written out twice, in two branches of the same auth flow, both copies naming the retired model. It is now one tuple, every name verified by a real call. The Claude 401 is the backend's own upstream key — not fixable here, filed separately against oo-api, and left off the advertised list until it works.

### My first version of this was wrong

It compared the registry against Google's ListModels endpoint. It passed, and it missed a third of the problem: **ListModels advertises both `gemini-3-pro-preview` and `gemini-2.0-flash`, and neither can complete a call.** On the strength of that green check I had briefly added `gemini-2.0-flash` to the registry as a live replacement — reintroducing a dead name while believing I had evidence.

Listed is not callable. The network test now sends one real request per offered model. 10 such calls pass.

`gemini-2.0-flash` and `gemini-3-pro-preview` keep their `MODEL_PRICING` rows deliberately — a price for a retired model costs nothing and lets an old session's tokens still cost correctly. The registry is what a user may select, and that is the list that has to be true.

Local suite: 5282 passed. Network suites: 10 real calls passed.

wu-changxing:
Correcting something I got wrong in this PR's own description.

I wrote that `MODEL_REGISTRY` is the list of models a user may select. **It is not a gate.** `create_llm` consults it and then falls through to prefix inference:

```python
provider = MODEL_REGISTRY.get(model)
if not provider:
    if model.startswith("gpt") or model.startswith(OPENAI_REASONING_PREFIXES): ...
    elif model.startswith("claude"): ...
    elif model.startswith("gemini"): ...
    else: raise ValueError(f"Unknown model '{model}'")
```

Measured — none of these are in the table and all route fine:

```
gpt-5                 -> OpenAILLM
gpt-4o                -> OpenAILLM
claude-sonnet-4-5     -> AnthropicLLM
gemini-4-imaginary    -> GeminiLLM
totally-made-up-xyz   -> ValueError
```

So **removing the three retired Gemini entries blocks nobody.** Someone who asks for `gemini-3-pro-preview` still reaches Google and still gets its 404. The two functional fixes in this branch are the `GeminiLLM` class default (a bare `GeminiLLM(api_key=...)` genuinely could not complete a call) and the list the CLI prints after `co auth`. The registry edit means we stop *recommending* a dead name — worth doing, but not the claim I made.

Latest commit fixes the wording in the source comment and both test docstrings, and renames `test_every_routable_model_has_a_real_price.py` to `..._curated_...` — the invariant was right, the name for it was wrong.

Worth saying plainly: I have filed three issues this release about comments describing behaviour the code does not have (#717, #724, #726) and then shipped one in the same branch. It surfaced from asking what actually happens to an unlisted model instead of assuming the table was a gate — the same move that caught the ListModels mistake earlier in this PR.

98 tests pass across the affected files.

wu-changxing:
Third correction on this branch, and this one is the same bug I opened the PR to fix.

**My replacement list put two models in front of new users that they cannot call.** Verified against a freshly generated identity — which is what a new user is:

```
co/gemini-3.6-flash   ok        co/gpt-5             403 paid_account_required
co/gemini-3.5-flash   ok        co/o4-mini           403 paid_account_required
co/gemini-2.5-pro     ok        co/claude-sonnet-4   403 paid_account_required
co/gemini-2.5-flash   ok
```

> Your free $5 credits work with Google-routed models. Purchase credits to unlock all models

Free credits are exactly what the surrounding CLI message is handing out (100k tokens for a GitHub star), so the list has to be what a free account can call. The paid models are real and reachable, so they now get their own line instead of being dropped:

```
You can use ConnectOnion models with the 'co/' prefix:
  • co/gemini-3.6-flash
  • co/gemini-3.5-flash
  • co/gemini-2.5-pro
  • co/gemini-2.5-flash

co/gpt-5, co/o4-mini, co/claude-sonnet-4 need purchased credits: https://o.openonion.ai
```

### How I got it wrong

I verified with this machine's own account. It has a **$715.94** balance. `co/gpt-5` answered for me and 403s for every person that message is written for. Same shape as the ListModels mistake earlier in this PR: a check that passes because it was run from the wrong position.

The network test now authenticates a fresh identity and asserts **both** directions — the advertised names complete a call, and the paid ones are refused with `paid_account_required`. If free credits ever cover them, that second half fails and says to move them onto the main list.

This also revises what I said above about `co/claude-sonnet-4` being a 401 from an upstream credential: on a funded account it is a 401, on a free account the request is refused before it reaches Anthropic. oo-api#110 still stands for the 401 itself.

### The e2e suite that should have caught all of this

`tests/e2e/real_api/` named `gemini-3-pro-preview` in five places and the two retired 2.0-exp models in two more. Run against the live provider, five tests fail with the 404. It is in the repo, marked `real_api`, and deselected in CI — so nobody runs it. Updated to live models; 12 pass.

Local suite 5282 passed. Network suites: 17 real calls.

wu-changxing:
Fourth commit, from following the free-account discovery one step further.

**Picking a paid model on a free account is the most likely first failure anyone has, and it was the one status with no translation.** `_call` maps 402 → `InsufficientCreditsError` and 503 → `ProviderServiceError`, then logs and re-raises everything else. So this reached the user as a raw `openai.PermissionDeniedError` with the response JSON printed twice — once by the logger, once by the traceback:

```
[co] ○ gpt-5                                           1/100
APIStatusError: status=403, message=Error code: 403 - {'detail': {'error': 'paid_account_required', ...
RAISED openai.PermissionDeniedError
Error code: 403 - {'detail': {'error': 'paid_account_required', 'message': "Model 'gpt-5' uses a paid provider...
```

Two lines above that, the banner had just said `balance: $5.00` / `credits on me, go build —aaron` — the same $5 the error is about. And 402, the *rarer* case of a depleted balance, is the one that got a formatted message.

Now:

```
❌ 'gpt-5' needs purchased credits

Free credits cover Google-routed models. These work now:
   • co/gemini-3.6-flash
   • co/gemini-3.5-flash
   • co/gemini-2.5-pro
   • co/gemini-2.5-flash

💡 To use gpt-5:
   • Purchase credits: https://o.openonion.ai
   • Check balance: Run 'co status' in terminal
```

Keyed on the backend's `error='paid_account_required'`, **not on the status number** — a 403 can mean other things, and guessing from the number would tell a suspended account to go buy credits. Two tests cover that (`account_suspended` and a bodyless 403 both reach the caller untouched).

`FREE_MANAGED_MODELS` moves to `core.usage` beside the other model tables, so the CLI and the error render one tuple. My first version had `core/exceptions.py` importing from `cli/`, which inverts the layering — caught reviewing my own diff.

### The new fix broke one of my own tests, as expected

The previous commit's network test asserted `"paid_account_required" in str(error)`. That passed **only because the raw JSON was reaching the user** — precisely what this commit stops — so it directly contradicted the new test asserting the JSON is not dumped. It now asserts the exception type and `model_requested`.

Verified against the live backend with a freshly authenticated free account for both `co/gpt-5` and `co/o4-mini`. 5298 local, 13 network (run twice; one earlier failure was a transient upstream hiccup, not reproducible).

wu-changxing:
Two CI failures on this branch since the last comment, both mine, and the second one is worth recording as a general fact about this repo.

### #711 reproduced and fixed

`co browser go_to https://chat.openonion.ai` failed while doing the release e2e — every command, no recovery — while `co browser status` said `Browser: open`. `_context_is_alive()` asked the wrong question:

```python
# Alive = the context still answers protocol calls. Zero open pages is
# still alive (a page opens on demand), so listing pages IS the check.
list(self.browser.pages)
```

`pages` is a local list, not a protocol call. Measured with the browser process killed, which is how it happens in the field:

```
is_closed()      ->  False   a flag set by close(), and close() never ran
len(ctx.pages)   ->  1       the local list still holds the dead tab
ctx.cookies()    ->  TargetClosedError
ctx.new_page()   ->  TargetClosedError   <- what the user hits
```

**My first fix used `is_closed()` alone and failed the same real replay it was meant to fix.** Third time on that issue a local attribute was mistaken for asking the browser — the attempt before mine checked `context.browser`, which is None for `launch_persistent_context`. So the tests now open a real browser and SIGKILL it.

Before/after, same sequence: `status` goes `open` → `not open`, and the next command goes `TargetClosedError` → `Navigated to https://chat.openonion.ai/`.

### typer vendors its own Click — worth knowing beyond this PR

My typo tests failed in CI and passed locally:

```
typer 0.20.0   groups found: 12   nested contact: True
typer 0.27.1   groups found:  0   nested contact: False
```

typer 0.27's group class is `typer._click.core.Command`, so `isinstance(group, click.Group)` against the **installed** click is False. `pyproject` asks for `typer>=0.20.0`; this machine had 0.20.0, CI resolves 0.27.1.

The same assumption was inside the fix: `except click.UsageError` never matches on 0.27, so the handler was **inert in exactly the version CI runs** — and passed there only because 0.27 has no doubling to fix. It now catches broadly and acts only on an object carrying the attributes it uses.

### What I changed about how I verify

Three CI failures this release came from checking against one local version. There is now a second environment pinned to CI's typer, and **the full suite passes on both: 5348 under typer 0.20.0 and 5348 under 0.27.1.** I ran it there before pushing rather than after.

Also in these commits: a colon in a SKILL.md description no longer hides the skill from the model (eight installed skills were invisible), and bare `co` no longer claims "Commands:" while omitting eight real ones.

wu-changxing:
Round with no code change — two leads investigated, both turned out to be correct behaviour, and one real-machine e2e completed. Recording it because "checked and fine" is worth as much as a fix when the question was whether to ship.

### ANNOUNCE message size — not a bug

`co announce` inlines every published SKILL.md body into one WebSocket frame. On this machine that is **254 KB across 19 skills**, and `websockets` defaults `max_size` to 1 MB, so it looked like a ceiling a few more skills would hit silently.

Measured against the live relay:

```
  262,586 B  -> ANNOUNCE_OK
1,573,306 B  -> ANNOUNCE_OK
8,389,050 B  -> ANNOUNCE_OK
```

The 1 MB default governs what a client will *receive*, not what it sends, and `co sub sync` reads profiles over HTTP rather than the socket. No limit to add.

**I did break something doing this and fixed it:** those probes published a `sizetest` profile with an 8 MB junk body over my real one on the relay. Re-announced and verified — alias `changxing`, 39 skills, no `sizetest` left. Probing shared state with real writes was the wrong way to answer the question; the first two sizes were enough.

### doctor's `broken symlink` — reporting correctly

`skill user/email-outreach: broken symlink` points at a landpages directory that no longer exists. I created that link by hand in July; the fanout did not. doctor is right and there is nothing to change.

### Real-machine nw-e2e — passes end to end

```
naturewill-mapping.service     loaded active running
local /health                  200
NRestarts                      0
0.0.0.0:8000                   python pid 61333
https://nw-map-10e68f6d.agents.openonion.ai/health   200
```

`NRestarts 0` matters more than `active running` — a crash loop shows as active too. And the bare port 8000 timing out from outside, which I read as a fault last round, is the firewall doing its job: 443 fronts it and the hostname answers.

Branch state: 23 commits, 18 shipped files, +787/-120, CI green on every completed run.

wu-changxing:
Another round with nothing to fix, and two wrong measurements of my own worth recording — both the same mistake I had just written a commit about.

### The 62k prompt claims to list every skill. It does.

`_inject_skills_to_system_prompt` tells the model "the list below is the whole set", which is the kind of absolute claim this release has repeatedly found to be false. Measured:

```
session system message : 89,076 chars
skills discovered      : 118
rendered lines         : 118
discovered but missing : 0
```

No truncation, no cap. The claim holds.

Also checked and clean: the assembled prompt names only `co/gemini-3.6-flash` — no retired model reaches it — and no paragraph over 120 characters appears twice (0 duplicated characters of 62,737), so nothing is being paid for at every request.

And the frontmatter fix earlier in this branch shows up where it matters: the three skills whose YAML does not parse now carry descriptions into the real prompt rather than arriving blank.

### The two things I measured wrong

**I read `agent.system_prompt` and got 0 rendered skill lines**, then nearly reported "118 discovered, none listed". The injection edits the *session's* system message — `agent.current_session['messages']` — not the attribute. The code says so plainly four lines from where I was looking.

**Before that I matched skill names with a guessed pattern** and got 68 of 118, which reads like a cap. Counting the rendered `- `/name`` lines instead of pattern-matching names gave 118. A guessed pattern that under-matches looks exactly like a real truncation bug.

Last commit's lesson was "ask the loader, not the file". This round: ask the object the code writes to, not the one with the similar name. Same failure, one layer over.

Branch: 33 commits, CI green on every completed run, 5646 local tests.

PR #739, opened 2026-08-08
fix: record token usage from structured_complete (#730)

### LINKED ISSUE #730: A structured call spends money that nothing records
`structured_complete()` returns the parsed model and records no usage. No provider's implementation does — all eight `TokenUsage(...)` constructions in `core/llm.py` are in `complete()`. So a structured call contributes zero tokens and zero cost to the session, the run summary, and the saved eval, while the account is really debited.

## Measured

Against the real backend on `co/gemini-3.6-flash`:

```
structured result: value=391
balance before/after: 690.4215 690.4208   delta: 0.0007
```

$0.0007 spent, nothing recorded anywhere. `OpenOnionLLM.structured_complete` is:

```python
completion = self._call(lambda: self.client.beta.chat.completions.parse(...))
return completion.choices[0].message.parsed
```

`completion.usage` is right there, including the `cost_usd` the server states — the same field `complete()` reads and trusts — and it is discarded with the rest of the response.

## Why it matters

`llm_do(..., output=Schema)` is one of the two public entry points and takes this path. So does anything asking for structured output inside an agent. The visible consequences:

- the run summary undercounts a session that used structured output, by the whole of it
- a saved eval records a run as costing less than it did
- reconciling `co status` balance against reported cost shows a gap with no explanation

This is the same shape as the `0 tokens · $0.0000` summary and the token count that could not be reconciled with its own cost (both fixed in #725): a number that reads as authoritative and is not.

## Why I am not just fixing it

The fix changes a public signature. `structured_complete()` returns a `BaseModel` — the schema instance — so there is nowhere to put the usage without either changing what it returns (breaking every caller and the `LLM` ABC) or holding it as state on the LLM instance for the caller to read afterwards. On an LTS that is a shape decision, not a repair. Options as I see them:

1. Return `(instance, usage)` — honest, breaks the ABC and every call site.
2. Record `self.last_usage` on the LLM and have `llm_do`/the agent read it after the call. No signature change; introduces per-instance state that is wrong under any concurrent use.
3. Have `structured_complete` route through `complete()` with a tool-shaped schema so there is one accounting path. Biggest change, removes the duplication that caused this.

Option 3 is the one that stops it recurring — eight construction sites is why the fix landed in one of them — but it is also the one that touches every provider.

Found while auditing the release's own cost reporting; the eight-site duplication is the same pattern as the `_replace`/`signature_already_used` findings in #725.

kimberlyscottbase:
## Summary
`structured_complete()` was spending tokens/cost and discarding `completion.usage`. This implements option 2 from #730: keep the `BaseModel` return type, store `TokenUsage` on `llm.last_usage` (and `llm_do.last_usage` for the one-shot helper).

- Shared `_usage_from_openai_chat()` so OpenOnion `complete` / `structured_complete` read the same `cost_usd` / `total_tokens` fields
- All eight providers set `last_usage` after a structured call
- Unit test covers the measured OpenOnion case (`cost_usd` + server `total_tokens`)

Fixes #730

## Test plan
- [x] Mocked `OpenOnionLLM.structured_complete` asserts `last_usage.cost` / `total_tokens`
- [ ] `pytest tests/unit/test_structured_complete_usage.py`
- [ ] Optional: live `co/` structured call and confirm `llm.last_usage.cost` matches balance delta

### DIFF
--- a/connectonion/core/llm.py
+++ b/connectonion/core/llm.py
@@ -206,6 +206,52 @@ def _is_paid_account_required(error) -> bool:
     return isinstance(detail, dict) and detail.get('error') == 'paid_account_required'
 
 
+
+def _usage_from_openai_chat(
+    model: str,
+    response,
+    *,
+    trust_server_cost: bool = False,
+) -> Optional[TokenUsage]:
+    """Build TokenUsage from an OpenAI-shaped chat.completions response.
+
+    Used by complete() and structured_complete() so a structured call cannot
+    silently drop the same usage fields complete() already knows how to read
+    (#730). trust_server_cost=True is for OpenOnion, where the proxy states
+    cost_usd / total_tokens that local arithmetic cannot reconstruct.
+    """
+    if not (hasattr(response, "usage") and response.usage):
+        return None
+
+    input_tokens = response.usage.prompt_tokens
+    output_tokens = response.usage.completion_tokens
+    cached_tokens = 0
+    if hasattr(response.usage, "prompt_tokens_details") and response.usage.prompt_tokens_details:
+        cached_tokens = getattr(response.usage.prompt_tokens_details, "cached_tokens", 0) or 0
+
+    cost = None
+    total_tokens = 0
+    if trust_server_cost:
+        # `is not None` rather than a truth test — a free call reports 0.0,
+        # and falling back there would invent a charge for it.
+        cost = getattr(response.usage, "cost_usd", None)
+        server_total = getattr(response.usage, "total_tokens", 0) or 0
+        # Only when it exceeds the sum. Equal or smaller means the provider is
+        # restating those two fields and there is nothing extra to report.
+        if server_total > input_tokens + output_tokens:
+            total_tokens = server_total
+    if cost is None:
+        cost = calculate_cost(model, input_tokens, output_tokens, cached_tokens)
+
+    return TokenUsage(
+        input_tokens=input_tokens,
+        output_tokens=output_tokens,
+        cached_tokens=cached_tokens,
+        cost=cost,
+        total_tokens=total_tokens,
+    )
+
+
 @dataclass
 class LLMResponse:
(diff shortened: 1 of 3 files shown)


wu-changxing (changes requested):
Blocking for 1.6.0. The focused new tests pass, but the existing provider/llm_do suite regresses:

```
pytest -q tests/unit/test_structured_complete_usage.py tests/unit/test_structured_complete_errors.py tests/unit/test_llm_do.py tests/unit/test_openonion_llm.py tests/unit/test_anthropic_llm.py tests/unit/test_groq_llm.py tests/unit/test_llm.py tests/unit/test_usage.py
# 14 failed, 72 passed
```

Concrete failures:

- `llm_do` now reads `response.usage` unconditionally; valid existing response objects without that attribute raise `AttributeError`.
- Groq structured responses whose `usage` is absent/a test double flow into arithmetic and raise `TypeError`.
- Anthropic structured responses without usage raise `AttributeError`.
- `last_usage` is not cleared at call start, so a failed or no-usage second call can leave the previous successful call's cost attached and cause mis-accounting.
- `llm_do.last_usage` is a process-global function attribute. Concurrent calls overwrite one another, which reintroduces the attribution problem that the PR comments already acknowledge for shared LLM instances.

Please make usage optional/defensive, clear stale state before each attempt, avoid the global `llm_do` side channel (or document and synchronize a deliberate design), add regressions for no-usage/failure/concurrency, and run the existing related suites above.

wu-changxing:
Closing this implementation, not the underlying issue. The focused test passes, but the existing provider/llm_do suite has 14 regressions: missing usage objects raise, stale usage survives failures, and the process-global llm_do.last_usage side channel misattributes concurrent calls. Correcting that safely requires a deliberate cross-provider/public-API accounting design, not a release-candidate patch. #730 remains open and is being moved to the observability milestone. This PR is excluded from 1.6.0.

PR #741, opened 2026-08-08
Route every managed client through one backend

### LINKED ISSUE #733: CONNECTONION_BACKEND_URL redirects three modules out of nineteen, silently
`CONNECTONION_BACKEND_URL` looks like the way to point the CLI at a different backend. It reaches three modules:

```
honours CONNECTONION_BACKEND_URL   (3)
  cli/commands/email_commands.py
  useful_tools/get_emails.py
  useful_tools/send_email.py

hardcodes https://oo.openonion.ai and ignores it   (16)
  cli/commands/auth_commands.py      core/llm.py
  cli/commands/deploy_commands.py    derive.py
  cli/commands/doctor_commands.py    network/trust/trust_agent.py
  cli/commands/server_commands.py    transcribe.py
  cli/commands/status_commands.py    useful_plugins/image_result_formatter.py
  cli/commands/sub_commands.py       useful_tools/gdrive.py
                                     useful_tools/gmail.py
                                     useful_tools/google_calendar.py
                                     useful_tools/microsoft_calendar.py
                                     useful_tools/outlook.py
```

19 modules hardcode the production host. So setting the variable gives you email against your target and **auth, model calls, deploy, announce, trust, doctor, status, subscriptions, Drive, Gmail, Outlook and Calendar against production** — split-brain, with nothing saying so.

## How I found it

Not by reading. I set `CONNECTONION_BACKEND_URL=http://127.0.0.1:9` (a closed port) and ran `co init --yes` in a throwaway directory with a throwaway `HOME`, expecting authentication to fail so I could see what `co init` does when it does. It printed:

```
✓ Authenticated (Balance: $5.00)
```

$5.00 is a new account's free credit. The auth path ignored the variable and **created a real account on production** — which is the failure mode in one line: someone pointing the CLI at staging, or at a local backend, gets production side effects they specifically tried to avoid. (Throwaway directory deleted; the account is an unused one with free credit.)

## Why I am not fixing it here

The fix is a single resolver — one `backend_url()` that every call site reads — and it is mechanical across 19 modules. It is also the kind of change where getting one site wrong sends a request to the wrong host, and PR #725's CI has not been able to run for hours (GitHub Actions outage), so a 19-module network-behaviour change would land verified only by my local runs. Not on the eve of an LTS.

## Options

1. One `backend_url()` resolver, every site reads it. Correct, and the only version that makes the variable mean what it looks like.
2. Rename it to `CONNECTONION_EMAIL_BACKEND_URL` so it stops promising more than it does. Small, honest, and leaves no way to redirect the rest.
3. Leave it and document the three. Cheapest, and keeps a trap that already caught me.

(1) is what I would do, in its own PR with CI green on it.

## Adjacent, found in the same probe

`co init` discards the result of `authenticate()` — `auth_success = authenticate(...)` is assigned and never read (pyflakes flags it), so init prints its success banner and setup guidance regardless. Whether that is right is a product question: the project files ARE valid without a key and `co auth` can be run later. But nothing in the final banner distinguishes the two outcomes.

Also `global_dir.mkdir(exist_ok=True)` in the same path lacks `parents=True`, so a HOME that does not exist yet (containers, sandboxes) gives a raw `FileNotFoundError` traceback instead of a message.

wu-changxing:
## Summary

- add one runtime `backend_url()` resolver; `CONNECTONION_BACKEND_URL` is canonical and legacy URL variables are compatibility aliases
- route auth, managed models, deploy/server, doctor/status, email, OAuth-backed tools, images, trust, transcription, announce, subscriptions, remote-agent connections, relay clients, and host relay defaults through it
- migrate old generated `host.yaml` production-relay defaults at runtime while preserving genuinely custom relay overrides and explicit relay disablement
- make the packaged host template omit a fixed relay and make the built-in topup skill use the selected backend
- keep the SLIP-0013 `ACCOUNT_URI` fixed because it is a cryptographic domain separator, not a network destination
- make `co doctor` report an unreachable configured backend instead of crashing
- add a source-level guard so no executable Python/YAML/skill network client can reintroduce a production-origin literal
- document precedence and scope

Closes #733

## Verification

- original 380 targeted cross-module unit tests passed
- original 124 CLI E2E tests passed
- follow-up HTTP/WebSocket/host/relay/email integration audit: **139 passed** on the stacked 1.6 candidate
- focused PR-branch routing/template follow-up: **132 passed**
- final stacked 1.6 candidate: **6047 passed, 18 skipped, 174 deselected, 0 failed** (`--randomly-seed=3447603744`)
- GitHub Actions: **8/8 checks passed** on the final PR head
- full unit audit previously reached 1735 passed before stopping after sandbox-only failures (Unix socket creation and writes to the real `~/.co` are denied here); every related failure was rerun in isolation and passes
- `python -m compileall -q connectonion`
- `git diff --check`

## Review focus

This changes routing behavior across the client. Please specifically verify staging/local auth, one `co/*` call, deploy/server dry paths, email, OAuth refresh, `connect()`, `co call`, host startup with an old generated `relay_url`, announce/sub, and a custom private relay override before marking ready.


wu-changxing:
Release audit found and fixed a remaining split-brain path in `aeb0840`.

The earlier branch unified HTTP and announce, but default WebSocket paths still bypassed `CONNECTONION_BACKEND_URL`: public `connect()`, `co call`, `network.relay.connect()`, and `host()` could still talk to production while auth/models/email talked to staging. Existing generated `host.yaml` files also contained an explicit production relay, and the built-in topup skill curled production directly.

The follow-up now:

- derives all omitted relay defaults from the shared backend selector,
- treats the legacy generated production relay value as a default while preserving custom/private relay overrides and explicit disablement,
- removes the fixed relay from newly generated host config,
- routes the packaged topup skill through `backend_url()`, and
- expands the source guard to executable Python, YAML, and packaged skills.

Evidence: focused branch suite **111/111**; stacked candidate HTTP/WebSocket/host/relay/email suite **139/139**; compile and diff checks pass. GitHub CI is rerunning on the new head. No merge or release performed.

wu-changxing:
Final follow-up evidence on `5e8184a`:

- GitHub Actions: **8/8 passed** (Python 3.10–3.13 plus Windows browser/E2E)
- PR merge state: **CLEAN**
- focused PR-branch routing/template suite: **132 passed**
- stacked routing/email suite: **139 passed**
- final stacked 1.6 candidate: **6047 passed, 18 skipped, 174 deselected, 0 failed** (`--randomly-seed=3447603744`, 516.28s)

No merge performed; this is ready for the requested independent review. Because this routing foundation overlaps the later Connect candidates, it should be merged first and those PRs rebased afterward while preserving both behavior sets.

wu-changxing:

PR #742, opened 2026-08-08
security: bind admin signatures to request routes

### LINKED ISSUE #729: An admin signature is not bound to the route it authorises
The admin routes verify a signature over `{timestamp}` alone:

```python
data = {"payload": {"timestamp": float(header_ts)}, "from": header_from, "signature": header_sig}
```

So the signature says "I am an admin, some time in the last five minutes" and nothing about **what** it authorises. Replay is now refused (#725 calls `signature_already_used()`, the record CONNECT uses for #649), so one signature is one call — but which call is still not fixed by the signature.

Measured before the replay fix, one signature over `{timestamp}`:

```
1st use on /admin/logs                       200
SAME signature reused on /admin/sessions     200
SAME signature replayed on /admin/logs       200
```

The middle line is what remains in principle: a signature captured in flight, if it has not been spent, can be spent on a route the signer did not choose — `/admin/trust/block` rather than `/admin/logs`.

## The answer already exists in this repo

`network/host/auth.py` defines a signed-GET scheme for exactly this and explains itself above the header names:

```
# The signature travels in headers over {method, path, timestamp}, canonicalised
# exactly as CONNECT and INPUT already are, and is then verified by the same
# _authenticate_signed below. One canonicalisation, one freshness window, one
# blacklist check -- a second implementation of any of those is how the halves
# of a gate drift apart, which is most of what this release has been about.
```

It signs `{method, path, timestamp}` and uses `x-co-from` / `x-co-signature` / `x-co-timestamp`. The admin routes are the second implementation that comment warns about: different header names, weaker payload.

## Why it is filed rather than fixed

Adding method and path to the signed payload changes what a client must sign. Anything already signing admin calls — the dashboard, scripts, `co` itself — breaks until it signs the new shape, and 1.6.0 is a long-term release. That is a coordinated change with a compatibility window, not a line to slip into a bug-fix branch.

Sketch, if it helps whoever takes it: accept both shapes for one release, prefer `x-co-*` when present, warn when the `{timestamp}`-only form is used, then require the bound form in the next major. The verification itself would come from `auth.py`'s `verify_signed_request` rather than a third copy.

Found doing #670 in #725 — widening the signed path to `/admin/logs` and `/admin/sessions` is what made me look at what that signature actually covers.

wu-changxing:
Closes #729.

## What changed

- bind signed admin requests to the actual HTTP method and path
- add `sign_request_body()` for POST admin clients
- use the existing `x-co-*` signed-GET scheme instead of a second header format
- preserve legacy Bearer access and legacy signed reads for `/admin/logs` and `/admin/sessions`
- refuse unbound signatures on trust and super-admin mutations
- allow the `x-co-*` headers through CORS
- document the route-bound payload format

## Security property

A signature created for `POST /admin/trust/promote` cannot be submitted to
`POST /admin/trust/block`, and a GET signature cannot be moved to another
admin route. Legacy unbound signatures remain usable only on the two read-only
activity routes, so they cannot be redirected into a mutation.

## Verification

- 156 focused auth/ASGI/host tests passed
- compileall passed
- diff check passed
- broader unit run reached 1,606 passed before being stopped; its failures were existing sandbox-dependent socket/browser/event-loop tests, not touched paths

Ready for external security review before 1.6.0. Do not merge or release yet.

### DIFF
--- a/connectonion/network/asgi/http.py
+++ b/connectonion/network/asgi/http.py
@@ -51,7 +51,10 @@ def pydantic_json_encoder(obj):
 CORS_HEADERS = [
     [b"access-control-allow-origin", b"*"],
     [b"access-control-allow-methods", b"GET, POST, OPTIONS"],
-    [b"access-control-allow-headers", b"authorization, content-type"],
+    [
+        b"access-control-allow-headers",
+        b"authorization, content-type, x-co-from, x-co-signature, x-co-timestamp",
+    ],
 ]
 
 
--- a/connectonion/network/host/auth.py
+++ b/connectonion/network/host/auth.py
@@ -190,6 +190,29 @@ def sign_request(keys: dict, method: str, path: str, timestamp=None) -> dict:
     }
 
 
+def sign_request_body(keys: dict, method: str, path: str, payload=None,
+                      timestamp=None) -> dict:
+    """Build a signed JSON request whose authority is bound to its route.
+
+    POST admin calls carry arguments as well as the method and path. Keeping
+    all of them in one signed payload prevents a captured promote request from
+    being submitted to block, or to any other admin endpoint.
+    """
+    from ... import address as _address
+
+    signed_payload = dict(payload or {})
+    signed_payload.update(
+        method=method.upper(),
+        path=path,
+        timestamp=int(timestamp if timestamp is not None else time.time()),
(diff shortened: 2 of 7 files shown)


wu-changxing:

PR #743, opened 2026-08-08
security: verify subscribed skills against publisher signatures

### LINKED ISSUE #654: co sub sync installs unverified skills — and a skill's frontmatter grants permissions
`sub_commands.py` states the trust assumption in its own header:

> ⚠️ v1 trusts the relay — no Ed25519 signature verification (relay strips signer/signature from profile responses).

That is an honest limitation, recorded by whoever built it. Two things make it worth revisiting before a long-term release.

## 1. A skill is not only instructions — it carries permission grants

`SKILL.md` frontmatter has a `tools:` list, and invoking the skill auto-approves those patterns for the turn. Measured earlier this session on a real agent:

```
during the turn : ['Bash(git status)', 'read_file']
```

So a synced skill does not merely suggest what an agent should do; it arrives with a request for auto-approval attached, from a source that was not verified.

## 2. It is written verbatim, and fanned out to five other tools

```python
(skills_root / name / "SKILL.md").write_text(body, encoding="utf-8")
```

then `fanout.install_all()` symlinks or copies it into `~/.claude`, `~/.codex`, `~/.openclaw`, `~/.cursor` and `~/.kiro`. One `co sub sync` puts unverified content into five agents' skill directories, four of which are not ours.

## 3. The caveat is developer-facing only

The warnings are at lines 23–24 of the module docstring. `co sub sync --help` says:

```
Sync one publisher: fetch profile, mirror skills, fan out to coding agents.
```

Nothing about verification, nothing about where the files land. The person taking the risk is the one who does not see it.

(The sibling caveat on line 24 — aliases are mutable, so a first-time sub requires a `0x` address — *is* enforced in code. This one is not surfaced at all.)

## Options

1. **Say it at the point of risk.** Print what is about to be installed and where, and that it is unverified, before writing. Cheapest, no protocol work.
2. **Strip or quarantine `tools:` on sync.** A subscribed skill keeps its instructions and loses its ability to pre-authorise anything; the operator adds patterns themselves if they want them.
3. **Verify the publisher's signature**, which is the real answer and needs the relay to stop stripping it — the docstring says it currently does.

(2) is the one I would take first: it removes the part that acts on its own, and does not need the relay to change. (1) costs nothing and should happen regardless.

Related to #653 in kind rather than in mechanism: both are about permissions arriving from somewhere other than the operator.

wu-changxing:
Closes #654 once the backend companion is deployed.

Backend companion: openonion/oo-api#120.

## What changed

- add a portable Ed25519 profile signature to ANNOUNCE
- fetch all public skill bodies and reconstruct the exact signed profile
- verify against the pinned 0x publisher address before any filesystem write
- refuse unsigned, mismatched, or relay-tampered profiles
- refuse signed alias/skill path traversal names
- bind each subscribed alias to its pinned publisher address so a second valid signer cannot overwrite another publisher's local bundle
- keep stripping remote tools grants so authenticated content does not import authority
- document the verification and legacy-profile behavior

## Security property

A relay-modified alias, skill list, description, or body fails verification before `~/.co/subs`, `subscriptions.txt`, or any coding-agent directory is touched. A separately signed publisher also cannot claim an alias already bound locally to another address. Unsigned legacy profiles stay discoverable but cannot be installed.

The v1 attestation proves authenticity, not freshness: a relay replaying an older still-valid signed bundle is a separate monotonic-version protocol decision and should remain visible to external review rather than being improvised into this release.

## Validation

- signature + subscription focused regression after alias binding: 20 passed
- broader subscription/grant focused suite at PR head: 40 passed
- previous protocol/documentation and CLI entrypoint checks remain covered
- `git diff --check`
- all 8 GitHub jobs green on `d90e54e`

## Rollout

Deploy openonion/oo-api#120 first, then release this client. Existing publishers must re-announce with 1.6.0 to become installable under the new gate.

Ready for external security review. Do not merge or release yet.

### DIFF
--- a/connectonion/cli/commands/sub_commands.py
+++ b/connectonion/cli/commands/sub_commands.py
@@ -3,7 +3,7 @@ Purpose: `co sub` — record a subscription relationship to a published agent, m
 LLM-Note:
   Dependencies: imports from [json, re, shutil, pathlib, httpx, rich.console, rich.table, .fanout] | imported by [cli/main.py via handle_sub_sync_one/sync_all/list/remove] | tested by [tests/unit/test_sub_commands.py, tests/cli/test_cli_sub.py]
   Data flow:
-    sync_one(target, relay?) → _resolve_target() validates 0x address or matches an alias already in subscriptions.txt → _fetch_profile() GET /api/agents/<addr>/profile → _mirror_bundle() writes ~/.co/subs/<alias>/agent.json + for each skill GET /api/agents/<addr>/skills/<name> → unwrap body (handles both JSON {body:...} and raw text/markdown content-types) → write SKILL.md → append `<address> <alias>` to ~/.co/subscriptions.txt (deduped) → fanout.install_all() symlinks/copies into ~/.claude, ~/.codex, ~/.openclaw, ~/.cursor, ~/.kiro
+    sync_one(target, relay?) → validate/pin 0x publisher address → GET signed profile metadata + every published body → reconstruct the exact profile and verify its Ed25519 profile-v1 signature before filesystem writes → strip remote tools grants → mirror under ~/.co/subs/<alias>/ → record subscription → fan out to coding agents
     sync_all(relay?) → walks ~/.co/subscriptions.txt, calls sync_one per entry, tolerates per-publisher failures, prints summary (ok/failed counts)
     list() → reads ~/.co/subscriptions.txt + ~/.co/subs/<alias>/agent.json → Rich table with alias, full address, version, skill count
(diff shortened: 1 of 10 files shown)


wu-changxing:
Security follow-up pushed in `d90e54e`:

- found that a second independently valid signer could choose an alias already subscribed locally and overwrite the first publisher's `~/.co/subs/<alias>` bundle
- bind an alias to its already pinned publisher address and reject a conflicting signer before any filesystem write
- red regression proved the old overwrite and the focused suites are now green (20/20 integration; 40/40 broader subscription/grant suite)

Freshness is deliberately not overstated: profile-v1 authenticates content but does not prevent replay of an older valid signed bundle. That separate design decision is now tracked in #768 for external release review. No merge or release performed.

wu-changxing:

Issue #180, opened 2026-07-06
Gmail auth via oo-api stops working after ~1–2 hours — backend must own token refresh; frontend should drive the connect flow


wu-changxing:
## Goal

Gmail authentication through the OpenOnion backend (oo.openonion.ai) must work end-to-end and must **not** be limited to ~1–2 hours. Users connect via the frontend (o.openonion.ai), and everything downstream (CLI, agents, deployed agents) keeps working indefinitely via the stored refresh token.

## What already works (verified against production, 2026-07-06)

- `GET /api/v1/oauth/google/init` returns a valid Google auth URL with `access_type=offline&prompt=consent` and Gmail + Calendar scopes (HTTP 200 verified live).
- Redirect URI points at the backend `GET /api/v1/oauth/google/callback`, which exchanges the code, saves `access_token` + `refresh_token` + `expires_at` to the DB, then redirects to the frontend success page.
- Frontend has a Dashboard "Link Google" button (popup + `postMessage`) and `app/oauth/google/callback/page.tsx`.
- `POST /api/v1/oauth/google/refresh` proxy exists (client sends refresh_token, backend uses its client secret).
- Python `Gmail` / `GoogleCalendar` tools refresh via that endpoint when the local token is within 5 min of expiry.
- The OPENONION_API_KEY JWT is long-lived (verified: ~10 years), so refresh calls are not blocked by JWT expiry.

## Why auth still dies after ~1–2 hours

1. **The backend never refreshes its own stored token.** Google access tokens live 1 hour. `GET /api/v1/oauth/google/credentials` (oo-api `oauth/routes.py`) returns the DB row as-is — one hour after connecting, it hands out a dead `access_token`. Every consumer is forced to implement its own refresh loop. The Python Gmail/Calendar tools do; anything that doesn't (fresh `co auth google` pull, gmail_plugin, connectonion-ts / oo-chat, deployed agents with env baked at deploy time) breaks as soon as the first hour passes.
2. **Refresh-token durability depends on the Google Cloud consent screen publishing status.** If the OAuth app (client `941600173705-...`) is still in *Testing*, Google expires refresh tokens after 7 days and caps test users. It must be Published (and verified for the restricted `gmail.*` scopes) for auth to be truly unlimited.
3. **Client-side refresh writes back to `~/.co/keys.env` only if the file already exists**, and deployed agents can't update their baked-in env at all — so even where refresh works, the new token isn't always persisted.

## Proposed fix

[...]

Issue #121, opened 2026-03-23
Replace plan mode with TodoList in co ai workflow


wu-changxing:
## Problem

The current `co ai` agent workflow forces plan mode (`enter_plan_mode` → `write_plan` → `exit_plan_and_implement`) every time a user wants to create an agent. This is over-engineered for simple agents where the tools are obvious.

Plan mode has dedicated tools, READ-ONLY restrictions, a YAML spec format, and a rigid enter/exit workflow — all for something that a TodoList can handle more naturally.

## Proposal

Replace plan mode with the existing `TodoList` tool for agent creation workflow:

- **Simple agents**: `co create` → edit `agent.py` directly. No planning needed.
- **Complex agents**: Use `todo` to break down the work into steps, then execute them. The user can see progress and provide feedback naturally.

TodoList is more flexible — it doesn't lock the agent into a read-only mode, doesn't require a rigid YAML format, and lets the agent work incrementally.

## Changes needed

- Remove `enter_plan_mode`, `exit_plan_and_implement`, `write_plan` tools from co ai
- Remove `tools/enter_plan_mode.md` prompt
- Remove `system-reminders/plan_mode.md` reminder
- Update `workflow.md` to use TodoList-based approach
- Update `main.md` references to plan mode
- Remove plan mode plugin from agent.py tool list

## Context

Part of the system prompt simplification effort — the base prompt was already reduced from ~25,800 to ~12,186 chars by making workflow/index/examples load on-demand via intent detection.

wu-changxing:
## 🔍 Summary

Replace the rigid plan mode workflow (`enter_plan_mode` → `write_plan` → `exit_plan_and_implement`) with the existing `TodoList` tool. Plan mode forces READ-ONLY restrictions, a YAML spec format, and a dedicated enter/exit ceremony — all of which `TodoList` handles more naturally and flexibly. This removes 3 tools, 2 prompt files, 1 system reminder, and simplifies the agent creation workflow.

## 📊 Workflow: Current vs Proposed

```
CURRENT (Over-engineered):
                                                       
User: "build me an inbox agent"                        
       │                                               
       ▼                                               
┌──────────────────┐                                   
│  detect_intent() │ → is_build: true                  
└────────┬─────────┘                                   
         │ loads workflow.md + agent.md reminder        
         ▼                                             
┌──────────────────┐   ┌─────────────────────────┐     
│ enter_plan_mode()│──▶│ session.mode = 'plan'   │     
[...]

PR #744, opened 2026-08-08
Recover Google tool sessions after token expiry

wu-changxing:
Client half of #180. Depends on openonion/oo-api#121 being deployed first.

## What changed

- Gmail and Google Calendar get their first token from the authenticated backend account without depending on a locally baked expiry or refresh token
- google-auth now has a backend refresh handler, so cached Gmail and Calendar services recover and retry after an access-token 401
- neither tool sends its locally stored refresh token back to the backend; the authenticated backend account owns refresh
- rotated refresh tokens are persisted when the backend returns one
- `reauth_required` becomes an actionable `co auth google` message without exposing provider response bodies
- backend calls have an explicit 15-second timeout
- `co auth google` no longer deletes the existing database credential before opening the browser; it watches for a changed expiry instead, so running deployments keep working during reauthorization
- Google integration docs no longer tell users to construct `Credentials` with client secrets they never receive

## Verification

- combined Gmail + Google Calendar + CLI auth regression: 98 passed
- `git diff --check`
- all 8 GitHub jobs green on `361d280`

## Release note

Ready for external review. No merge, deployment, or release performed. Rollout order is oo-api #121 first, then this client change.

### DIFF
--- a/connectonion/cli/commands/auth_commands.py
+++ b/connectonion/cli/commands/auth_commands.py
@@ -244,9 +244,16 @@ def handle_google_auth():
     api_url = "https://oo.openonion.ai/api/v1/oauth"
     headers = {"Authorization": f"Bearer {api_key}"}
 
-    # Clear any existing connection first - this ensures we wait for NEW OAuth to complete
-    # (otherwise /google/status returns connected=true immediately from old credentials)
-    requests.delete(f"{api_url}/google/revoke", headers=headers)
+    # Keep the existing refresh token alive while re-authenticating. Deleting it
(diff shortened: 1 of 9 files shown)


wu-changxing:
Scope-completeness follow-up pushed in `74815c1`: the original #180 acceptance text names both Gmail and Google Calendar, but this PR previously hardened only Gmail. Calendar now uses the same server-owned refresh flow, cached-service 401 handler, rotated-token persistence, stable redacted errors, and request timeout. Combined focused regression: 97/97. No merge or release performed.

wu-changxing:

Issue #731, opened 2026-08-06
gemini-3.6-flash's cached price is 10% of input, where every other Gemini row is 25%


wu-changxing:
`MODEL_PRICING` states the rule two lines above the data:

```python
# Google Gemini models - cached = 25% of input (75% discount)
```

Every row follows it except one:

| model | input | cached | ratio |
|---|---|---|---|
| **gemini-3.6-flash** | 1.50 | **0.15** | **10%** |
| gemini-3.5-flash | 1.50 | 0.375 | 25% |
| gemini-3-pro-preview | 2.00 | 0.50 | 25% |
| gemini-2.5-pro | 1.25 | 0.3125 | 25% |
| gemini-2.5-flash | 0.15 | 0.0375 | 25% |
| gemini-2.0-flash | 0.10 | 0.025 | 25% |

`0.15` is exactly `gemini-2.5-flash`'s **input** price, one line below — the shape of a copy that was not adjusted when the input price went from 0.15 to 1.50. It was introduced with that value in 01dba93 ("Add Gemini 3.6 Flash model support", #220), so it has not drifted since.

## The input price is not in doubt

Reconciled against what the backend actually charged, on two calls:

```
in=   4  total=  75   charged=$0.000539  predicted=$0.000539  ratio 1.0009
in=2006  total=2092   charged=$0.003654  predicted=$0.003654  ratio 1.0000
```

where `predicted = input_tokens x $1.50/1M + (total - input) x $7.50/1M`. So 1.50 and 7.50 are exactly right, and the billing model is: input at the input rate, every remaining token — completion and reasoning both — at the output rate. That leaves `cached` at 10% of a **confirmed** input price, against a rule the file states for itself.

## Why I did not just change it

Writing 0.375 would be inventing a price, which is the thing #725 exists to remove. What is needed is Google's published context-caching rate for this model. If it really is 10%, the comment needs the exception; if it is 25%, the row needs 0.375.

## Blast radius

Small, and worth stating so this is not treated as urgent:

- The `co/` path never reaches it. The server states `cost_usd` and the client uses that; `cached_tokens` also came back **0** from this backend even on a deliberately repeated 6k-token prefix.
- It applies to a **direct** Gemini API call, where `calculate_cost` uses this table and Gemini does report cached tokens for context caching.

So the effect is under-reporting cost on direct-API context-cached calls, by 15% of the input rate on the cached portion.

## What landed meanwhile (#725)

[...]

PR #745, opened 2026-08-08
Document Gemini 3.6 cached input pricing

wu-changxing:
Resolves the uncertainty in #731 without changing the price.

Google's official Gemini API pricing page currently lists the Gemini 3.6 Flash standard paid tier at:

- input: $1.50 / 1M tokens
- output (including thinking): $7.50 / 1M tokens
- context-cached input: $0.15 / 1M tokens

Source: https://ai.google.dev/gemini-api/docs/pricing#gemini-3.6-flash

The existing `0.15` value was correct. This PR replaces the misleading blanket 25% rule/unverified warning with the provider-published 90% discount exception, and changes the regression test to pin that exact exception while retaining the 25% checks for the other rows.

Verification: 32 passed, 1 expected skip across the cached-pricing and usage suites; `git diff --check` passed.

Ready for external release review. No release performed.


wu-changxing:

PR #746, opened 2026-08-08
Refresh hosted agent balance metadata

### LINKED ISSUE #732: The published balance is a startup snapshot, so a broke agent advertises credit it does not have
On one screen of chat.openonion.ai, at one moment, the `oo` agent (0x97d85463…3abf) said both of these:

- the pill beside its address: **`$4.99  Top up →`**, in neutral styling
- the reply to my message:

```
Account: 0x97d85463...3abf Balance: $0.0005 Required: $0.0015 Shortfall: $0.0010
```

Same account, four orders of magnitude apart.

## Cause

`network/host/server.py::_extract_agent_metadata` says so itself:

```python
# This is a one-time startup snapshot; it refreshes on restart.
```

The agent started when it had $4.99 and has spent down to $0.0005 since. The pill is faithfully showing the snapshot.

## Why it defeats the thing it was built for

`oo-chat/components/agent-address.tsx` is careful about this number: `<= 0` renders "Out of credits" instead of "$0.00", and `< 1` turns the pill red, with the comment

> Below one dollar an agent is a turn or two from refusing, which is close enough that the reader should hear about it before it happens rather than after.

A stale snapshot inverts exactly that: at $0.0005 the agent is refusing *now*, and the pill is neutral and reassuring. The warning cannot fire, because the number it tests never changes.

## The frontend cannot work around it

`app/[address]/page.tsx` layers the authenticated `profile` frame over the public directory entry, and prefers `profile` — but `_build_agent_profile` fills `balance_usd` from the same startup metadata. Both paths carry the snapshot, so there is no fresher value for a client to choose. Whatever is done about this has to happen here.

## Options

1. Refresh it on a schedule (the agent is the only party that can — the balance is gated by its private key) and re-announce, or attach it to the periodic ping that already exists.
2. Send it with the timestamp it was taken, and let clients age it out — the pill can then say "at startup" or stop showing it.
3. Read it fresh when a client asks, on the authenticated socket only.

(1) is what makes the low-balance warning work as designed. (2) is the smallest change that stops the number lying. (3) puts a network call on a display path.

I have not picked one — it changes a published field's meaning on the eve of 1.6.0, and the sensible frontend behaviour depends on which way this goes, so changing oo-chat first could be wasted or wrong.

Found while running the chat end-to-end for the 1.6.0 loop. The round trip itself worked: message → relay → agent → LLM attempt → error surfaced in the UI with an actionable message and a Retry button.

wu-changxing:
Fixes #732.

## What changed

- managed-key hosted agents refresh their balance every 60 seconds in a worker thread, so the ASGI event loop stays responsive
- the shared metadata is updated for `/info` and authenticated `AGENT_PROFILE` consumers
- the relay profile is updated from the same value
- signed heartbeat ANNOUNCEs now retain the profile instead of silently dropping it, allowing the relay directory to receive refreshed balance data
- transient balance API failures retain the last known value and retry next interval instead of killing the refresh task
- bare `create_app()` users get the same refresh lifecycle even without a relay

The existing oo-chat low/out-of-credit UI needs no behavioral change; it can now receive a current value instead of a startup snapshot.

## Verification

- `pytest -q tests/unit/test_host_relay.py tests/unit/test_relay.py tests/unit/test_host_routes.py` — 89 passed
- compileall and `git diff --check` passed

Ready for external release review. No deployment or release performed.

### DIFF
--- a/connectonion/network/host/server.py
+++ b/connectonion/network/host/server.py
@@ -3,7 +3,7 @@ Purpose: Host agent as HTTP/WebSocket server with trust-based access control
 LLM-Note:
   Dependencies: imports from [network/asgi/, network/host/ws_router/ (run_ws_session), network/trust/, network/host/session/, network/host/auth.py, network/host/http_router.py, network/announce.py, network/relay.py] | imported by [network/__init__.py as host()] | tested by [tests/e2e/test_host.py]
   Data flow: host(create_agent, port, trust) → _create_route_handlers() wraps all routes → asgi_create_app() creates FastAPI/Starlette app → uvicorn.run() starts server → each request calls create_agent() for fresh instance → executes via input_handler()/ws_input() → returns result + session | trust enforcement via extract_and_authenticate() at request boundary
-  State/Effects: starts HTTP server on specified port | creates .co/logs/ directory | stores sessions in SessionStorage (in-memory with TTL) | optionally announces to relay server, publishing a display profile (alias/tools/model + project-level skills only — user/builtin skills stay private) with the first ANNOUNCE of each connection | each request gets fresh agent instance (no state bleeding)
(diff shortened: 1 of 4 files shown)


wu-changxing:

PR #747, opened 2026-08-08
Stop warning against documented host usage

### LINKED ISSUE #727: The default template warns against itself on first run
A brand-new project prints a yellow warning the moment it starts, about code the CLI wrote:

```
$ co create qs --yes && cd qs && python agent.py

  ◎    qs
●     connectonion v1.5.11
      gemini-3.6-flash · 24 tools · 118 skills
      credits on me, go build —aaron

Warning: host(agent) — pass a factory function instead: host(lambda: Agent(...))
```

The template is four lines and ends with `host(agent)` — exactly what the warning tells you not to do. So the first thing a new user sees is the framework disagreeing with the code it just generated.

## The advice would make the default template unusable

`server.py:734` — "Each request calls create_agent() to get a fresh Agent instance." Measured on this machine:

```
create_agent(role="coding")   first: 25.18s   second (warm): 8.62s   24 tools, 118 skills
```

Following the warning means **every request pays 8.6s** before the model is called, and each call also prints the full banner and makes a balance round-trip. For the co-ai agent that is not a trade-off, it is a broken server.

So the shape in the template looks right and the guidance looks wrong for this case — but the guidance is protecting something real (`host(agent)` shares one instance across all requests, with no per-request isolation), so which way this resolves is a decision about the hosting contract, not a docs fix.

## What I think the options are

1. **Do not warn for this shape.** If session isolation already comes from the session mechanism rather than from separate Agent instances, the warning is stale and costs every new user a scare.
2. **Keep the warning, make it say the trade-off** — that a factory isolates state and costs a full agent construction per request — so the reader can choose rather than assume they have done something wrong.
3. **Make construction cheap enough that a factory is viable**, which is the real fix if per-request isolation is wanted. 8.6s warm is mostly skill loading; 118 skills are being read per construction.

Not picking one. Option 2 is the cheapest honest improvement if 1 and 3 are not on for 1.6.0.

Found while fixing the Quick Start in `docs/connectonion.md` (#724) — the rewritten section tells the reader `python agent.py` just works, and it does, but this is the first line of output after it.

wu-changxing:
Fixes #727.

`host(agent)` is the generated template, the documented beginner path, and an explicit design decision across the repository. The runtime alone warned users that this supported path was wrong and suggested a factory that costs roughly a full agent/skill construction on every request.

This change:

- stops warning against `host(agent)`
- documents the actual contract: an instance shares configured state/resources; a factory creates a fresh Agent per request and pays that construction cost
- keeps factory behavior unchanged for users who need per-request isolation
- adds a regression test that exercises the documented instance path and ensures the contradictory warning stays gone

Verification: host relay + `co init` + `co create` suites, 73 passed; `git diff --check` passed.

Ready for external release review. No release performed.

### DIFF
--- a/connectonion/network/host/server.py
+++ b/connectonion/network/host/server.py
@@ -731,14 +731,19 @@ def host(
     Configuration: .co/host.yaml (required) with code param overrides.
     Run 'co init' to generate the config file.
 
-    Each request calls create_agent() to get a fresh Agent instance.
-    This ensures complete isolation between concurrent requests.
+    Passing an Agent instance is the simple path and shares that instance.
+    Passing a factory creates a fresh Agent for each request.
 
-    State Control via Closure:
-        # Isolated state (default, safest) - create inside:
+    State Control:
+        # Simple/default: share one configured agent and expensive tool setup:
+        agent = Agent("assistant", tools=[BrowserTool()])
+        host(agent)
+
+        # Per-request isolation: create everything inside a factory:
         def create_agent():
             browser = BrowserTool()  # Fresh per request
             return Agent("assistant", tools=[browser])
+        host(create_agent)
 
         # Shared state (advanced) - create outside, capture via closure:
         browser = BrowserTool()  # Shared across all requests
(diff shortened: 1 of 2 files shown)


wu-changxing:
Cross-PR review note: this PR overlaps `connectonion/network/host/server.py` with #741.

Merge/rebase #741 first, then preserve both changes during conflict resolution: keep this PR's documented `host(agent)` behavior, but retain #741 head `5e8184a`'s `DEFAULT_BACKEND_WS_URL` import, configured-backend relay fallback, legacy generated production-relay migration, and updated relay docstring.

The stacked integration candidate currently contains both and its focused host/relay/backend suite passes.

wu-changxing:

PR #748, opened 2026-08-08
Refuse browser do across billing accounts

### LINKED ISSUE #728: co browser do bills whoever started the daemon, not who ran the command
Ran `co browser do` from a worktree whose `.env` holds my machine key (balance $711) and got:

```
Account:     0x561605f3...dbe4
Balance:     $-0.0702
Required:    $0.0088
```

That is a different account from the one the command was run under, and the caller has no way to tell from the invocation.

## Why

The daemon is shared and long-lived by design — one browser, many callers — and `do` runs the agent **inside it**:

```python
if verb == "do":  # explicit natural-language agent
    command = " ".join(tokens[1:])
    return self._run_nl(command)
```

So the LLM credentials come from the daemon's environment, not the caller's. Verified:

```
daemon pid 42504
daemon cwd /Users/changxing/project/OnCourse/platform     ← another session started it
that dir's .env  OPENONION_API_KEY=eyJhbGciOiJI…          ← account 0x561605f3
my worktree .env OPENONION_API_KEY=…                      ← account 0x10e68f6d, $711
```

Every page verb (`go_to`, `click`, `get_text`) is free of this — they touch no model. Only `do` spends, and it spends the daemon owner's credits.

## Why it matters beyond this machine

On a single-user machine with one account this is invisible. On a machine where several projects each have their own identity — which is what `co create` produces, and what this machine has — the first session to touch the browser silently becomes the payer for everyone else. The error names an account the reader has nothing to do with, so the first symptom is confusing rather than informative.

It also crosses a boundary the rest of the CLI is careful about: `co status`, `co keys` and `co ai` all resolve credentials from the caller's directory.

## Not proposing the fix

Three shapes I can see, and picking between them is a call about what the daemon is:

1. **Carry the caller's key with the request.** `do` already sends a caller identity for tab contention (`CO_WHO`); a key could ride along. Means a credential crosses a socket, which is worth thinking about even locally.
2. **Refuse `do` when the caller's key differs from the daemon's**, and say so — cheap, honest, and pushes the user to run `do` from the session that owns the browser.
3. **Say who is paying.** Leave the behaviour and name the account in the error and in `co browser status`, so at least it is not a surprise.

3 is the smallest and fixes the confusion without deciding the design; 1 is the only one that makes the billing match the invocation.

Found doing the release e2e for 1.6.0 — the page half works fine (chat.openonion.ai loads, 5 agents online, history renders); this is what stopped the interactive half.

Separately, the balance is **negative** ($-0.0702), so the backend let it go under zero before refusing. That is oo-api's side and not this issue.

wu-changxing:
Fixes #728 using the issue's option 2: refuse model-backed `do` unless the caller and daemon billing accounts are both known and match.

## What changed

- the client adds its public OpenOnion account address to the existing per-user local envelope; no API key/private key crosses the transport
- before `do` claims a tab, constructs an Agent, or spends tokens, the daemon compares that address with the account in its own environment
- mismatch exits with code 5 and names both shortened public addresses plus the restart action
- legacy/accountless clients and an unreadable daemon identity fail closed instead of silently spending from an unverifiable account
- hex addresses compare case-insensitively
- inability to identify the caller's payer also fails before connecting for `do`
- page-only browser commands remain shared across accounts because they do not invoke a model
- legacy plain page requests remain readable by an upgraded daemon

## Verification

- 10 focused billing-boundary tests passed
- complete daemon/transport suite passed outside the managed socket sandbox: **105 passed**
- `git diff --check` passed

Ready for external release review. No merge or release performed.

### DIFF
--- a/connectonion/cli/browser_agent/client.py
+++ b/connectonion/cli/browser_agent/client.py
@@ -2,7 +2,7 @@
 Purpose: Thin client for the browser daemon — wraps one command as a JSON envelope, sends it over the platform transport (POSIX Unix socket / Windows named pipe), and maps the reply to stdout/stderr/exit code.
 LLM-Note:
   Dependencies: imports from [socket, os, json, sys, time, pathlib, browser_agent.daemon (default_sock_path, _owner_alive), browser_agent.transport] | imported by [cli/commands/browser_commands.py] | tested by [tests/e2e/cli/test_browser_daemon.py]
-  Data flow: send(line, headless, tab) → _caller() derives identity (CO_WHO, else claude-<CLAUDE_JOB_DIR>, else "") → build wire-v1 envelope json{v:1, caller, tab, line} (structured caller/tab means any character in a name is safe — nothing is spliced into the shlex-parsed command line) → _connect(default_sock_path()) or _spawn_daemon() → POSIX: send, half-close, read reply to EOF over raw AF_UNIX | Windows: send_bytes/recv_bytes one framed message each way over an HMAC-authenticated named pipe (transport.win_connect + load_or_create_authkey) → header "OK" prints payload to stdout & returns 0 | header "ERR"/"ERR <n>" prints payload to stderr & returns the int (1 default, else 2/3/4) → a pre-upgrade daemon's "unknown command: {…" reply is rewritten to a restart hint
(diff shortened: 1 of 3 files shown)


wu-changxing:
Release audit found and fixed a fail-open path in b1f9bc8: legacy/accountless `do` requests and an unavailable daemon billing identity could still invoke the model, and a rejected mismatch claimed the tab first. Billing verification now happens before any tab claim and permits `do` only when both public account addresses are present and equal. Full daemon/transport regression: 105 passed outside the Unix-socket sandbox. No merge or release performed.

wu-changxing:

PR #750, opened 2026-08-08
security: sign every protocol-v2 command frame

wu-changxing:
Advances #649.

## What changes

- advertises `signed_commands: 1` inside the signed CONNECT payload
- signs each complete INPUT, EXEC, runtime-input, approval, ask-user, and ADMIN command with type, recipient, timestamp, and a random nonce
- binds each command to the caller that opened the socket and to the intended host
- rejects replayed command signatures and top-level type/content substitution
- dispatches the verified payload instead of unsigned compatibility copies
- keeps protocol-v1 clients working; v1 connections remain on the legacy connection-authenticated behavior

## Compatibility / rollout

The capability is signed, so it cannot be stripped from a v2 CONNECT without invalidating authentication. New clients duplicate command fields at the top level so they can still talk to v1 hosts. Deploying a v2 host first is compatible with old clients; the stronger command guarantee applies once both sides are v2.

This is Ready for external security review before 1.6.0. It does not publish or release anything.

## Verification

- 60 focused security/admin tests passed
- 207 broader network/auth/onboarding/EXEC/skill-permission tests passed
- `python -m compileall -q connectonion/network`
- `git diff --check`

### DIFF
--- a/connectonion/network/connect.py
+++ b/connectonion/network/connect.py
@@ -1,8 +1,8 @@
 """
-Purpose: Client SDK for talking to remote ConnectOnion agents over websockets — handles relay/direct endpoint resolution, signed CONNECT/INPUT frames, streaming UI events, and onboard (invite/payment) flows.
+Purpose: Client SDK for talking to remote ConnectOnion agents over websockets — handles relay/direct endpoint resolution, signed CONNECT and per-command frames, streaming UI events, and onboard (invite/payment) flows.
 LLM-Note:
   Dependencies: imports from [asyncio, json, time, uuid, dataclasses, typing, httpx, websockets (lazy), .. address (sign)] | imported by [network/__init__.py (re-exports connect, RemoteAgent, Response), connectonion/__init__.py (top-level re-export)] | tested by [tests/unit/test_connect.py, tests/unit/test_connect.py]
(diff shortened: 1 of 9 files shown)


wu-changxing:
Security follow-up pushed in `10731af` after adversarial review. Three concrete gaps are now regression-tested and fixed:

- a signed command with a missing `to` is rejected (recipient binding is mandatory when the host address is known)
- CONNECT authentication is bound to `agent_metadata.address`, not the caller-controlled CONNECT recipient
- an authenticated WebSocket rejects a second CONNECT, preventing an in-place v2 → v1 capability downgrade

Local evidence before push:
- targeted red phase: all 3 new attack-path tests failed on the old implementation
- protocol/auth/host regression set: **252 passed**
- ASGI/WebSocket compatibility set: **81 passed**
- `git diff --check`: clean

The PR remains Draft. CI is intentionally being allowed to rerun before merge/release consideration.

wu-changxing:
Final verification update for head `354cc16`:

- all 8 GitHub CI jobs are green (Python 3.10–3.13 and Windows)
- stacked 1.6 candidate: **6023 passed, 18 skipped, 174 deselected, 0 failed** (`--randomly-seed=1870689475`)
- corrected WebSocket E2E fixture: 28/28 focused tests pass
- rebuilt 1.6.0 wheel/sdist pass `twine check`; installed-wheel suite is 8/8
- installed wheel directly rejects a validly signed command whose payload omits `to`

PR remains Draft for independent security review. No merge, tag, upload, or deployment was performed.

wu-changxing:

PR #751, opened 2026-08-08
Handle managed provider failures without tracebacks

wu-changxing:
Related to openonion/oo-api#110.

A production probe on 2026-08-08 confirmed that `co/claude-sonnet-4` still receives an upstream 401 because the deployed Anthropic credential is invalid. That credential must be fixed in oo-api; this PR fixes the client behavior when a provider fails.

## Changes

- translates managed-route 401s into the shared `LLMAuthenticationError` family
- tells users that a `co/*` credential rejection is service-side, rather than asking them to change a key they do not own
- catches known provider failures at the one-shot `co ai` boundary, prints one actionable message, and exits 1
- preserves original exceptions as `__cause__` for library/debug callers
- deliberately lets programmer errors such as `TypeError` keep their traceback

## Verification

- 50 focused provider/CLI/error tests passed
- 115 broader OpenOnion/Anthropic/Groq/Agent tests passed
- compileall and diff check passed

Ready for external review before 1.6.0. This does not fix or rotate the production Anthropic credential and does not publish anything.

### DIFF
--- a/connectonion/cli/commands/ai_commands.py
+++ b/connectonion/cli/commands/ai_commands.py
@@ -1,10 +1,16 @@
 """
-Purpose: AI coding agent CLI command
+Purpose: AI coding agent CLI command with concise provider-failure reporting
 LLM-Note:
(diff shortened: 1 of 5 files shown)


wu-changxing:

PR #752, opened 2026-08-08
Raise vulnerable dependency floors for 1.6.0

wu-changxing:
## Summary

- require `python-dotenv>=1.2.2` instead of `>=1.0.0`
- require `PyNaCl>=1.6.2` instead of `>=1.5.0`
- regenerate the stale `uv.lock`, which had still selected `python-dotenv 1.2.1` and `PyNaCl 1.6.1` and described a much older project dependency graph
- extend the regression test to verify both declared security floors and the concrete locked versions/specifiers

A fresh pip install generally selected current releases already, but an existing environment upgrading ConnectOnion could retain the older versions because they still satisfied the prior lower bounds. A locked uv environment would also have continued installing the vulnerable versions before the follow-up fix. Those versions are affected by the dotenv symlink-overwrite and libsodium point-validation advisories found during the 1.6.0 audit.

## Verification

- `uv lock --check` passes against the regenerated lock
- dependency/security and address crypto tests pass
- wheel and sdist build successfully in the earlier artifact rehearsal
- built wheel metadata contains `pynacl>=1.6.2` and `python-dotenv>=1.2.2`
- package-content scan found no `.env`, credential files, private keys, or test directories; the only private-key signature was the intentional OpenSSH serialization implementation

Ready for external review before 1.6.0. No merge or release performed.


wu-changxing:
Additional release-artifact gate passed after the dependency-floor change:

- `pytest -q tests/e2e/test_the_wheel_works_when_installed.py -m slow`: 8 passed

This installs the built wheel into a throwaway venv outside the repository and verifies the import source, packaged trust policies/prompts/template/docs, `co --version`, and `co init` output.

wu-changxing:
Release audit found the tracked uv.lock was stale and still selected python-dotenv 1.2.1 plus PyNaCl 1.6.1, despite the new pyproject floors. `uv lock --check` reproduced the inconsistency. a33b60a regenerates the lock against the current project graph and extends the regression to assert both locked versions and metadata specifiers. `uv lock --check` and focused crypto/dependency tests now pass. No merge or release performed.

wu-changxing:

PR #753, opened 2026-08-08
Make release artifact uploads version-safe

wu-changxing:
## Summary

- require removing old `dist/` artifacts before building a release
- run `twine check` against the exact current-version wheel and sdist
- upload only those two explicit files instead of `dist/*`
- add release-procedure regression tests for ordering and version scoping

`python -m build` does not clean `dist/`. During the 1.5.13 → 1.6.0 transition, the old artifacts otherwise remain beside the new files. PyPI uploads are not atomic, so a broad glob can partially upload before failing on a stale/duplicate artifact.

## Verification

- release checklist/version/release-history tests: 20 passed
- `git diff --check` clean

Ready for external review. This PR does not build 1.6.0, upload to PyPI, create a tag, or create a GitHub release.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -100,9 +100,16 @@ When releasing a new version:
 - [ ] Create git tag: `git tag vX.Y.Z`
 - [ ] Push commits: `git push`
 - [ ] Push tag: `git push origin vX.Y.Z`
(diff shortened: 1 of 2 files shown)


wu-changxing:

PR #758, opened 2026-08-08
security(doctor): redact and diagnose credentials

### LINKED ISSUE #468: co doctor --fix: diagnose credentials and guide provider authentication
Parent epic: #213

## Why

Users should get one actionable answer when managed keys or provider credentials are absent, expired, shadowed, or stored in the wrong place.

## Scope

- Inspect credential presence and source without printing secrets
- Distinguish OpenOnion, OpenAI, Anthropic, Google, Microsoft, and OAuth connection problems
- Offer the correct login/connect command and re-check after it completes

## Definition of done

- [ ] No secret value appears in normal output or logs
- [ ] Every supported provider has a concrete next action
- [ ] Tests cover missing, expired, shadowed, and successful credentials

## Dependencies

Use the credential discovery rules already shared by co status.

## Non-goals

Do not expand the parent epic's product surface.

wu-changxing:
Implements #468 with an explicit credential-safety boundary.

This removes the API-key prefix preview from normal `co doctor` output and adds redacted diagnostics for OpenOnion/model-provider keys plus Google and Microsoft OAuth. It reports source precedence, the value source that wins when another is shadowed, missing scopes, malformed/expired tokens, and the exact reconnect/configuration command without returning secret values. Running `co doctor` again re-reads every source and verifies the resulting state.

Credential repair remains manual by design: `co auth google` and `co auth microsoft` revoke the existing provider connection before starting a new browser login, so they are not attached to the broad `doctor --fix --yes` approval in #759/#760. Doctor names the exact command; the user explicitly runs that provider-specific mutation, then reruns doctor.

Validation:
- 9 focused credential tests passed, covering missing, connected, expired, refreshable, malformed, missing-scope, and shadowed states
- 158 doctor/status-related unit tests passed with an isolated HOME
- output tests prove neither full secrets nor the former 20-character preview appear
- `py_compile` and `git diff --check` passed

Closes #468 after merge.

### DIFF
--- a/connectonion/cli/commands/doctor_commands.py
+++ b/connectonion/cli/commands/doctor_commands.py
@@ -278,32 +278,55 @@ def handle_doctor():
     else:
         config_table.add_row("Keys", "[yellow]○[/yellow] Not found (run 'co auth' to create)")
 
-    # Check for API key
-    api_key = os.getenv("OPENONION_API_KEY")
-    if api_key:
-        api_key_display = f"{api_key[:20]}..." if len(api_key) > 20 else api_key
-        config_table.add_row("API Key", f"[green]✓[/green] Found in environment")
-        config_table.add_row("Key Preview", f"[dim]{api_key_display}[/dim]")
-    else:
-        # Check .env files
-        from dotenv import load_dotenv
-        local_env = Path(".env")
-        global_env = Path.home() / ".co" / "keys.env"
+    # The same redacted discovery `co status` uses. Doctor used to print the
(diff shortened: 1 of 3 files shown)


wu-changxing:
Companion email troubleshooting docs now use the precedence-aware co doctor output instead of grepping one keys.env file: https://github.com/wu-changxing/connectonion-docs-site/pull/29. Draft only; no merge performed.

wu-changxing:

Issue #25, opened 2025-10-05
Update docs website with current features and getting started guide


wu-changxing:
## Description

Update the docs website (docs-site) to reflect current ConnectOnion features and provide an excellent first-time user experience.

## Problem

New users visiting https://docs-site need:
- Clear getting started path
- Up-to-date feature documentation
- Working examples they can copy-paste
- Visual navigation that guides them

### Homepage & Navigation
- [ ] Update homepage with clear value proposition
- [ ] Add "Quick Start" prominently on homepage
- [ ] Update navigation with all current features
- [ ] Add search functionality

### Getting Started Page
- [ ] 60-second quick start (install → first agent → working)
- [ ] Step-by-step tutorial with code blocks
- [ ] Common errors & solutions
- [ ] Next steps after basic usage

### Feature Documentation
- [ ] Agent basics (create, tools, system_prompt)
- [ ] @xray decorator explained
- [ ] Tool creation (function-based + class-based)
- [ ] LLM providers (OpenAI, Anthropic, Google)
- [ ] Console logging & behavior tracking

### Examples & Templates
- [ ] Copy-paste examples for common use cases
- [ ] Template agent code
- [ ] Best practices guide
- [ ] Real-world examples

### Visual Design
- [ ] Mobile-friendly responsive design
- [ ] Code blocks with syntax highlighting
- [ ] Copy button for all code examples
- [ ] Clear visual hierarchy

## User Journey to Optimize

**First-time visitor:**
1. Lands on homepage → Sees "What is ConnectOnion?" in 10 seconds
2. Clicks "Quick Start" → Has working agent in 2 minutes
3. Explores examples → Finds relevant use case
4. Reads feature docs → Understands @xray, tools, etc.
5. Joins community → Discord link visible

## Files to Create/Update

- [ ] `docs-site/app/page.tsx` - Homepage
- [ ] `docs-site/app/getting-started/page.tsx` - Quick start
- [ ] `docs-site/app/features/*/page.tsx` - Feature pages
- [ ] `docs-site/lib/navigation.ts` - Updated navigation
- [ ] `docs-site/components/CodeBlock.tsx` - Improved code blocks
- [ ] `docs-site/app/examples/page.tsx` - Examples gallery

## Success Criteria

- [ ] First-time user can create working agent in < 3 minutes
- [ ] All current features documented
- [ ] Mobile-friendly (test on phone)
- [ ] No broken links
- [ ] Search works for key terms

## Priority

[...]

PR #761, opened 2026-08-08
docs: keep the shipped quickstart safe

wu-changxing:
## Summary
- replace the shipped `eval()` calculator with a deterministic typed tool
- replace simulated search output with a real local transformation
- correct credential wording in the CLI flow and generated file tree
- add regression tests so unsafe/fake examples do not return

## Why this is separate
`docs/quickstart.md` ships in the wheel and is copied into generated projects. The docs-site fix alone would leave this source stale and able to drift back into published documentation.

## Validation
- red test reproduced both unsafe examples before the change
- `pytest -q tests/unit/test_the_quickstart_uses_safe_real_tools.py tests/unit/test_packaging.py tests/unit/test_the_shipped_reference_offers_real_templates.py`
- 20 passed
- `git diff --check`

Related to #25. This PR does not release or publish anything.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -19,7 +19,7 @@ co create my-agent
 # Navigate to the project
 cd my-agent
 
-# Run your agent (API key setup is automatic!)
+# Run your agent (uses your configured provider or ConnectOnion login)
 python agent.py
 ```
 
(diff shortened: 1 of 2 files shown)


wu-changxing:

PR #764, opened 2026-08-08
Deliver runtime input before a turn closes

### LINKED ISSUE #762: runtime_input never reaches a turn that runs no tools
`useful_plugins/runtime_input.py` hooks `@before_iteration`, so a message pushed into a running session is only picked up if that session starts **another** iteration. A turn that calls a tool does; a plain question does not, and the message is acknowledged and then dropped.

For a chat client this is the common case rather than an edge one — most turns answer without calling anything.

## Reproduced both ways

One hosted agent, `Agent(name="…", tools=[echo], plugins=[skills, runtime_input], model="co/gemini-2.5-flash")`, one client, messages ~3s apart.

**Turn that calls a tool — picked up:**

```
[co] > "echo the word banana"
↳ RUNTIME_INPUT session=59723425... prompt=also what is SSE...
[co]   ▸ echo(text="banana")                           ✓ 0.00s
[co] ✓ complete · 6.6s
```

The reply addressed both the echo and SSE, exactly as `RUNTIME_INPUT_FRAME_PREFIX` asks for.

**Turn that calls nothing — dropped:**

```
[co] > "what is HTTP in one line"
↳ RUNTIME_INPUT session=59723425... prompt=and TCP too...
[co] ✓ complete · 3.0s
```

One iteration, so `apply_runtime_input` never runs again. The reply covered HTTP only; the second message got no answer and no error. `ws_router/session.py` had already sent `RUNTIME_INPUT_ACK`, so from the client there is nothing to distinguish this from the working case.

## Why it is hard to see from outside

`RUNTIME_INPUT_ACK` means "queued", not "will be read". A client that receives it has no way to learn the message was never drained — the run completes normally and its OUTPUT simply answers the earlier prompt. We spent a while treating this as a client bug before finding `pop_runtime_inputs` was never called a second time.

## Possible shapes of a fix

- Drain the queue once more before the run completes (an `@on_complete`/`after_iteration` counterpart to the existing handler), so a single-iteration turn still sees it. If anything is pending, that would presumably need one more LLM call to answer it.
- Or, if a single-iteration turn genuinely cannot absorb late input, say so on the wire — a `RUNTIME_INPUT_DROPPED` (or an ACK carrying whether it will be read) would at least let a client fall back deliberately.

Happy to test a patch against the same setup.

## Environment

`connectonion` at `1.0.2` (from source, `network/host/server.py` with `_build_agent_profile`), relay `wss://oo.openonion.ai`, model `co/gemini-2.5-flash`.

wu-changxing:
## What changed

- drain runtime input once more at the final no-tool iteration boundary and run another LLM iteration when a follow-up arrived
- make the WebSocket runtime-input window explicit and atomic, so a message is either queued or rejected instead of receiving a false ACK
- reject runtime input for agents that did not opt into the `runtime_input` plugin
- return failed agent threads from `running` to `connected`, preventing later input from being queued to a dead worker
- make `after_iteration` fire for final no-tool iterations while keeping the no-progress guard from counting an old tool call twice

## Root cause

`runtime_input` only drained its queue in `before_iteration`. A normal answer with no tool calls returned immediately, so input received during that LLM call remained queued forever even though the host had sent `RUNTIME_INPUT_ACK`. The same false-ACK state occurred after an agent exception because the active-session registry stayed `running`.

## User impact

A follow-up sent while a plain answer is being generated is now visible in session history and receives another LLM answer. If the turn has already closed its acceptance window, the client gets a retryable `RUNTIME_INPUT_REJECTED` error rather than a success acknowledgement. Agents without the plugin also fail explicitly.

## Validation

- focused agent/event/IO/host regression suite: **122 passed**
- full randomized non-paid suite (seed `762`): **5873 passed, 18 skipped, 174 deselected**, 0 failed
- `ruff` on the new runtime-input implementation/tests: clean
- `git diff --check`: clean

Closes #762.

Ready for human review before inclusion in the 1.6.0 release candidate. No release, tag, merge, or deployment was performed.

### DIFF
--- a/connectonion/core/agent.py
+++ b/connectonion/core/agent.py
@@ -449,10 +449,9 @@ class Agent:
             if not response.tool_calls:
                 content = response.content or ""
                 self.current_session['messages'].append({"role": "assistant", "content": content})
-                return content
-
-            # Process tool calls
-            self._execute_and_record_tools(response.tool_calls)
+            else:
+                # Process tool calls
+                self._execute_and_record_tools(response.tool_calls)
 
             # Fire after_iteration
             self._invoke_events('after_iteration')
@@ -461,6 +460,14 @@ class Agent:
             if self.current_session.pop('stop_signal', None):
                 self._invoke_events('on_stop_signal')
                 return "What would you like me to do?"
+
+            if not response.tool_calls:
+                if self.current_session.pop('_continue_iteration', False):
+                    # An accepted follow-up is new work. Give it the LLM call it
+                    # needs even when the original request used its full budget.
+                    max_iterations += 1
+                    continue
+                return content
 
         # Hit max iterations
         return f"Task incomplete: Maximum iterations ({max_iterations}) reached."
--- a/connectonion/network/host/ws_router/agent_io.py
+++ b/connectonion/network/host/ws_router/agent_io.py
@@ -2,7 +2,7 @@
 Purpose: Agent thread orchestration + io → client streaming — the bridge between the sync agent and the async transport
 LLM-Note:
   Dependencies: imports from [...io (WebSocketIO), ..session (session_to_chat_items via lazy import), asyncio, threading, rich.console] | imported by [.session (start_agent), .connect (resume_forwarding)]
-  Data flow: start_agent: validate INPUT → create WebSocketIO → register in registry BEFORE thread.start (race-safe) → spawn _agent_thread_body (thread target running route_handlers["ws_input"]) → spawn forward_task = forward_agent_msgs_to_client → return (io, task) | resume_forwarding: same forward_task spawn but on existing active.io | forward_agent_msgs_to_client: read_msgs_from_agent → send_msg loop → on agent finish read result_holder → emit OUTPUT (success) / ERROR (exception)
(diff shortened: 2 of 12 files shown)


wu-changxing:
Stacked 1.6 integration evidence:

- cherry-picked after the existing email, security, doctor, dependency, docs, and release-workflow candidates with no conflicts (`ab4a781`);
- focused runtime-input/agent/event/IO/host suite: **122 passed**;
- full randomized stacked suite (seed `1600762`): **5969 passed, 18 skipped, 174 deselected**, 0 failed, in 499.74s;
- GitHub matrix: **8/8 success**; PR remains Draft and CLEAN;
- integration worktree and `git diff --check`: clean.

No merge, tag, release, or deployment was performed.

wu-changxing:
Final stacked 1.6 validation includes this runtime-input fix plus the complete skill, doctor, release, email, security, and dependency candidate set:

- focused runtime/skill/doctor suite: **159 passed**
- full Connect suite: **6020 passed, 18 skipped, 174 deselected, 0 failed**
- random seed: `1600757`
- `git diff --check`: clean

The initial sandboxed run produced permission cascades when tests tried to write `~/.co`, bind sockets, and launch Chrome; rerunning the identical seed with the required test permissions passed fully. Human review/merge checklist: #765.

wu-changxing:
Review follow-up: the implementation's atomic accept/reject and final no-tool drain behavior remains sound. I found the packaged protocol and IO docs still described the old iteration-start-only semantics; commit e828d89 now documents truthful ACK/retryable rejection and the final drain-or-seal boundary. Focused runtime/IO/no-progress suite: 44 passed, 0 failed. No merge performed.

wu-changxing:

PR #769, opened 2026-08-08
Release 1.5.15 integration candidate

wu-changxing:
## Purpose\n\nBrings the exercised 1.5.15 test-release integration candidate back to `main`. Commit `0a54e00` repairs the artifact acceptance runner after the v1.5.14 workflow stopped before PyPI because `pytest` was missing. No 1.5.14 package was uploaded.\n\n## Key release gates\n\n- oo-api production migration applied and backend SHA `7975470` deployed successfully\n- production email send + same-key retry returned one provider message id and one Sent row\n- full CI for the 1.5.14 attempt passed on Python 3.10-3.13 plus Windows E2E/transport before the runner-only build failure\n- release workflow fix has a regression test (focused release/version suite 20/20 passed)\n- local 1.5.15 wheel/sdist passed `twine check`\n- clean 1.5.15 wheel install acceptance: 8/8 passed\n- central 1.6.0 review checklist: #765\n\n## Review note\n\nThis PR intentionally excludes unfinished profile-freshness work tracked in #768. Please review this integration as the deployed/published patch baseline before 1.6.0.

Issue #32, opened 2025-10-17
[BUG]Iteration problem


Wood-forest:
**Describe the bug**
Even the task is done successfully, agent will auto run the next iteration

<img width="2380" height="1232" alt="Image" src="https://github.com/user-attachments/assets/72c03922-f538-47e6-9f4c-0a9340d978de" />

[2025-10-16 00:09:00] TOOL_START name=edit_playwright_script args={'file_path': 'test_project/screenshot_script.py', 'instructions': [{'type': 'change_browser', 'content': ['chromium', 'firefox']}]}
[2025-10-16 00:09:00] TOOL_END name=edit_playwright_script ok=1 result={'status': 'success', 'diff_preview': '--- before\n+++ after\n@@ -3,14 +3,14 @@\n \n def run_task():\n     with sync_playwright() as p:\n-        browser_type = "chromium"\n+        browser_type = "firefox"\n         browser_launcher = getattr(p, browser_type)\n         browser = None\n         try:…
[2025-10-16 00:09:31] TOOL_START name=edit_playwright_script args={'file_path': 'test_project/screenshot_script.py', 'instructions': [{'type': 'change_browser', 'content': ['chromium', 'firefox']}]}
[2025-10-16 00:09:31] TOOL_END name=edit_playwright_script ok=1 result={'status': 'success', 'diff_preview': '', 'changes': 0}
[2025-10-16 00:09:31] AGENT_OUTPUT Task incomplete: Maximum iterations (2) reached.
[2025-10-16 00:11:55] user input: I want to add test_project to git
[2025-10-16 00:12:00] LLM_TASK {"tool":"manage_git","arguments":{"action":"add","args":{"paths":"test_project"}}}
[2025-10-16 00:12:44] TOOL_START name=manage_git args={'action': 'add', 'args': {'paths': 'test_project'}}
[2025-10-16 00:12:44] TOOL_END name=manage_git ok=1 result=✅ Added changes from 'test_project'
[2025-10-16 00:12:51] AGENT_OUTPUT ✅ Added changes from 'test_project'
[2025-10-16 00:12:55] user input: I want to commit test_project to git
[2025-10-16 00:12:59] LLM_TASK What commit message would you like to use for committing to Git?
[2025-10-16 00:13:15] TOOL_START name=manage_git args={'action': 'commit', 'args': {'message': 'Update screenshot script to use Firefox browser'}}
[...]

PR #770, opened 2026-08-08
1.6.0 final review candidate — security, email, and release hardening

### LINKED ISSUE #213: Epic: co doctor --fix first-run bootstrap and safe repair
## Goal

A non-technical user — especially on Windows — should get from `pip install connectonion` to "everything works" through **one command path, zero errors**. Today the pieces exist but the user has to know them; when something is missing they get an error and must act themselves.

## Proposal

`co doctor --fix`: doctor already *diagnoses*; `--fix` makes it *repair* everything repairable, interactively confirming anything that downloads or spends:

1. **Browser binary missing** (the #1 Windows trap): offer/run `python -m patchright install chrome` automatically, then verify with a real headless `about:blank` launch — not just the stealth-driver check (doctor today shows ✓ even when no Chrome exists and launch would fail).
2. **No keys / auth expired**: run the `co auth` flow (or print the exact one command if a browser login is required).
3. **Corrupt browser IPC state**: clear a stale authkey/pid sidecar (transport.py self-heals the authkey already; `--fix` should also know these files exist — today `co doctor`/`co reset` don't).
4. **PATH problems**: detect `co` not being on PATH (invoked via `python -m`), print the pipx/Scripts-dir remedy.
5. Exit summary in plain language: what was fixed, what still needs the human (login, credits).

## What already exists (build on, don't duplicate)

- The **install-connectonion skill** (#201/#211) is this exact flow for agent-driven setups (Claude Code/codex driving `co`) — auto-corrects, retries, ends with a plain-language summary. `co doctor --fix` is the same playbook for humans without an agent.
- First-run robustness already hardened (PR #209): offline `co init` is graceful (no traceback, scaffold + keypair still created, "run `co auth` later" guidance); windows-e2e CI simulates the full fresh-user path (wheel install → CLI discovery → patchright install → real Chrome → parallel agents → offline first run).

## Acceptance

- Fresh Windows VM: `pip install connectonion && co init --yes && co doctor --fix` → `co browser go_to example.com` works, no manual steps besides the auth browser login.
- windows-e2e CI gains a `doctor --fix` step (it can validate the browser-install repair end to end; auth repair stays mocked/offline-asserted).
- Downloads always announced before they run; `--yes` for unattended use.

Refs: #208 (async transport RFC), PR #209 (Windows native browser + e2e CI), #211 (install-connectonion skill rewrite).

### LINKED ISSUE #301: Epic: portable skill dependency contracts from install through deploy

### LINKED ISSUE #467: co doctor --fix: inspect and repair browser and runtime prerequisites
Parent epic: #213

## Why

A fresh install should not fail on the first browser or agent command because an executable, browser runtime, or supported Python environment is missing.

## Scope

- Report Python, package, browser daemon, browser binary, and OS-level prerequisite state
- Offer only safe, known repairs and show what will change before doing it
- Re-run checks after repair and distinguish repaired, skipped, and still-blocked items

## Definition of done

- [ ] A fresh supported machine reaches a working browser/runtime path in one command
- [ ] Repeated runs are idempotent
- [ ] Tests cover missing, repaired, unsupported, and permission-denied states

## Dependencies

Parent #213 defines the top-level doctor workflow.

## Non-goals

Do not expand the parent epic's product surface.

### LINKED ISSUE #468: co doctor --fix: diagnose credentials and guide provider authentication

### LINKED ISSUE #469: co doctor --fix: deterministic repair plans, JSON output, and non-interactive safety
Parent epic: #213

## Why

A repair command must be scriptable and must never interpret silence as approval.

## Scope

- Represent checks and proposed repairs as a stable plan
- Add machine-readable output for CI/support bundles
- Require explicit approval per mutation unless an explicit non-interactive flag is supplied
- Record which repairs ran and their results

## Definition of done

- [ ] The same starting state produces the same ordered repair plan
- [ ] Non-interactive mode fails closed without its explicit approval flag
- [ ] JSON schema and idempotency are tested

## Dependencies

Depends on the concrete repair actions delivered by the other #213 children.

## Non-goals

Do not expand the parent epic's product surface.

### LINKED ISSUE #470: Define a versioned dependency manifest for skills
Parent epic: #301

## Why

A skill needs a portable way to declare the Python packages, executables, environment variables, OAuth connections, and platform capabilities it requires.

## Scope

- Choose the SKILL.md frontmatter/schema shape
- Separate required from optional capabilities
- Include version constraints and human-readable setup hints
- Version the schema for future additions

## Definition of done

- [ ] Invalid manifests fail with the skill name and exact field
- [ ] Existing skills without a manifest continue to load
- [ ] Parser and schema examples are documented and tested

## Dependencies

Parent #301 owns the end-to-end dependency contract.

## Non-goals

Do not install anything in this issue.

### LINKED ISSUE #471: Validate skill requirements in co doctor, skills install, and co ai startup
Parent epic: #301

## Why

A declared dependency is useful only if failure happens before the skill starts burning agent iterations.

## Scope

- Resolve a skill manifest against the current machine
- Report packages, binaries, credentials, and capabilities separately
- Integrate the same validator into doctor, skill installation, and startup diagnostics
- Keep optional requirements informational

## Definition of done

- [ ] Missing requirements produce one actionable preflight report
- [ ] Validation has no network side effects
- [ ] The three entry points share one implementation and tests

## Dependencies

Depends on the manifest schema child of #301.

## Non-goals

Do not auto-install dependencies here.

### LINKED ISSUE #472: Package and install declared skill runtime dependencies during deploy
Parent epic: #301

## Why

A skill that works locally must carry enough information to work on a clean deployment target.

## Scope

- Include skill manifests and required assets in deploy packaging
- Install supported declared Python dependencies in the agent venv
- Preflight unsupported binaries/capabilities before restart
- Stamp the realized dependency state rather than only the requested state

## Definition of done

- [ ] A clean server can deploy a skill with declared dependencies successfully
- [ ] A failed install is retried on the next deploy
- [ ] Unsupported requirements fail before the service is restarted
- [ ] Integration tests cover changed and unchanged manifests

## Dependencies

Depends on the #301 manifest and validator children; align install verification with #465.

## Non-goals

Do not expand the parent epic's product surface.

### LINKED ISSUE #557: The default co-ai skills are for developing connectonion, not for using it

### LINKED ISSUE #654: co sub sync installs unverified skills — and a skill's frontmatter grants permissions

### LINKED ISSUE #727: The default template warns against itself on first run

### LINKED ISSUE #728: co browser do bills whoever started the daemon, not who ran the command

### LINKED ISSUE #729: An admin signature is not bound to the route it authorises

### LINKED ISSUE #732: The published balance is a startup snapshot, so a broke agent advertises credit it does not have

### LINKED ISSUE #733: CONNECTONION_BACKEND_URL redirects three modules out of nineteen, silently

### LINKED ISSUE #762: runtime_input never reaches a turn that runs no tools

### LINKED ISSUE #768: Signed subscribed profiles can be rolled back by replaying an older valid bundle
## Finding

The coordinated attestation work in oo-api #120 and Connect #743 proves that a fetched profile was signed by the pinned publisher. It does not prove that the profile is the newest one the publisher announced.

An untrusted or compromised relay can retain an older valid `{profile, profile_signature}` pair and later serve it with the matching old skill bodies. The subscriber reconstructs that old profile, verifies it successfully, and overwrites the local bundle. This can restore instructions the publisher removed or roll back a security fix without forging Ed25519.

The relay's outer ANNOUNCE timestamp prevents replay at ingestion, but it is not part of the portable `profile-v1` signature returned to subscribers. The profile API's `updated_at` is relay-authored and therefore cannot establish publisher freshness.

## Reproduction model

1. Publisher signs profile A.
2. Publisher later signs fixed profile B; subscriber installs B.
3. Relay serves the still-valid signed profile A and A's bodies.
4. `co sub sync` accepts A and replaces B locally.

## Decision needed

Choose a publisher-authenticated monotonic freshness scheme, for example:

- a signed sequence number with the subscriber persisting the highest accepted value; or
- a signed issued-at/version value plus an explicit reset/key-rotation policy.

The design must define first install, local state loss, publisher rollback intent, identity/key rotation, and multi-device publishing. A relay-authored timestamp alone is insufficient.

## 1.6.0 relationship

This was found during external-review preparation for #765. #743 should state its exact property as authenticity, not freshness. An external reviewer should explicitly decide whether authenticated rollback is accepted as a documented v1 limitation for 1.6.0 or must be fixed before release.

wu-changxing:
## Purpose

Final human-review candidate for ConnectOnion 1.6.0. This PR does not merge, tag, or publish 1.6.0.

## Candidate contents

- traceable, idempotent email sends and stable retry/error behavior
- per-project invite credentials, fail-closed unset onboarding, and no shared credential in copyable docs
- signed remote commands, route-bound admin signatures, authenticated owner-bound session status, and owner-bound browser billing
- publisher-attested skills plus monotonic profile-v2 rollback/equivocation rejection
- runtime-input delivery, backend URL consistency, dependency-aware skills, doctor repairs, safe quickstarts, and hardened release artifacts
- Microsoft OAuth credentials remain CLI-local; the backend receives no durable token or encrypted credential payload
- refreshed exact dependency lock with regression gates against stale metadata and patched-floor rollback
- React documentation points React users to the self-contained `@connectonion/react` package

## Final-review follow-up

Cross-language review found Python and TS/browser canonical JSON differed for Unicode and nested objects. The compatibility fix remains in the candidate. Current head `f88b29de55bd7dc792baef6995721d40811d08e6`:

- preserves existing Python escaped-Unicode signatures
- accepts TS/browser UTF-8 signatures
- accepts legacy TS top-level-only nested-object ordering during migration
- is covered by 59 focused host/auth protocol tests

The matching TS fix is connectonion-ts #33 head `14ee054`, which recursively canonicalizes new signed payloads.

## Current verification

- exact head: `f88b29de55bd7dc792baef6995721d40811d08e6`
- merged current `main` at `8bbc271` after #773 landed; candidate now includes the exact 1.5.19 release history
- version metadata and `uv.lock` agree on 1.5.19; focused release/version/timeout tests: 20 passed; loopback OAuth state test passed
- invite-code security suite: 25 passed; copyable docs no longer teach the retired shared `OpenOnion` credential
- GitHub Actions rerun is in progress for the documentation guard
- local focused auth/protocol suite: 59 passed
- sdist/wheel, Twine, installed-wheel acceptance, dependency floors, and forced artifact installation are covered by the candidate
- GitHub reports MERGEABLE/CLEAN; no review decision has been submitted

## Exercised patch baseline

`connectonion==1.5.19` is public and byte-verified against CI artifacts. PR #773 merged as `8bbc271`; this branch merged that main history in `1340b46`. Its complete matrix is being rerun before review approval.

Trusted Publishing remains externally misconfigured (`invalid-publisher`); do not weaken OIDC. Configure the exact publisher/environment before 1.6.0.

## Cross-repository rollout

- oo-api #129: merged as `94bbace`; deployed and healthy
- oo-api #131: head `c556936`, fixes operator Sent-mail lookup (#130), CI green and exact head deployed; awaiting review
- connectonion-ts #33: head `14ee054`, full CI green; signed SESSION_STATUS and canonical signing update
- connectonion-react #9: head `5c6c67a`, CI green; self-contained React SDK
- docs-site #32: head `679c9c1`, quality CI/Vercel rerunning; documents Python 1.5.19, SDK split, and CLI-local Microsoft OAuth
- Python metadata remains pre-1.6; one dedicated reviewed commit will bump version/changelog after all merges

## Human gate

No review decision has been submitted. Do not merge this PR or publish 1.6.0 until the refreshed matrix is green, independent review accepts the consolidated diff, and coordinated repositories are approved.

Closes #213
Closes #301
Closes #467
Closes #468
Closes #469
Closes #470
Closes #471
Closes #472
Closes #557
Closes #654
Closes #727
Closes #728
Closes #729
Closes #732
Closes #733
Closes #762
Closes #768

Supersedes #769 and the closed component PRs as the final Connect review surface.


wu-changxing:
Final automated verification is green at head 363cae22f5604d047ef565eed4561d6f08b05e14: Python 3.10/3.11/3.12/3.13, Windows browser 3.10/3.12/3.13, and Windows E2E all passed. GitHub reports CLEAN and MERGEABLE. This remains an external-review candidate; do not publish 1.6.0 from it yet.

wu-changxing:
Microsoft OAuth CLI-only update at 942af33: co auth microsoft creates an ephemeral Curve25519 key, polls an owner-bound one-time encrypted handoff, decrypts locally, and writes credentials only to local environment files. It no longer calls the retired backend status/credentials storage endpoints and no longer revokes a working local connection before replacement succeeds. Focused CLI suite: 9 passed. The earlier restricted full run had 5,927 passes; all 138 socket/browser permission failures passed when rerun with normal local permissions.

wu-changxing:
Design follow-up at 1bd821e removes backend polling entirely. The browser returns the sealed payload directly to a one-command 127.0.0.1 CLI listener; only the CLI decrypts and persists it. Outlook and MicrosoftCalendar now persist the newest locally rotated refresh token. Evidence: real loopback/CLI 10/10; Outlook+Calendar 48/48. Microsoft documents a 90-day default for this non-SPA flow and says refresh can return a replacement token; expiry/revocation still falls back to co auth microsoft.

wu-changxing:
Current-head local artifact rehearsal completed without upload. Built exactly connectonion-1.5.15.tar.gz and connectonion-1.5.15-py3-none-any.whl from 1bd821e; twine check passed for both and installed-wheel acceptance passed 8/8 from outside the source tree. Rehearsal-only SHA-256: sdist e28f43fa7c734498d76bf29f5d3f50a4712ab440ab477be75041c753a643289e; wheel 1cb9634418072d8397aae4f4367f064272e7a5ba0f6890cf22788d730f25b65a. These files are local evidence only and must not be uploaded as the final 1.6.0 artifacts.

wu-changxing:
Final-review finding fixed in 477e6c5: Python and TS/browser canonical JSON differed for Unicode and nested objects. Host verification now accepts existing Python signatures, current TS UTF-8 signatures, and legacy TS top-level-only sorting. Focused auth/protocol tests: 59 passed. Windows browser and packaged Windows E2E checks are green; the four full Python matrices are still running. This remains an unmerged 1.6 candidate.

wu-changxing:
Refreshed this candidate after #773 merged.

- current head: `1340b463d1c550d44f82fc7a25c99a0e82a78059`
- merge parents: prior candidate `477e6c5` + current main `8bbc271`
- version metadata and `uv.lock` now agree on 1.5.19
- focused release/version/timeout tests: 20 passed
- loopback Microsoft OAuth state test passed; credentials remain CLI-local
- Actions run: https://github.com/openonion/connectonion/actions/runs/31306131846

At this checkpoint, lockfile, all three native Windows browser jobs, and packaged Windows E2E are green. Python 3.10–3.13 full suites are still running. No merge or 1.6.0 publish requested.

wu-changxing:
Default invite credential follow-up at `f88b29d`:

- confirmed shipped policies contain no literal invite code
- confirmed `co init` and `co create` mint one per project and do not rotate/duplicate it
- confirmed unset `CO_INVITE_CODE` opens no invite door
- removed the retired public `OpenOnion` credential and default payment amount from the copyable host configuration
- added a regression test preventing that shared credential from returning to copyable docs
- focused invite/onboarding suite: 25 passed

Full Actions matrix is rerunning. No merge or 1.6.0 publish requested.

PR #780, opened 2026-08-09
Add resumable JSON mode to co ai

### LINKED ISSUE #333: co ai one-shot mode needs session resume + structured output, so Codex/Claude Code can drive it as a subprocess today
## Motivation

`co ai` already has a one-shot CLI mode — no new protocol needed for Codex or Claude Code to invoke it right now:

```python
# connectonion/cli/main.py:239-262 → cli/commands/ai_commands.py
co ai "Create a calculator agent"   # runs once, prints result, exits
```

Either can already shell out to this today through their own bash/shell tool. #332 proposes a full ACP agent-server mode for the reverse direction, but that's more than this needs — the same reasoning #295 already used for the *other* direction applies here:

> `co ai` can already technically run `codex exec "..."` through its `bash` tool. What it can't do is treat Codex as a collaborator: every call starts a fresh session, the output comes back as an unstructured blob, and there is no way to say "continue the thing you were just doing."

Swap the names and it's the same gap on `co ai`'s side.

## What's actually missing

Looking at `handle_ai()` (`connectonion/cli/commands/ai_commands.py`):

```python
if prompt:
    result = agent.input(prompt)
    print("\n" + result)
```

1. **No session resume.** Every `co ai "..."` call is a brand-new `create_agent(...)` — there's no `--resume`/session-id flag, so a calling agent can't continue a prior delegation across turns.
2. **No structured output.** `print("\n" + result)` is plain text, mixed with whatever Rich console output the agent already emitted. A calling agent (or a script) has nothing clean to parse — no session id, no exit/error signal, no usage info.
3. Approval already has an out: `--yolo` skips interactive prompts, so a subprocess invocation isn't stuck waiting on a TTY. Nothing to add here.

## Proposal

Give the one-shot path what #177 already wants from `codex`/`claude_code`: a session id in, a session id out, and a machine-readable envelope.

```bash
co ai "Fix the failing tests" --json
# {"session_id": "...", "result": "...", "error": null}

co ai "Now also update the docs" --resume  --json
```

- `--json`: emit a single JSON envelope (session_id, result, error) instead of `print(result)`, leaving Rich console output as stderr/interactive-only
- `--resume `: reuse the prior turn's session/messages instead of a fresh `create_agent()` — same continuity #295 asks `codex` for, mirrored here

## Relationship to other issues

- **#177 / #295**: the same need (session continuity + parseable output), opposite direction. `codex`/`claude_code` tools want it when `co ai` calls them; this issue is what Codex/Claude Code want when they call `co ai`.
- **#332**: the ACP agent-server direction. That's the right answer if we ever need real-time bidirectional streaming or a non-CLI ACP client to drive `co ai`. This issue is the much smaller, already-possible-today path — ship this first, revisit #332 if a plain subprocess call turns out not to be enough.

## Definition of done

- `co ai "task" --json` prints one JSON object with `session_id`, `result`, `error` — no other stdout noise
- `co ai "next task" --resume  --json` continues the same agent session (tool state / message history intact)
- `--yolo` still works combined with `--json`/`--resume`
- A test that invokes the one-shot path and asserts the JSON shape, without requiring a live LLM call

wu-changxing:
Closes #333.

## What

- Add `co ai <prompt> --json` with one stable stdout envelope: `{session_id,result,error}`.
- Add `--resume <session-id>` for multi-turn subprocess handoff.
- Persist versioned, atomic UUID-keyed snapshots under `~/.co/ai/sessions/`, preserving the full JSON-compatible Agent session and explicit TodoList state.
- Bind each session to its project cwd and hold a per-session OS lock across load → run → save.
- Re-arm an explicitly requested YOLO budget on resume.
- Exclude process-local background-task tools from JSON mode because their handles cannot survive subprocess boundaries.
- Document the machine contract, errors, permissions, and resumability limits.

## Design

This follows DD-010, DD-011, DD-017, DD-019, DD-020, and DD-021: progressive CLI disclosure, global user state under `~/.co`, explicit session/eval data, `Agent.current_session` as source of truth, deterministic machine output, and local human-readable atomic storage.

The issue discussion records the rejected alternatives and final boundaries:
- no ACP dependency for this milestone;
- no pickle or generic object serialization;
- no silent fresh session when resume fails;
- no false promise that process-local background handles persist.

## Validation

- Independent expert review: no remaining blockers.
- Related co-ai/ULW/CLI suite: **122 passed**.
- Full unit suite: **5,631 passed, 13 skipped**; the only failure is an unrelated external checkout mismatch (docs-site advertises 1.5.20 while this repo is 1.6.0).
- Packaging/project-directory checks: **42 passed**.
- `git diff --check`, Ruff on the new/changed clean-scope files, and Python 3.10 compile checks passed.

Draft only. Please review before merge.


wu-changxing:
This Draft is now rebased/retargeted onto #785 (`agent/hard-stop-216`) and has passed a fresh design + failure-injection review.

Two recovery contracts are now explicit regressions:
- a failed resumed follow-up may mutate its in-memory session/Todo state, but must leave the last disk snapshot unchanged;
- `os.replace()` is the final fallible commit step. The temp fd is set to 0600 before fsync/replace; there is no post-commit chmod that could make the CLI report failure after the new snapshot was already committed.

Review caught and fixed the post-replace chmod window before push. Final expert result: Blocker 0 / Medium 0 / Minor 0.

Verification:
- stacked JSON/resume/hard-stop focused suite: 134 passed
- final atomicity/CLI/ULW suite: 42 passed
- ruff + `git diff --check`: pass
- new head: c08f7c8d

Push used exact `--force-with-lease`; PR remains Draft and nothing was merged.

wu-changxing:
Merged after review. The resumable one-shot design is careful — canonical-UUID IDs (no path traversal), 0700/0600 perms, atomic mkstemp+fsync+os.replace, an exclusive flock/msvcrt lock per resume, cwd binding that refuses a resume from the wrong project, version + shape validation on load, and `restore_tool_state` that only rehydrates known `todolist` state rather than deserializing arbitrary objects. Wrapping the whole run in `redirect_stdout(sys.stderr)` so only the final envelope hits stdout is the right call.

One minor note, non-blocking: `save_snapshot` runs after `agent.input`, so a turn that completed but whose session isn't JSON-serializable (e.g. a plugin left a `set` like `shell_approved_cmds` in the session) reports `error` with `result=null`, discarding the completed output. Failing loud is defensible for a resumable contract, but if a serialization failure becomes common in practice, consider emitting the result alongside a 'not resumable' warning instead of dropping it.

PR #781, opened 2026-08-09
Add a resumable Claude Code tool

### LINKED ISSUE #398: Add a built-in Claude Code tool with session resume and explicit permissions
## Motivation

Issue #177 originally grouped Claude Code and Codex behind one proposed copyable-tool API. The Codex half has since shipped in PR #226 using the native `codex app-server` protocol, with session resume, live events, and per-action approval. Claude Code remains a separate provider integration with a different CLI protocol and permission model, so it should not stay hidden inside an otherwise completed issue.

## Scope

Add a copyable, function-based `claude_code` tool that:

- starts a headless Claude Code task in an explicit working directory
- returns a stable structured envelope including provider, session ID, status, result, and error
- accepts an explicit session ID to resume a previous Claude Code conversation
- makes its permission mode explicit and never silently selects the most permissive mode
- fails with a clear one-line result when the Claude Code binary is unavailable
- keeps provider-specific parsing behind the tool rather than changing the Agent loop

The public shape should stay close to the existing `codex` tool where the provider contracts genuinely overlap, without forcing both implementations through one leaky abstraction.

## Non-goals

- Wiring the tool into `co ai`; that remains #295.
- Exposing `co ai` as an ACP server; that remains #332.
- A generic abstraction covering every coding CLI before two implementations have proven the shared contract.
- Silent automatic permission escalation.

## Proposed API

```python
def claude_code(
    prompt: str,
    session_id: str = "",
    cwd: str = "",
    permission_mode: str = "default",
    model: str = "",
    timeout: int = 600,
    agent=None,
) -> str:
    """Run or resume Claude Code and return a structured JSON envelope."""
```

## Definition of done

- A first call returns a resumable Claude Code session ID.
- A second call with that ID continues the same session.
- Missing binary, timeout, authentication failure, and non-zero exit are represented clearly.
- Permission mode is explicit, validated, and covered by tests.
- Unit tests do not require a real Claude Code installation.
- A real-binary test is isolated behind an opt-in marker.
- Documentation explains the differences from the existing Codex app-server adapter.

## Related

- Split from #177.
- Codex mechanism already shipped in PR #226.
- `co ai` integration: #295.

wu-changxing:
Closes #398.

## Design

This adds Claude Code as a copyable built-in function tool while keeping provider execution separate from agent-loop orchestration, following DD-012 and DD-024.

- runs one headless CLI subprocess per turn with an explicit working directory
- returns a stable `provider/session_id/resumed/status/result/error` envelope
- resumes only when the provider returns the same non-empty session ID
- exposes explicit permission modes; `default` maps to the current `manual` CLI behavior and never silently escalates
- preserves normal Claude Code authentication and project instructions
- accepts both the documented JSON result object and the event-array form observed from the current CLI
- uses no shell, closes stdin, rejects Windows `.cmd`/`.bat` wrappers, and terminates the process tree on timeout
- keeps this adapter independent from `co ai`; a generic coding-CLI abstraction is deferred until another provider proves it useful

## Validation

- focused adapter, Codex compatibility, public import, and copy-command suite: **109 passed**
- complete unit suite before the final security refinements: **5651 passed, 13 skipped**, with one unrelated existing failure because the external `docs-site/lib/version.ts` still says `1.5.20` while this repository says `1.6.0`
- Ruff, `git diff --check`, and `compileall`: passed
- opt-in real CLI tests added but not run during final validation to avoid external model calls
- independent expert review: **0 blocker, 0 medium, 0 low**
- adversarial local process-tree probe: a grandchild ignored SIGTERM and held pipes open; timeout still returned within a bounded cleanup window with no descendant left behind

This is intentionally a Draft PR for maintainer review. It has not been merged.


wu-changxing:
This Draft is now rebased/retargeted onto #780 and has completed a fresh failure-path review.

Review caught and fixed two issues before push:
1. A failed resumed provider response could return a different valid session ID and drift the error envelope. Resumed envelopes now keep the requested ID authoritative on success and error; any mismatch is reported while preserving provider detail.
2. POSIX process-group probing could raise `PermissionError` on macOS and turn a real timeout into a launch error. Cleanup errors now degrade to best-effort parent terminate/kill without changing the public TimeoutExpired contract, and a real subprocess regression covers the POSIX path.

Docs now accurately promise best-effort Windows process-tree / POSIX launch-group cleanup and explicitly exclude deliberately detached descendants.

Final expert result: Blocker 0 / Medium 0 / Minor 0.
Verification: 50 adapter tests passed; broader focused suite 79 passed; ruff and `git diff --check` pass. New head: ad67686a. Push used exact `--force-with-lease`; PR remains Draft and nothing was merged.

Issue #511, opened 2026-07-31
[1.8] Default paid WTF Browser and explicit free Chrome


wu-changxing:
# Outcome

Keep one ConnectOnion browser API while making paid **WTF Browser** the default and preserving explicit free Chrome use. ConnectOnion is a controller; the standalone App owns browser billing and renewal.

## Canonical 1.8 choices

| requested choice | behavior |
|---|---|
| omitted / `wtf` | Start or attach to paid WTF Browser. Fail with a typed action if unavailable; never silently fall back. |
| `chrome` | Explicit compatibility mode using installed/system Chrome; no paid API, artifact, session or charge. |
| legacy `onion` / `auto` | Resolve to `wtf` for migration only. |
| legacy `system` | Resolve to `chrome` for migration only. |

Price is `$0.025 / 15 minutes` for one running WTF Browser runtime. Tabs/controllers do not multiply charges. Attach/detach does not own the paid session; final browser close releases once.

## Browser/API prerequisites — complete

- [x] WTF Browser `.222` macOS/Linux artifacts and Onionwright dev4 are immutable-published and catalogued.
- [x] Direct human App start/renew/release and separate Onionwright paid launch/release pass against the real Preview ledger.
- [x] oo-api #213–#215 are deployed at `a6f782db14bab0fb0b2a92cab21e34d093559ff3`.
- [x] The 2026-09-02 same-owner attach attempt reached native approval without creating a second paid session; the sequence-0 session released before renewal.

## ConnectOnion implementation

Draft PR #1353 implements the canonical `wtf|chrome` selection, persistent preference, no-fallback behavior, exact Onionwright consumer and package/docs/tests for proposed `1.8.0a5`.

Current PR state: Draft, 22/22 checks complete with no failures, but `CONFLICTING/DIRTY` against main. Its selected dependency is Onionwright `0.0.13.dev5`, which is separate from the already-shipped dev4 Browser Preview and still needs exact artifact/catalogue evidence.

## Remaining before closing

- [ ] Rebase/reconcile #1353 without losing current main changes.
- [ ] Publish/redownload/catalogue the exact dev5 dependency if a5 remains the selected cut.
- [ ] Finish the unlocked same-owner Allow → detach acceptance in openonion/browser#109.
- [ ] Complete installed-wheel/hosted lifecycle and docs-site gates, then merge #1353.

Release control: #1154. Preview consumer: #1315. Browser runtime: openonion/browser#109.

wu-changxing:
[...]

PR #782, opened 2026-08-09
browser: choose the engine from the attested tier, and never withhold one

wu-changxing:
The decision half of #511. Server half is openonion/oo-api#134 (CI green), which is why the tier is now something a client can trust.

    free   patchright + the Chrome already on the machine
    pro    the onion browser, a patched Chromium

## The rule this is built around

**Nothing here may stop a browser from starting.**

That is not caution, it is preserving measured behaviour. E2E on a clean Ubuntu 22.04 box with no `co` installed:

```
pip install connectonion   ->  co 1.6.0, patchright 1.61.2, onionwright absent
co browser --headless go_to https://example.com
                           ->  Navigated to https://example.com/
ps -eo args | grep chrome  ->  /opt/google/chrome/chrome
take_screenshot            ->  21606 bytes of real PNG
```

A default install already drives system Chrome through patchright, with **no licence call anywhere on that path**. It cannot be blocked today by billing, an expired attestation, or an oo-api outage. Adding tiering must not hand a free user a new way to fail.

So every branch ends in an engine:

| input | engine | note |
|---|---|---|
| no attestation | patchright | silent — offline or no account is not a sales moment |
| `tier=free` | patchright | "unlocks at $10.00" |
| `pro`, not active | patchright | "balance empty" |
| `pro`, binary absent | patchright | "not installed yet" |
| `pro`, present | onion | silent |

## The branch I did not expect to need

`onionwright` is **not** a dependency of `connectonion` — verified on the same box, it is not in `pip list`. So a paying customer's first launch has a pro attestation and nothing on disk to run. Treating that as an error would leave the person who paid worse off than the person who did not, which is why it falls back and says so rather than raising.

A malformed payload reads as *no licence* on purpose: guessing `pro` tries to launch something that may not exist; guessing `free` just works.

## Tests

27, written before the module and shown red (collection error — no such module). Six name a branch. The rest is a parametrised sweep over every tier × active × present combination asserting only that *an engine comes out*, because that property is worth more than any individual branch.

```
27 passed          tests/unit/test_browser_engine_choice.py
225 passed         tests/unit -k browser
```

## Not in this PR

Wiring `choose()` into the launch path (`installed_browser_path` / `launch_persistent_context`), fetching the binary when a pro user lacks it, and the three-stage e2e (free -> fund to $10 -> pro -> spend down -> demote). Kept separate so the decision can be reviewed on its own — it is the part that is pure, and the part where a mistake is silent.

https://claude.ai/code/session_01YE2E7YwTvLnPp2qjfcofRq

### DIFF
--- a/connectonion/useful_tools/browser_tools/engine.py
+++ b/connectonion/useful_tools/browser_tools/engine.py
@@ -96,10 +96,31 @@ def choose(attestation, onion_present):
     return ONION, None
 
 
-def _cached_attestation():
-    """The licence this machine already holds, or None. Never raises.
+# The paid engine is not reachable from here yet, and this names exactly what
+# is missing rather than leaving a stub that reads as finished.
+#
+# `onionwright.licence.license.load(path, server_key)` needs a cache path and
+# the pinned server key; `onionwright.launcher.resolve_binary(token,
+# server_key, pin)` needs an authenticated token and a Chromium revision.
+# connectonion holds none of those three today. Supplying them is the open
+# work in openonion/connectonion#511 — not something to fake here.
+#
+# Until then the defaults below raise, `resolve()` reads that as "no licence"
+# the same way it reads an offline machine, and everyone gets the free engine.
+# That is the right failure direction and the wrong end state, so
+# test_the_paid_path_is_not_wired_up_yet asserts it out loud: when the wiring
+# lands, that test fails and has to be deleted, which is the point of it.
+PAID_WIRING_PENDING = (
+    "onionwright is installed but connectonion has no licence configuration "
+    "yet (cache path, pinned server key, Chromium revision) — see "
+    "openonion/connectonion#511"
+)
 
-    The import is the point. `onionwright` is not a dependency of
+
+def _cached_attestation():
+    """The licence this machine already holds, or None.
+
+    The import is load-bearing. `onionwright` is not a dependency of
     connectonion, so on a default install this raises ImportError immediately
     and we are done -- no token read, no HTTP, no clock check. A free machine
     reaches the network exactly as often as it did before tiering existed,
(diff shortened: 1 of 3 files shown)


wu-changxing:
## Self-review: two defects in my own code, both hidden by the tests passing

Re-read against `onionwright@main` rather than against my memory of it. Both defects fail *safe* — they degrade to the free engine — which is exactly why 44 green tests did not catch either.

### 1. `_onion_path()` calls an API that does not exist

```python
def _onion_path():
    from onionwright import launcher
    return getattr(launcher, "installed_path", lambda: None)()
```

There is no `launcher.installed_path`. Checked:

```
$ grep -c "def installed_path" src/onionwright/launcher.py
0
```

The real surface is `resolve_binary(token, server_key, pin, home=None) -> Path` and `launch(playwright, pin, ...)`. My `getattr(..., lambda: None)` default means this returns `None` on every machine, including one with the paid package installed — so `onion_present` is always false and **the pro branch is unreachable in production**.

### 2. `_cached_attestation()` is a stub that raises

```python
raise NotImplementedError("reading the cached attestation is wired up with …")
```

`resolve()` catches every exception from the loader and reads it as "no licence", so this too resolves to free — on a machine that has paid, has a valid cached attestation, and has the binary.

### Why the test suite is green anyway

Every test injects both lookups:

```python
engine.resolve(load_attestation=lambda: {...}, onion_path=lambda: "/opt/onion/chrome")
```

so the production defaults are never executed. The parametrised sweep asserts *an engine always comes out* — true, and still true with both defaults broken, because broken defaults produce the free engine. I proved the property I wrote the tests for and not the one that matters commercially: **a paying customer gets the thing they paid for.**

That is the same class of mistake as the two defects this whole change exists to fix. `/license/download` computed an attestation and ignored it; `attest()` accepted a tier and never checked it. Both were code that looked like it was doing the work. So was this.

### What I am changing

- Delete `_onion_path`'s guessed API. Either it calls `resolve_binary` with real config or it does not exist — a `getattr` fallback that silently answers "absent" is worse than an import error, because an import error is visible.
- The defaults need a token, the server public key, and a Chromium pin. connectonion does not have that config yet, which is the actual open work — not something to paper over with a stub that reads as finished.
- A test that fails while the production default cannot reach the pro branch, so the gap is visible in CI rather than in this comment.

Until that lands, the honest description of this PR is: **the policy is correct and tested; the wiring to the paid package is not done.** The commit message and PR body both need to say that in those words rather than implying `resolve()` is ready.

wu-changxing:
## E2E on Linux against this branch — the free path is unchanged

Ubuntu 22.04, fresh venv, installed straight from the PR branch.

```
$ pip install "git+https://github.com/openonion/connectonion.git@feat/browser-engine-select"
$ co --version                      co 1.6.0
$ pip list | grep -iE 'onionwright|patchright'
patchright               1.61.2     # onionwright still absent, as intended
```

The decision, run on that machine:

```
resolve()             -> patchright
note                  -> (none)
_cached_attestation   -> ModuleNotFoundError
_onion_path           -> ModuleNotFoundError
```

Both production lookups die at the import on a machine without the paid package, which is the design — no token read, no socket, no way for licensing to interfere with a free user's browser. `note` is `None` rather than a sales line, because an install with no account is not a moment to sell into.

And the browser still starts, byte-identical to the baseline captured before any of this existed:

```
co browser --headless go_to https://example.com   ->  Navigated to https://example.com/
ps -eo args | grep chrome                         ->  /opt/google/chrome/chrome …
take_screenshot                                   ->  21606 bytes   (baseline: 21606)
```

### One false alarm, and how it was separated

The first attempt failed with `TargetClosedError: Page.goto: Target page, context or browser has been closed`. That was mine: I had run `pkill chrome` in the same command, which left the daemon holding a dead browser.

Rather than assume, I ran the **released 1.6.0 install on the same box under the same conditions** as a control — it navigated fine, then the branch install navigated fine too once the daemon was closed cleanly with `co browser close` instead of killed. So the failure was the harness, not the change. Worth writing down because "my e2e failed, therefore my code is broken" and "my e2e failed, therefore my e2e is broken" look identical until you run the control.

PR #778, opened 2026-08-09
Replace co ai plan mode with progressive todos

wu-changxing:
## What changed

- Remove `enter_plan_mode`, `write_plan`, and `exit_plan_and_implement`.
- Remove the associated implementation, prompts, reminder, help text, default permission, and obsolete documentation.
- Use TodoList for genuinely multi-step work while handling simple work directly.
- Remove `plan` from the main approval modes.
- Normalize legacy frontend requests and persisted `plan` sessions to `safe`, with an explicit frontend acknowledgement.
- Preserve the isolated read-only planning sub-agent.
- Remove stale test imports and make subprocess probes fail loudly.

## Why

The previous workflow imposed a state machine, plan files, read-only restrictions, and an approval ceremony on every agent-creation task. TodoList provides visible progress without making that complexity an entry requirement.

Deleting only the three tools would be incomplete because older clients could still enter a plan state with no available exit path. This PR removes that state end to end.

## Design references

- DD-010: Progressive Disclosure
- DD-011: Complexity proportional to ambition
- DD-020: Hard limits in code, soft policy in prompts
- DD-024: Sub-agent isolation

## Validation

- Focused changed-area suite: 110 passed.
- Full unit suite: 5611 passed, 13 skipped.
- The only full-suite failure is an unrelated external checkout mismatch: `docs-site` advertises 1.5.20 while ConnectOnion is 1.6.0.
- Post-review documentation and permission suite: 60 passed.
- Post-rebase focused suite: 28 passed.
- `git diff --check`: passed.
- `compileall`: passed.
- Independent review: no blocking findings.

Closes #121

PR #783, opened 2026-08-09
Let co ai delegate work to Codex

### LINKED ISSUE #295: [FEATURE] Let co ai hand work to Codex instead of reimplementing it
## Motivation

`co ai` can already *technically* run `codex exec \"...\"` through its `bash` tool. What it can't do is treat Codex as a collaborator: every call starts a fresh Codex session, the output comes back as an unstructured blob, and there is no way to say \"continue the thing you were just doing.\" So in practice `co ai` never delegates — it does everything itself, even work another agent on the machine is better at.

We also teach Codex. It is odd that our own coding agent has no first-class way to use it.

## The precedent already exists in this repo

We just made exactly this call for the browser. `co ai` and the templates no longer carry an in-process `BrowserAutomation`; they drive the `co browser` CLI (#234). The reasoning applies unchanged here:

> one daemon owns the profile and the tabs, so panels no longer share a page and 40 tool schemas leave the request

Same shape: don't reimplement a coding agent inside `co ai`, drive the one the user already installed.

## What's actually missing

`codex-cli 0.145.0` exposes everything needed — this is a wiring problem, not a capability gap:

| Need | Codex already provides |
|---|---|
| Continue a delegation across turns | `codex exec resume <id>` / `--last` |
| Target a specific repo | `-C, --cd <DIR>` |
| Parse what happened | `--json` (JSONL events) |
| Constrain the final answer | JSON Schema for the model's final response |

So the work is: capture the session id, thread it through `co ai`'s turns, and surface progress.

## Proposal

Add a `codex` tool to `co ai` (`connectonion/cli/co_ai/tools/`) that wraps the CLI and keeps the session:

```python
def codex(prompt: str, session_id: str = \"\", cwd: str = \"\") -> str:
    \"\"\"Delegate a coding task to Codex. Pass session_id to continue a previous
    delegation. Returns the result and the session id to resume with.\"\"\"
```

`co ai` stays the planner and reviewer: it decides what to delegate, hands over a scoped task, reads the result, and keeps ownership of the plan and the todo list.

## Relationship to #177

**#177 is the library-level feature** — copyable `claude_code` / `codex` tools any ConnectOnion agent can import. This issue is the `co ai` side of it, and it carries questions #177 doesn't:

- **Approval**: `co ai` has `tool_approval`. Codex has its own approval model. Two nested prompts for one action is a bad experience — decide which layer asks, and make the other trust it. Related: #269 (auto-approval governance).
- **Long runs**: a Codex delegation can outlast a turn. Does it go through `run_background` / `task_output`, like other long work in `co ai`?
- **Not installed**: degrade with a clear message, not a stack trace. `co doctor` should probably report it (#213).
- **Skills**: we already fan the same SKILL.md out to Claude/Codex/Cursor. If `co ai` delegates to Codex, the skill the user asked for should travel with the task.

## Open questions

1. **Delegate, or run on it?** This issue proposes delegation (Codex as a tool). The other reading — `co ai` using Codex as its *model backend* — is a much larger change and probably belongs in the ACP direction of #239. Worth stating which one we mean before anyone builds.
2. Should `claude_code` land at the same time, or is Codex first deliberately?
3. Does the delegated task inherit `co ai`'s plan-mode state, or start clean?

## Definition of done

- `co ai` can hand a scoped task to Codex and continue that same Codex session on a later turn
- Exactly one approval prompt for one delegated action
- Missing `codex` binary produces a one-line explanation
- Docs page + a test that does not require the real binary

wu-changxing:
Closes #295.

## What changed

- Registers a thin co-ai Codex delegation tool with an explicit working directory and resumable session ID.
- Maps the current co-ai mode onto Codex policy:
  - Safe: read-only + manual operator approval
  - Accept Edits: workspace-write + manual operator approval
  - YOLO/ULW: workspace-write + deny unexpected approval callbacks
- Confines hosted non-admin requesters to read-only + deny, regardless of their requested mode.
- Reapplies sandbox, model, and approval policy when a thread is resumed; never selects danger-full-access.
- Uses method-specific Codex app-server responses and shows the concrete command, file root/scope, or permission profile to the operator.
- Makes automatic mode rely on app-server's `approvalPolicy=never` for work inside the selected sandbox and fail closed on every unexpected v2/legacy approval callback.
- Shares one end-to-end timeout budget, performs bounded process-tree cleanup, captures bounded stderr diagnostics, and fails pending/future requests after EOF.
- Removes the obsolete co-ai Plan wrapper/policy while retaining the separate planning sub-agent.

## Design basis

The issue discussion records the final design after reviewing DD-010, DD-012, DD-024 and the approval governance from #269: keep a thin provider wrapper, let the active co-ai mode own policy, preserve the operator-only approval boundary, and avoid a duplicate outer prompt.

## Verification

- 178 focused approval, Codex, co-ai, JSON-resume, hard-stop, and packaging tests passed.
- Ruff and `git diff --check` passed.
- Two independent final expert reviews: Blocker 0 / Medium 0 / Minor 0.
- The earlier full stacked release rehearsal passed 5,796 unit tests; its only intentional external failure was the unreleased docs-site version mismatch tracked by #792.
- No real Codex model call was used for verification.

Draft only. Please review; do not merge automatically.


wu-changxing:
Integration rehearsal follow-up: removed the obsolete co-ai Plan-mode mapping, tests, and docs so this Draft composes with #778 instead of restoring a deleted workflow. The resume-policy test now proves the current Safe → YOLO/ULW policy is reapplied across the same Codex session.

Local author review:
- `git diff --check`: pass
- focused co-ai/Codex suite: 31 passed
- combined 1.7 rehearsal (all ten Draft heads): 5,796 passed, 13 skipped; the sole failure is the intentional cross-repo version guard because the current docs-site checkout still advertises 1.5.20 while the package is 1.6.0

Commit: 6361966f. PR remains Draft; nothing was merged.

wu-changxing:
Rebased this Draft PR onto #781 (`agent/claude-code-tool-398`) and revalidated the stacked Codex integration.

Review found and fixed two approval-boundary issues:
- hosted non-admin requesters are now confined to `read-only + deny`; only the operator/admin (or a local session with no hosted requester) can answer nested Codex approval prompts;
- `approval="auto"` relies on `approvalPolicy="never"` for work inside the selected sandbox and fails closed on every unexpected v2/legacy approval callback, so a callback cannot redefine filesystem/network scope.

Also preserved the `background_tools` conditional, removed only the obsolete co-ai Plan wrapper/policy, and retained the separate planning sub-agent.

Validation:
- 178 focused approval/Codex/co-ai/JSON-resume/hard-stop/packaging tests passed
- Ruff passed; `git diff --check` passed
- two independent final reviews: Blocker 0 / Medium 0 / Minor 0

This PR remains Draft and has not been merged.

wu-changxing:
Merged after review. The approval hardening is the right call — `auto`/`deny` fail closed on any callback because `approvalPolicy=never` already permits sandboxed work, so a callback is an unexpected boundary expansion. And the policy layering is correct: the co ai wrapper (`cli/co_ai/tools/codex.py`) derives sandbox/approval from the session mode rather than exposing them to the planner, and forces hosted non-admin requesters down to read-only/deny — so danger-full-access and auto are never reachable through delegation. Per-method response shapes and the end-to-end timeout budget both look right.

Minor, non-blocking observation: after this change `auto` and `deny` are behaviorally identical in the low-level tool (both use `approvalPolicy=never` and both return False from `_approval_allowed`). That's fine for the co ai path (which only uses manual/deny), but a direct library caller might expect `auto` to mean 'approve every callback'. Worth a doc line clarifying that `auto` = 'run within the sandbox, deny escalations' if you keep both names.

PR #785, opened 2026-08-09
Hard-stop in-flight LLM calls and tools

### LINKED ISSUE #216: Hard stop: land INTERRUPT in ~300ms during in-flight LLM calls and tools (abandonable-step wrapper)
# Hard Stop for the Agent Loop

## Problem (current graceful-stop gap)

The graceful stop shipped in #188 / openonion/oo-chat#24 works, but it only lands at an **iteration boundary**:

- The INTERRUPT frame reaches the agent-visible mailbox in <1ms: `network/host/ws_router/session.py:113-116` → `WebSocketIO.send_to_agent()` (`network/io/websocket.py:102-106`), running on the asyncio thread which is never blocked by the agent.
- But the sync agent thread only *reads* the mailbox at event boundaries: `poll_interrupt` (`useful_plugins/tool_approval/approval.py:721-735`, `@after_iteration`) drains it via `receive_all('INTERRUPT')` and sets `stop_signal`, consumed at `core/agent.py:439-441`.
- Nothing checks the mailbox during the two blocking windows:
  1. **`self.llm.complete(...)`** at `core/agent.py:471` — synchronous, non-streaming (`core/llm.py:8`), no cancellation hook, SDK timeouts up to 600s (`llm.py:1009-1014` for OpenOnion). A 30s+ completion blocks the stop for its full duration.
  2. **`result = tool_func(**tool_args)`** at `core/tool_executor.py:195` — plain sync call, no timeout, no kill mechanism. A minutes-long browser/shell tool blocks the stop for minutes.
- Nothing drains INTERRUPT **between tools in a batch** either — the mid-batch `stop_signal` check at `tool_executor.py:57-63` only fires when an approval rejection sets it (`approval.py:583`).
- The **no-tool_calls exit path** (`core/agent.py:427-430`) returns without ever running `after_iteration`, so an INTERRUPT during the final LLM call is never drained at all.
- An INTERRUPT arriving while a tool is blocked in `io.receive()` (approval wait, `approval.py:552`; `ask_user`) is **swallowed as the response** — `receive()` pops the first message regardless of type (`websocket.py:64-69`). The approval path stops "by accident" (`approved=False` → `reject_hard`); `ask_user` misreads it as an answer.
- The frontend has **zero feedback** for a sent INTERRUPT: `interrupt()` (`oo-chat/components/chat/use-agent-sdk.ts:263-266`) fires the frame and nothing changes locally; the Stop button (`chat-input.tsx:365-378`) appears dead until the in-flight step finishes.

Net effect: clicking Stop during a long LLM call or tool does nothing visible for 30s–minutes.

## Goals / Non-goals

**Goals**
- Stop lands end-to-end in **≤ ~300ms** in every agent state: mid-LLM-call, mid-tool, between tools, waiting on approval/ask_user, and on the no-tool_calls final call.
- `current_session['messages']` remains **valid for the next turn** on every stop path: no dangling assistant `tool_calls` without `role:"tool"` results, Gemini-3 `extra_content`/thought_signature echoes preserved.
- Reuse the shipped stop machinery unchanged: `stop_signal` → `on_stop_signal` → `"What would you like me to do?"`, the mid-batch result substitution at `tool_executor.py:57-63`, and normal return through `input()` so persistence (`http_router.py:68-74`), reconnect, and the OUTPUT wire contract are untouched.
- Works for **all 8 LLM providers** identically, including `co/*` managed keys.
- Backward compatible both directions: old frontend + new backend gets faster stops for free; new frontend + old backend degrades to today's graceful stop.

**Non-goals**
- Killing arbitrary in-flight Python. Not safely possible in CPython (locks, sockets, half-written files); `PyThreadState_SetAsyncExc` cannot even land during a GIL-released socket read. The semantic is **"stop stops the agent, not the world"**: an abandoned tool's side effects complete in the background; its result is discarded.
- LLM streaming. It is the right long-term mechanism for true server-side cancel (and is already planned, `llm.py:127`), but it rewires the happy path of every LLM call across 8 providers with unverified parity risks (OpenOnion proxy SSE passthrough, Gemini-3 `extra_content` on deltas). It slots in later behind the same call-site contract introduced here — it must not be smuggled into a stop fix.
- A two-stage graceful/force UX. Since nothing is killed in either stage, "force" and "graceful" abandonment have identical backend semantics; a second click + grace window is pure friction. One click, one stop.

## Recommended architecture

**Core mechanism (one idea):** the INTERRUPT is already in the mailbox instantly; the entire problem is that the agent thread can't look at it while blocked. So run each blocking step (`llm.complete()`, `tool_func()`) on a **disposable daemon worker thread** while the agent thread polls the mailbox every 200ms. On interrupt: abandon the worker (result discarded), set `stop_signal`, and let every already-shipped consistency mechanism do the rest. ~30 lines, zero protocol changes required, all providers covered uniformly.

Phased:

- **Phase 1 — backend hard stop (ships alone, ~70 LOC):** `run_interruptible` helper + LLM and tool call-site wraps + interrupted-trace closure + mandatory INTERRUPT handling in blocking `receive()` gates. Stop latency drops from "whenever the step finishes" to ≤ ~300ms with **no frontend or protocol changes** — the existing terminal OUTPUT ("What would you like me to do?") arriving within ~0.5s *is* the feedback.
- **Phase 2 — polish (~60 LOC across repos):** `INTERRUPT_ACK` router echo + registry-based INTERRUPT routing (fixes the pre-CONNECT silent drop), frontend "Stopping…" state, cooperative `should_stop` for built-in long tools (shell `Popen.terminate()`, browser between-action checks), best-effort LLM socket abort.
- **Future (separate issue):** streaming `complete()` with between-chunk cancel for true server-side abort and token savings. Drop-in at the `(response|None, interrupted)` contract.

### Phase 1 — backend

**New: `connectonion/core/interrupt.py` (~30 LOC)**

```python
def run_interruptible(fn, io, poll_seconds=0.2):
    """Run fn to completion, or abandon it if the user sends INTERRUPT.
    Returns (result, False) normally, (None, True) on interrupt."""
    if io is None or not hasattr(io, 'receive_all'):
        return fn(), False                # local/CLI agents: direct call, zero threads, zero overhead
    if io.receive_all('INTERRUPT'):       # pre-spawn check: interrupt already pending -> never start the step
        return None, True
    box = {}
    def body():
        try: box['result'] = fn()
        except BaseException as e: box['error'] = e
    worker = threading.Thread(target=body, daemon=True)
    worker.start()
    while worker.is_alive():
        worker.join(timeout=poll_seconds)
        if io.receive_all('INTERRUPT'):
            return None, True             # abandon worker; its result is discarded
    if 'error' in box: raise box['error'] # exceptions re-raise on the agent thread — error→retry path preserved
    return box['result'], False
```

Notes:
- Plain daemon `Thread`, **not** `ThreadPoolExecutor` — since 3.9, executor threads are non-daemon and atexit-joined, so an abandoned 10-minute worker would hang interpreter shutdown.
- `receive_all('INTERRUPT')` (`websocket.py:71-86`) is a non-blocking *selective* drain: mode_change / prompt_update / runtime-input frames are untouched.
- The pre-spawn check closes the window where a stop clicked just before a tool starts would still launch it.
- If the worker finishes in the same 200ms window as an interrupt, the completed result is returned (work is done, don't waste it); the queued frame is drained by the next wrapper call or `poll_interrupt`. Pin this race with a test.
- Verified safe for `@xray` tools: xray context is a process-global singleton with plain attributes (`connectonion/debug/xray.py:48-58`, `builtins.xray = self`), not thread-local — the worker sees the injected context.

**`connectonion/core/agent.py` (~12 LOC)** — in `_get_llm_decision()`, wrap line 471:

```python
response, interrupted = run_interruptible(
    lambda: self.llm.complete(self.current_session['messages'], tools=tool_schemas), self.io)
if interrupted:
    self._record_trace({'type': 'llm_result', 'id': llm_id, 'status': 'interrupted', ...})
    self.current_session['stop_signal'] = 'user_interrupt'
    return None
```

In `_run_iteration_loop()` after line 425: `if response is None:` → pop `stop_signal`, fire `on_stop_signal`, return `"What would you like me to do?"` — the existing exit path verbatim.

The closing `llm_result` trace entry with `status:'interrupted'` is **required**, not optional: the frontend's thinking indicator is keyed on the streamed `llm_call` trace (`agent.py:462-468`) and `hasActiveUI`/`isLoading` (`use-agent-sdk.ts:136-152, :189`) stays stuck true without it. This also covers the no-tool_calls gap: the wrapper runs on every LLM call, including the last one.

**`connectonion/core/tool_executor.py` (~8 LOC)** — in `execute_single_tool()`, wrap line 195:

```python
result, interrupted = run_interruptible(lambda: tool_func(**tool_args), agent.io)
if interrupted:
    agent.current_session['stop_signal'] = 'Interrupted by user'
    trace_entry.update(result='Interrupted by user', status='interrupted',
                       timing_ms=(time.time() - tool_start) * 1000)
    agent._record_trace(trace_entry)      # resolves the 'running' tool_call chip in the UI
    return trace_entry
```

The already-shipped machinery then takes over unchanged: `execute_and_record_tools` sees `stop_signal` at `tool_executor.py:57-63`, substitutes the stop string as this tool's result message, synthesizes `"Rejected by user"` results for every remaining tool_call id, and breaks. The pre-spawn check in the wrapper also gives us the **between-tools** stop for free — no separate `before_each_tool` poller needed (which could not actually prevent tool execution outside the tool_approval plugin anyway: `tool_executor.py:181→195` runs `tool_func` regardless).

**Blocking-receive gates — mandatory hardening (~10 LOC).** `check_approval` (`approval.py:552-560`), `ask_user` (`useful_tools/ask_user.py:33-46`), and `DiffWriter` (`useful_tools/diff_writer.py:234`) must handle a popped `{'type':'INTERRUPT'}` explicitly, exactly like the existing synthetic `io_closed` (`approval.py:555` precedent): set `stop_signal='user_interrupt'` and return/raise a cancelled result. This turns the approval path's accidental stop into an intentional one and fixes `ask_user` misreading INTERRUPT as an answer.

**Compatibility.** `poll_interrupt` stays as-is — it becomes a redundant iteration-boundary backstop (mailbox usually already drained; harmless). INTERRUPT keeps flowing through the mailbox, so **no third-party `receive_all('INTERRUPT')` consumer breaks**. `input_handler` returns normally → JSONL record saved `status="done"` (`http_router.py:70-74`), no stuck-"running" sessions, no changes to checkpoint/merge/reconnect. Fresh `WebSocketIO` per INPUT (`agent_io.py:122`) already guarantees no stale-interrupt leakage across turns.

### Phase 2 — polish

- **Router (`ws_router/session.py`, ~8 LOC):** explicit `elif msg_type == "INTERRUPT":` branch above the `:113` fallthrough. Resolve the target io via the **`ActiveSessionRegistry` lookup by `session_id`**, not just the connection-local `active_io` — fixes the current silent drop when a reconnected/refreshed socket (relay path especially) sends INTERRUPT before its own CONNECT/INPUT. Still `send_to_agent(data)` (mailbox delivery unchanged), plus reply `{type:'INTERRUPT_ACK', session_id}` inline — precedent: `RUNTIME_INPUT_ACK` (`session.py:99-107`) and inline `SESSION_STATUS` (`session.py:51-56`). No running session → ACK `{status:'not_running'}`, frame dropped (never buffered into a future turn).
- **Frontend (`oo-chat`, ~20 LOC):** `interrupt()` sets a local `isStopping` flag (optimistic, 0ms); Stop button renders as a disabled "Stopping…" spinner; cleared when `isProcessing` drops (OUTPUT/ERROR — now ~0.5s away) or by a ~3s fallback timer (covers old backends and the disconnected no-op at `use-agent-sdk.ts:262`). ACK handling is additive if/when the TS SDK surfaces unknown frames — the fallback-timer version ships value regardless and must not block on SDK changes.
- **Cooperative stop for built-in long tools (~40 LOC, tools layer, zero core):** the wrapper additionally sets an `agent`-visible stop event exposed as `xray.should_stop`; the shell tool switches to `Popen` + 0.5s `wait(timeout)` polling and `terminate()`s the process group on stop (subprocesses, unlike threads, die); browser tools check between page actions. Opt-in convention, documented for third-party tools as: *interrupt = abandon, not kill*.
- **Best-effort LLM socket abort (~15 LOC, optional):** on abandoning an LLM worker, swap `self.client` to a fresh SDK client and `close()` the old one. Sold strictly as best-effort cleanup: closing an fd another thread is blocked in `recv()` on is **not** guaranteed to wake it on Linux, and pool-close-aborts-in-flight is semi-internal httpx/httpcore behavior. Correctness never depends on it — the worker is disposable either way, bounded by the ≤600s SDK timeout. The client swap also keeps reused-Agent library embedders working on the next turn; needs a per-provider unit test if we ship it.

## Message-consistency handling

Every stop path ends in one of two states, both already proven valid in production:

1. **Interrupted mid-LLM:** nothing from the aborted call is appended (all appends happen caller-side after `complete()` returns — `agent.py:429`, `tool_executor.py:40`). `messages` ends at the last tool result or the user message — byte-identical to what the shipped graceful stop leaves (`agent.py:439-441` appends nothing). Valid for OpenAI-format providers and for `AnthropicLLM._convert_messages` (`llm.py:414-516`); the next turn's `role:"user"` append (`agent.py:295-302`) is legal from either state. Partial output: none exists (non-streaming); the interrupted `llm_result` trace entry keeps observability.
2. **Interrupted mid-tool / mid-batch:** the assistant message with ALL tool_calls was appended at batch start (`tool_executor.py:40`); the shipped substitution at `tool_executor.py:57-63` writes `"Interrupted by user"` as the abandoned tool's result and synthesizes `"Rejected by user"` results for every remaining id — every `tool_calls[i].id` gets a `role:"tool"` message, and the Gemini-3 `extra_content` echo (`tool_executor.py:253-276`) is untouched. This code has been in production since the graceful stop shipped.

The abandoned worker can never corrupt the session: message appends are caller-side only, its boxed result is discarded, and on the server path its io is superseded per-INPUT (`agent_io.py:122`; `close()` blocks late sends, `websocket.py:94-96`). Because we never kill the agent thread, no repair/reconstruction machinery is needed anywhere.

## Alternatives considered

- **Streaming `complete()` with between-chunk cancel:** right long-term, wrong vehicle — rewires the happy path of every LLM call across 7-8 providers with unverified parity risks (OpenOnion SSE passthrough, Gemini-3 `extra_content` on deltas), and still has no answer for the in-flight non-cooperative tool, which is half the stated problem. Deferred to its own issue; slots in behind the same call-site contract.
- **`PyThreadState_SetAsyncExc` / signals:** the definition of a clever trick — cannot land during GIL-released socket reads (the case we care about most), can corrupt `finally` blocks; signals only reach the main thread and agents run on daemon threads. Rejected.
- **Event-only cooperative checks (no worker):** no check point exists *inside* `llm.complete()` or `tool_func()`; insufficient alone.
- **Rerouting INTERRUPT out of the mailbox into an Event:** breaks every existing `receive_all('INTERRUPT')` consumer and needs shims in approval/ask_user/diff_writer; the mailbox is already instant, so the Event buys nothing Phase 1 needs. Rejected; keep mailbox delivery.
- **Two-stage graceful/force stop:** abandon semantics are identical in both stages (nothing is killed either way), so the second click + 5s grace is friction without a safety payoff. Rejected; single click. Its optimistic-UI idea is kept in Phase 2.
- **Router-side abandon with session reconstruction from `session_sync` scans + `repair_messages()`:** fragile new consistency machinery solving a dangling-state problem that abandon-in-agent-thread never creates. Rejected. Its registry-lookup fix for pre-CONNECT INTERRUPT routing is kept in Phase 2.
- **Router-level fake stop (send OUTPUT immediately, thread runs on):** the UI lies for minutes; a user's next INPUT would be routed as runtime input into an agent they believe stopped. Rejected.
- **Subprocess-per-tool isolation:** real kill semantics but an architecture rewrite (serialization; breaks agent-injection for `ask_user`/`DiffWriter`). Rejected.
- **`before_each_tool` interrupt poller plugin:** setting `stop_signal` in that event does not prevent the pending tool from executing outside the tool_approval plugin (`tool_executor.py:181→195` runs `tool_func` regardless) — the wrapper's pre-spawn check does the same job correctly in core. Rejected.

## Risks & open questions

1. **Orphaned tool side effects** complete in the background (shell command finishes, browser click lands) — inherent to any no-kill design. Mitigations: documented semantic; Phase 2 cooperative stop for built-in long tools; fresh agent + superseded io per request bounds cross-turn contamination. Residual: a zombie tool can race the user's immediate next instruction on shared resources (e.g. a browser profile).
2. **Orphaned LLM worker** holds a socket up to the provider timeout (≤600s); tokens for the abandoned completion are billed — still strictly cheaper than today, where the completion finishes *and* its tool batch runs. Repeated stop-clicking burns credits invisibly on `co/*`; note it in docs.
3. **Worker abandoned inside a dead `receive()`** hangs forever on the old io's condition variable — daemon-thread leak, harmless, but the Phase 1 gate hardening makes it near-impossible; keep the hardening mandatory.
4. **Same-window race** (worker completes ≈ interrupt arrives): completed result wins, frame drains at next check — bounded; needs a pinned test.
5. **Trace-closure discipline:** every interrupted `llm_call`/`tool_call` trace entry must get a closing `interrupted` entry or `hasActiveUI` wedges `isLoading` true — easiest detail to drop in review; add an e2e assertion.
6. **Thread-per-step cost** (~50µs) is noise vs. network calls; pathological agents running hundreds of instant local tools pay it — if ever measured, a single reusable worker is a drop-in.
7. **HTTP `POST /input` path** runs the agent on the event loop thread (`http_router.py:286`) — the wrapper works but hard stop is effectively WS/relay-only until that path is threaded; document.
8. **Open (Phase 2):** does the TS SDK ignore unknown frames? Gates INTERRUPT_ACK; the local fallback-timer UI is independent of the answer. Should the stopped turn render as a "Stopped by user" divider instead of the closing agent bubble (competitor convention)? Would ride a `stop_reason` field on OUTPUT; defer with the ACK work.

## Rough scope estimate

- **Phase 1 (connectonion only, no protocol/frontend changes):** `core/interrupt.py` new ~30 LOC; `core/agent.py` ~12; `core/tool_executor.py` ~8; approval/ask_user/diff_writer hardening ~10. **~60-70 LOC** + unit tests (wrapper semantics incl. race + pre-spawn check, per-provider two-terminal-state message validity, mid-batch substitution, trace closure) + one e2e (interrupt during a slow fake LLM / slow fake tool; assert <1s return, valid next turn, isLoading clears).
- **Phase 2:** router INTERRUPT branch + ACK + registry lookup ~10 LOC; oo-chat `isStopping` UI ~20 LOC; shell/browser cooperative stop ~40 LOC; optional best-effort client swap-and-close ~15 LOC + per-provider tests.
- **Docs:** `docs/ARCHITECTURE.md` frame table (INTERRUPT / INTERRUPT_ACK, currently absent); "interrupt = abandon, not kill" contract in tool-authoring docs; wiki + docs-site pages.

Each phase is independently shippable and independently revertible; Phase 1 alone converts the graceful stop into a sub-second hard stop for every provider and every agent state.

wu-changxing:
## Summary

- wrap blocking LLM completions and tool calls in an abandonable worker with ~200ms interrupt polling
- preserve provider-valid tool-call/result adjacency and mark interrupted traces explicitly
- isolate agent-injected tools with revocable IO plus session/registry-membership snapshots so late workers cannot consume future replies or commit late framework state
- make approval, ask_user, DiffWriter, and plan-review gates treat INTERRUPT as control flow
- document the design in DD-025 and the custom IO cancellation protocol

## Design

This follows DD-012, DD-018, and DD-019: tool execution stays centralized, existing event names remain stable, and `stop_signal` / `on_stop_signal` remain the lifecycle authority. Python cannot safely kill arbitrary work, so this abandons daemon workers and discards late session fields, registry-membership changes, events, results, messages, and trace. Stateful tool instances, globals, captured objects, provider work, and external side effects are not rolled back; tools needing stronger semantics must use cooperative cancellation or a killable subprocess. Custom IO adapters without the atomic cancellation protocol retain graceful boundary stopping for agent-injected tools.

Completed work wins only when it is truly terminal. Runtime input that continues the loop keeps a simultaneous INTERRUPT for the next boundary instead of swallowing it. Interrupted batches skip `after_tools` and exit through `on_stop_signal`.

## Verification

- `226 passed` across all affected agent, tool, IO, approval, event, and runtime-input suites
- `5654 passed, 13 skipped` for the full unit suite with the unrelated docs-site version assertion deselected
- deterministic mailbox/session/registry-membership/runtime-input race regressions
- focused follow-up after the boundary clarification: 167 passed; `test_interrupt.py`: 12 passed
- two expert review gates: Blocker 0 / Medium 0 / Minor 0 after clarifying the stateful-instance boundary and preserving the current Plan regression until this PR is actually rebased on #778
- ruff on new files, compileall, and `git diff --check` pass

Known unrelated baseline: `tests/unit/test_the_version_agrees_with_itself.py::test_the_docs_site_advertises_the_version_that_exists` sees the sibling docs-site checkout at 1.5.20 while this repo is 1.6.0.

Fixes #216


wu-changxing:
Final integration review found and resolved one contract-clarity issue. A stopped Python worker can still mutate a shared stateful tool instance (or any captured object); arbitrary thread side effects cannot be rolled back safely. The code comment, DD-025, network IO docs, test name, and PR body now state the exact transaction boundary: late session fields, registry membership, events/results/trace, and future mailbox reads are isolated; stateful/external side effects require cooperative cancellation or subprocess isolation.

We deliberately did not add a generic `deepcopy` pseudo-transaction: it would break bound method identity, fail on locks/provider clients, and still could not roll back arbitrary Python effects.

Review result after correction: Blocker 0 / Medium 0 / Minor 0. The Plan interrupt test remains on this main-based Draft and will be removed only when #785 is actually rebased on #778.

Commit: 81cc833f. Focused follow-up: 167 passed; `test_interrupt.py`: 12 passed. PR remains Draft; nothing was merged.

wu-changxing:
This Draft is now actually stacked on #778 and its base has been retargeted to `agent/todo-workflow-121`. During the rebase, the regression importing #778's deleted `exit_plan_and_implement` tool was removed at the correct stack boundary.

Stack author review:
- combined #778 + #785 focused suite: 229 passed
- `git diff --check`: pass
- #785-only diff contains 16 files and no unrelated #779 changes
- new head: 4885fcf0

The pre-rewrite head is retained locally for recovery, and the push used an exact `--force-with-lease`. PR remains Draft; nothing was merged.

wu-changxing:
Merged after review — the interrupt design is sound (revocable IO lease + deepcopied session, transactional commit-on-success, worker abandonment for un-killable threads, and a safe synchronous fallback for legacy IO that cannot cancel a blocked receive).

One non-blocking follow-up: `execute_single_tool` does `copy.deepcopy(original_session)` on every interruptible `_needs_agent` tool call, and the session carries `messages` (the full conversation history). For long hosted sessions that is an O(history) copy per agent-needing tool call. Not a correctness issue — nothing in the session is unpicklable, so it won't raise — but worth trimming later (e.g. copy the mutable control keys and share `messages` by reference, since the worker's late writes are discarded on interrupt anyway).

Issue #771, opened 2026-08-09
Triage the security-scanning baseline before the 1.7.0 preview


wu-changxing:
## Current verified state — 2026-08-23

The historical Alpha.2 snapshot below is no longer current. Fresh GitHub security API readback now reports the following for all three first-party 1.7 repositories:

| Repository | Open secret alerts | Open CodeQL alerts | Open Dependabot alerts | Secret scanning | Push protection |
| --- | ---: | ---: | ---: | --- | --- |
| `openonion/connectonion` | 0 | 0 | 0 | enabled | enabled |
| `openonion/connectonion-react` | 0 | 0 | 0 | enabled | enabled |
| `openonion/oo-chat` | 0 | 0 | 0 | enabled | enabled |

The two historical Google API key alerts and the former O Chat clear-text identity CodeQL alert are resolved. No secret values were retrieved or copied during this audit.

Beta 10 release commit `5d35e2fb3c96238c29906e1d7f51d1814d1cf5ae` passed the Linux 3.10–3.13 matrix, native Windows browser/E2E jobs, macOS E2E, dependency-lock audit, exact-artifact build, Trusted Publishing, public-package verification, and GitHub Release finalization. The repository-wide open-alert baseline is clear, but the exact unchanged RC still needs the final security repeat.

## Completed baseline

- [x] Secret scanning and push protection are enabled on Core, React, and O Chat.
- [x] Core has 0 open Dependabot, CodeQL, and secret-scanning alerts.
- [x] React has 0 open Dependabot, CodeQL, and secret-scanning alerts.
- [x] O Chat has 0 open Dependabot, CodeQL, and secret-scanning alerts.
- [x] Historical credential alerts have a resolved disposition; no unexplained secret alert remains.
- [x] The O Chat identity finding was closed through code rather than dismissal.
- [x] Beta 10 passed its exact release matrix, lock audit, and artifact publication/verification workflow.

## Remaining RC and stable gate

- [ ] Repeat secret-scanning, CodeQL, dependency, and release-diff checks on the preserved unchanged RC source commit.
- [ ] Confirm the RC wheel/sdist source and public artifacts introduce no new finding.
- [ ] Keep the React/O Chat repository-setting baseline current through promotion.
- [ ] Record the intended merge-rule enforcement separately in `openonion/oo-chat#148`; current O Chat rulesets remain empty.

## Definition of done

[...]

PR #788, opened 2026-08-09
Close the pytest security alert and pin CodeQL

wu-changxing:
Part of #771.

## What changed

- raise the optional development floor from `pytest>=7.0.0` to the first patched release, `pytest>=9.0.3`;
- refresh the exact lock from pytest 9.0.1 to 9.1.1;
- move the existing Python CodeQL workflow from mutable major tags to immutable current stable commits:
  - `actions/checkout` v7.0.1;
  - `github/codeql-action` v4.37.6;
- remove `actions: read`, which CodeQL does not require for this public repository;
- keep `security-events: write` scoped to the analysis job.

## Why

Dependabot alert #1 reports `GHSA-6w46-j5rx-g56g` / `CVE-2025-71176`: pytest before 9.0.3 has insecure UNIX tmpdir handling. Updating only the lock would allow a future resolver to select a vulnerable version again, so the declared development floor and lock move together.

The CodeQL workflow is already the correct small design; this PR only makes its supply-chain and permission boundary explicit. No new matrix, build step, custom query, or abstraction is added.

## Validation

- `uv lock --check`
- full lock including the `dev` extra: `pip-audit 2.10.1` reports **no known vulnerabilities**
- `actionlint 1.7.12 .github/workflows/codeql.yml`
- full pytest 9.1.1 run: **6092 passed, 26 skipped, 174 deselected, 1 failed**

The one local failure is the existing cross-check against the neighboring dirty docs-site checkout: it advertises 1.5.20 while ConnectOnion is 1.6.0. The same focused test fails identically on an untouched detached `origin/main` worktree, so it is not introduced by this diff. CI uses its isolated checkout and remains the authoritative matrix.

## Honest weakness

The Dependabot alert will remain open until this PR is merged into the default branch. Also, this PR does not enable GitHub Advanced Security for private `oo-api`; that owner/plan decision is documented on #771.

## AI assistance

AI assisted with the release-surface audit, advisory triage, implementation, and review. Advisory metadata, official action versions and SHAs, CodeQL permissions, the full dependency audit, and baseline failure comparison were independently verified before opening this Draft PR.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -26,19 +26,18 @@ jobs:
     permissions:
       # Required to upload the SARIF result to the Security tab.
       security-events: write
-      actions: read
       contents: read
 
     steps:
       - name: Check out
-        uses: actions/checkout@v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
         with:
           languages: python
 
       - name: Analyze
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
         with:
           category: "/language:python"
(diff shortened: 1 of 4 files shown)


wu-changxing:
Review follow-up (`cc986506`):

- added a dev-extra-specific regression test that independently locks all three security invariants:
  1. `pyproject.toml` declares `pytest>=9.0.3`;
  2. root `uv.lock` metadata carries that floor with `extra == 'dev'`;
  3. the concrete locked pytest is at least 9.0.3;
- corrected the CodeQL schedule comment: weekly scans apply updated CodeQL queries to unchanged source; dependency advisories remain Dependabot/audit's responsibility.

Validation after the change: focused security-floor suite 2 passed; Ruff, actionlint 1.7.12, and diff check pass. The full GitHub matrix is rerunning. PR remains Draft and unmerged.

PR #790, opened 2026-08-09
fix(gmail): parse HTML bodies safely

### LINKED ISSUE #789: security: parse Gmail HTML bodies instead of filtering tags with regex
## Problem

The initial CodeQL scan from #771 reports `py/bad-tag-filter` in `Gmail._extract_body()`. The HTML fallback removes `<script>` and `<style>` with regular expressions, so valid tag variants such as `<script >` can bypass the intended filter.

This path handles untrusted email HTML before returning text to an agent/user. Even though it is a text-conversion boundary rather than an HTML-rendering boundary, its contract should be deterministic and should not claim filtering that common HTML syntax can evade.

### Option A — widen the regular expressions

Small diff, but still treats HTML as regular text. Each newly discovered syntax variant creates another patch, and malformed HTML remains ambiguous.

### Option B — add a sanitizer dependency

A sanitizer such as Bleach is useful when preserving safe HTML. Here the public contract returns plain text, so retaining markup would add behavior and a dependency without a user need.

### Option C — use BeautifulSoup (selected)

BeautifulSoup is already a required dependency and is already the repository's established implementation in `WebFetch.strip_tags()`. Parse once, explicitly remove `script`/`style`, extract text, and normalize whitespace. This is the smallest implementation with a clear plain-text contract.

This follows:

- ADR-010: keep the safe behavior automatic for the caller;
- ADR-012: keep MIME traversal separate from HTML-to-text conversion;
- ADR-020: make the security-relevant transformation deterministic and directly testable.

## Acceptance criteria

- [ ] HTML Gmail bodies are converted with the existing parser dependency, not regex tag filters.
- [ ] `script` and `style` content is excluded for ordinary, spaced, attributed, and mixed-case tags.
- [ ] HTML entities are decoded and human-readable separation is preserved.
- [ ] Plain-text and multipart preference behavior does not change.
- [ ] Focused Gmail unit tests and Ruff pass.
- [ ] The PR remains Draft until maintainer review; the agent does not merge it.

Parent: #771

wu-changxing:
## Summary

- replace regex-based Gmail HTML tag filtering with the repository's existing BeautifulSoup parser
- explicitly remove `script` and `style` elements before plain-text extraction
- preserve entity decoding and readable spacing
- add a regression test covering spaced, attributed, and mixed-case dangerous tags

Closes #789
Parent security baseline: #771

## Design

The design alternatives and ADR rationale were discussed on #789 before implementation. BeautifulSoup is already a required dependency and already establishes this repository's HTML-to-text pattern in `WebFetch.strip_tags()`. A sanitizer dependency would be unnecessary because Gmail's contract returns plain text, while expanding regexes would preserve the class of bug reported by CodeQL.

## Verification

- `python -m pytest tests/unit/test_gmail.py -q` — 56 passed
- `python -m ruff check connectonion/useful_tools/gmail.py tests/unit/test_gmail.py --ignore E402,I001,F401,E722` — passed; ignored codes are pre-existing findings in these legacy files
- `git diff --check` — passed

`ruff format --check` reports both whole legacy files would be reformatted; this PR intentionally avoids a noisy full-file formatting rewrite.

## Review notes

- Draft only: do not merge until maintainer review.
- The nine URL-substring and two clear-text-logging CodeQL findings are separately documented on #771 as test-only false positives; this PR does not hide or dismiss them.

### DIFF
--- a/connectonion/useful_tools/gmail.py
+++ b/connectonion/useful_tools/gmail.py
@@ -379,7 +379,16 @@ class Gmail:
             soup = BeautifulSoup(html, 'html.parser')
             for tag in soup(['script', 'style']):
                 tag.decompose()
-            return ' '.join(soup.get_text(separator=' ', strip=True).split())
+            block_tags = (
+                'address', 'article', 'aside', 'blockquote', 'br', 'caption', 'dd', 'details',
+                'div', 'dl', 'dt', 'fieldset', 'figcaption', 'figure', 'footer', 'form',
+                'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'header', 'hr', 'li', 'main', 'nav',
+                'ol', 'p', 'pre', 'section', 'summary', 'table', 'tbody', 'td', 'tfoot',
+                'th', 'thead', 'tr', 'ul',
+            )
+            for tag in soup(block_tags):
+                tag.insert_after(' ')
+            return ' '.join(soup.get_text().split())
 
         # Single part email
         if 'body' in payload and payload['body'].get('data'):
--- a/connectonion/useful_tools/gmail.py
+++ b/connectonion/useful_tools/gmail.py
@@ -373,15 +373,13 @@ class Gmail:
 
     def _extract_body(self, payload) -> str:
         """Extract body from email payload, preferring text/plain, falling back to stripped HTML."""
-        import re
-        from html import unescape
+        from bs4 import BeautifulSoup
 
         def strip_html(html: str) -> str:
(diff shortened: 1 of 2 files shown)


wu-changxing:
Addressed expert review's Medium finding in the latest commit:

- preserve adjacent inline text (`co<strong>de</strong>!` → `code!`)
- add separators only after structural block/line/table boundaries
- extend the regression test to lock both inline continuity and block separation

Reverification:

- focused regression: 1 passed
- full Gmail unit file: 56 passed
- Ruff (excluding documented pre-existing legacy-file codes): passed
- `git diff --check`: passed

Requesting re-review on the new head. The PR remains Draft and unmerged.

PR #791, opened 2026-08-09
Parse Gmail HTML bodies with BeautifulSoup instead of regex tag filters (#789)

### LINKED ISSUE #789: security: parse Gmail HTML bodies instead of filtering tags with regex

eeshsaxena:
Fixes #789

Draft per the issue's acceptance criteria (agent does not merge).

### Problem

`Gmail._extract_body()` removed `<script>`/`<style>` with regular expressions before returning email text. Ordinary HTML syntax variants evade that filter, so the script/style content is handed to the agent as body text:

- spaced tag: `<script >...</script >`
- attributed tag: `<script type="text/javascript">...`
- mixed case: `<SCRIPT>...</SCRIPT>`

CodeQL reports this as `py/bad-tag-filter`. This is the untrusted-email-HTML to text boundary, so the contract should be deterministic rather than claiming filtering that common syntax can bypass.

### Fix (Option C from the issue)

Parse the HTML with `beautifulsoup4` (already a required dependency) rather than pattern-matching tags, mirroring the repository's established `WebFetch.strip_tags()`:

```python
soup = BeautifulSoup(html, 'html.parser')
for tag in soup(['script', 'style']):
    tag.decompose()
text = soup.get_text(separator=' ')
return re.sub(r'\s+', ' ', text).strip()
```

`get_text` decodes HTML entities, so the manual `unescape` is dropped. No new dependency and the public contract stays plain text.

### Acceptance criteria

- [x] HTML bodies converted with the existing parser dependency, not regex tag filters
- [x] script/style excluded for ordinary, spaced, attributed and mixed-case tags
- [x] HTML entities decoded and human-readable separation preserved
- [x] Plain-text and multipart preference behavior unchanged
- [x] Focused Gmail unit tests pass (64 passed)

### Tests

Added `TestExtractBodyHtml` in `tests/unit/test_gmail.py` covering all four tag variants for both script and style, entity decoding, adjacent-element separation, and the multipart plain-text preference. Full `test_gmail.py` passes (64 passed).

### DIFF
--- a/connectonion/useful_tools/gmail.py
+++ b/connectonion/useful_tools/gmail.py
@@ -374,13 +374,20 @@ class Gmail:
     def _extract_body(self, payload) -> str:
         """Extract body from email payload, preferring text/plain, falling back to stripped HTML."""
         import re
-        from html import unescape
+        from bs4 import BeautifulSoup
 
         def strip_html(html: str) -> str:
-            html = re.sub(r'<style[^>]*>.*?</style>', '', html, flags=re.DOTALL | re.IGNORECASE)
-            html = re.sub(r'<script[^>]*>.*?</script>', '', html, flags=re.DOTALL | re.IGNORECASE)
-            text = re.sub(r'<[^>]+>', '', html)
-            text = unescape(text)
+            # Parse the HTML rather than filtering tags with regular expressions.
+            # A regex filter is evaded by ordinary syntax variants such as
+            # ``<script >`` or ``<SCRIPT>`` (CodeQL py/bad-tag-filter), and this
+            # path handles untrusted email HTML. Mirror WebFetch.strip_tags(),
+            # the repository's established parser-based conversion.
+            soup = BeautifulSoup(html, 'html.parser')
+            for tag in soup(['script', 'style']):
+                tag.decompose()
+            # get_text decodes HTML entities and the separator keeps text from
+            # adjacent elements human-readable instead of being run together.
+            text = soup.get_text(separator=' ')
             return re.sub(r'\s+', ' ', text).strip()
 
         # Single part email
--- a/tests/unit/test_gmail.py
+++ b/tests/unit/test_gmail.py
@@ -1229,3 +1229,65 @@ class TestGmailIntegration:
         assert 'get_labels' in agent.tools
         assert 'get_all_contacts' in agent.tools
         assert 'update_contact' in agent.tools
+
+
+class TestExtractBodyHtml:
+    """Tests for _extract_body HTML-to-text conversion (issue #789).
+
+    The HTML fallback must parse the markup rather than filtering tags with
+    regular expressions, so ordinary syntax variants such as a spaced or
+    attributed <script> tag cannot smuggle script/style content into the text
+    handed to the agent (CodeQL py/bad-tag-filter)."""
+
+    @pytest.fixture
+    def gmail(self):
+        with patch.dict(os.environ, {
+            "GOOGLE_SCOPES": "gmail.readonly gmail.send",
+            "GOOGLE_ACCESS_TOKEN": "test_token",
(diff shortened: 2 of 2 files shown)


wu-changxing (changes requested):
Thanks for addressing the parser boundary and for covering the tag variants. I checked this against Issue #789's design discussion and the repository's “test behavior, not implementation” principle.

**Medium — `soup.get_text(separator=' ')` corrupts ordinary inline text.**

The separator is inserted between every text node, not only between block elements. For example:

```python
BeautifulSoup('<p>co<strong>de</strong>!</p>', 'html.parser').get_text(separator=' ')
# 'co de !'
```

Real email HTML frequently wraps fragments in `<strong>`, `<span>`, `<a>`, and similar inline tags, so this can change words, punctuation, URLs, and identifiers handed to the agent. `test_adjacent_elements_are_separated` covers the block case but does not pin inline continuity.

Please preserve inline text while separating blocks—for example, insert a boundary only after block-level elements and then normalize whitespace—and add a regression assertion such as `co<strong>de</strong>! -> code!`.

I ran the PR's focused suite in an isolated checkout: **64 passed**. The parser/security direction is otherwise sound. There is also an earlier Draft implementation in #790 for the same issue; it already has the inline-continuity regression, dual expert review, CodeQL, and the full cross-platform CI matrix green. Please coordinate with the maintainer before both implementations advance. I have not merged or closed either PR.

wu-changxing:
Superseded by #790, now merged.

Both PRs fix #789 the same way (parse with BeautifulSoup, decompose script/style). They differ in how text is rejoined, and that decided it:

- #790 inserts a separator only after **block-level** tags, so inline formatting stays joined: `<p>co<strong>de</strong>!</p>` -> `code!`
- this PR uses `get_text(separator=' ')`, which separates every text node, so the same input becomes `co de !`

Mid-word inline formatting is common in real email HTML, so #790's output is the correct one. This PR's parametrized tag-variant tests (spaced / attributed / mixed-case script and style) are genuinely better coverage than what #790 shipped — grafting those onto main would be a welcome follow-up.

PR #797, opened 2026-08-10
Superseded: paid engine cache lookup spike (reimplement via Onionwright session API)

wu-changxing:
## Superseded — do not merge

This PR preserves useful investigation and real Linux launch evidence, but its architecture is no longer the ConnectOnion 1.8 contract.

It directly reads Onionwright's private cache layout:

- `~/.onionwright/attestation.json`;
- `~/.onionwright/chrome/<revision>/chrome`.

It also implements the superseded Free/Pro and cumulative-credit decision and does not provide the required paid-session heartbeat, `paid_until` watchdog, runtime licence-file refresh or actual launch-seam integration.

The canonical 1.8 implementation is now:

- #511 — PAYG product and launch contract;
- openonion/onionwright#40 — one public launcher/session-supervision API;
- openonion/oo-api#168 — atomic billing and server-signed runtime licence;
- openonion/browser#96 — C++ runtime verification;
- #1154 — cross-repository release control.

ConnectOnion must consume Onionwright's public API rather than its filesystem internals. The feature remains required for 1.8; this particular PR is closed so it cannot be mistaken for merge-ready implementation. Useful tests/evidence may be carried forward into a new PR against the canonical contract.

### DIFF
--- a/connectonion/useful_tools/browser_tools/engine.py
+++ b/connectonion/useful_tools/browser_tools/engine.py
@@ -98,23 +98,6 @@ def choose(attestation, onion_present):
 
 # The paid engine is not reachable from here yet, and this names exactly what
 # is missing rather than leaving a stub that reads as finished.
-#
-# `onionwright.licence.license.load(path, server_key)` needs a cache path and
-# the pinned server key; `onionwright.launcher.resolve_binary(token,
-# server_key, pin)` needs an authenticated token and a Chromium revision.
-# connectonion holds none of those three today. Supplying them is the open
-# work in openonion/connectonion#511 — not something to fake here.
-#
-# Until then the defaults below raise, `resolve()` reads that as "no licence"
-# the same way it reads an offline machine, and everyone gets the free engine.
-# That is the right failure direction and the wrong end state, so
-# test_the_paid_path_is_not_wired_up_yet asserts it out loud: when the wiring
-# lands, that test fails and has to be deleted, which is the point of it.
-PAID_WIRING_PENDING = (
-    "onionwright is installed but connectonion has no licence configuration "
-    "yet (cache path, pinned server key, Chromium revision) — see "
-    "openonion/connectonion#511"
-)
 
 
 def _cached_attestation():
@@ -130,10 +113,23 @@ def _cached_attestation():
     command is not the right place to discover that oo-api is slow. The
     refresh belongs to the licence client's own schedule (it renews at the
     12-hour half-life of a 24-hour attestation), not to `co browser go_to`.
-    """
-    import onionwright.licence  # noqa: F401
 
-    raise NotImplementedError(PAID_WIRING_PENDING)
+    The cached payload is read but its signature is not checked here, and that
+    is a deliberate scope choice rather than an omission. Verifying needs the
+    server key, which `onionwright.launcher.server_public_key()` fetches over
+    the network -- exactly what this function promises not to do. Nothing is
+    lost by trusting it: the only thing a forged local file buys is an attempt
+    to launch the paid binary, which compiles the server's public key into
+    itself (browser patches/core/0005) and refuses a licence it cannot verify.
+    The gate is the engine and the download, not this hint.
+    """
+    import json
+
+    from onionwright.launcher import HOME
+
+    path = HOME / "attestation.json"
+    payload = json.loads(path.read_text())["payload"]
+    return json.loads(bytes.fromhex(payload).decode())
 
 
 def _onion_path():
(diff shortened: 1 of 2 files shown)


wu-changxing:
Reviewed as the author of the half this completes — including the tombstone test it deletes, which is exactly what should happen to it.

The wiring is right and the two scope decisions are argued rather than assumed: not verifying the signature locally (it needs the network, and the binary is the real gate), and not downloading a missing binary from a browser command. Both convinced me.

Two defects found, both reproduced rather than reasoned about.

## 1. An expired attestation still selects the paid engine

`_cached_attestation()` returns the payload as parsed, and `choose()` reads only `tier` and `active`. `expires_at` is in the payload and is never looked at:

```
$ # payload from a month ago, tier=pro, active=true — a stale file on disk
  choose(attestation=expired, onion_present=True) -> onion
  note: None
  payload carries expires_at: True
  choose() reads expires_at:  False
```

That matters more than it looks, because of what happens next. The PR argues correctly that a forged local file buys nothing, since the binary compiles the server key and refuses a licence it cannot verify. But `choose()` returning `onion` means `co browser` **launches that binary**, and the binary then refuses — so the run ends with no browser at all.

That is the one outcome this module was built to make impossible. Every other branch degrades to patchright precisely so a licence problem can never cost someone their browser; a lapsed subscriber with a stale file on disk is the one input that gets past it.

The fix costs nothing and needs no network — the expiry is a number already in hand. `onionwright.licence.license` has `valid_at()`/`decide()` doing exactly this comparison, so it can be reused rather than rewritten.

## 2. `sorted()[-1]` picks the wrong Chromium when revisions differ in digit width

```python
executables = sorted((HOME / "chrome").glob("*/chrome"))
return executables[-1] if executables else None
```

The directory name is a dotted Chrome version (`chromium.pin` = `150.0.7871.187`), and string order is not version order:

```
150.0.7871.187 vs 150.0.7871.9   -> sorted picks 150.0.7871.9    WRONG (older)
150.0.7871.187 vs 151.0.8000.5   -> sorted picks 151.0.8000.5    ok
```

So a machine that has upgraded within a build line silently runs the **older** binary. For this product that is not cosmetic: the older build is the more-detectable one, and it fails in the way the customer is paying to avoid, with nothing on screen to suggest why.

Splitting on `.` and comparing as a tuple of ints fixes it:

```python
key=lambda p: tuple(int(n) for n in p.parent.name.split("."))
```

A non-numeric directory under `chrome/` would raise on that key, so it wants either a guard or a filter — worth deciding deliberately, since `resolve()` swallows exceptions from this function and would read a crash here as "no paid binary".

## Not blocking

Both are small and neither is in the tiering logic itself. Happy to push the fixes to this branch if that is easier than a round trip.

wu-changxing:
Review status: this PR is useful groundwork but is not merge-ready yet.

Two concrete integration blockers remain:

1. The PR body correctly says `resolve()` is not called by the browser launch path. Merging it would therefore not complete #511 or change production engine selection; the issue must remain open until the launch seam is wired and exercised end to end.
2. `_onion_path()` chooses `sorted(...)[-1]`. Chromium directory names are version strings, so lexicographic order can select `99...` over `150...`. More importantly, onionwright considers an extraction valid only when both `chrome` and `.extracted` exist and the marker names the verified archive. This lookup ignores that marker and can select a stale or partially extracted binary. Onionwright's real API avoids both ambiguities by requiring an explicit `pin` in `resolve_binary(token, server_key, pin)`.

Best-practice next step before more code:

- define one Chromium pin shared by the ConnectOnion launch call and onionwright;
- use onionwright's validated resolution/launch seam rather than scanning cache directories;
- keep the free path independent of licensing/network failure;
- add a real launch-path test proving free → patchright and eligible paid → the validated pinned onion binary;
- only then mark the PR ready and close #511.

I am marking #797 draft so it is not accidentally merged as a completed feature. No code has been discarded.

wu-changxing:
Re-review of `c875ee59`: no head change, so the existing blockers still stand.

#511 remains open, `resolve()` is still not wired into the production launch path, and the cache scan still bypasses onionwright’s pinned/validated resolver (including its extraction marker). The stale-attestation and lexicographic-version failure modes previously reproduced are therefore still reachable in this implementation.

Keep this Draft and do not close #511. The next reviewable revision should use one explicit Chromium pin, call onionwright’s validated resolver from the actual launch seam, preserve an independent free fallback, and include a real launch-path test for both tiers.

wu-changxing:
Closing this Draft implementation as deferred from the current release train. Paid/alternate browser-engine lookup remains tracked by #511 and belongs with the 1.8 remote-browser work after the stable session boundary in #1154. Reopen or create a fresh PR from current main when that train is active.

wu-changxing:
Release-target correction from Aaron: this work is required in **1.8**, not deferred beyond it. Reopening the PR and assigning the existing 1.8 milestone. Before merge, rebase onto current `main`, verify how this lookup slice connects to the actual browser launch path, and keep #511 as the canonical end-to-end acceptance issue.

PR #815, opened 2026-08-10
outlook: save attachments to disk with co outlook download

wu-changxing:
`co outlook` could send attachments but never read one back. An emailed file — a cover image, a research PDF — had to be fetched by hand through the web client, which is what happened today when a co-host emailed event artwork.

```
co outlook download 3 --to ./covers
✓ ./covers/ai-sales-engine-linkedin-cover-16x9.jpg
```

## What it does

`Outlook.download_attachments(email_id, out_dir)` writes an email's `fileAttachment` bytes to a directory and returns the saved paths. The CLI wraps it as `co outlook download <#> --to <dir>`, accepting the same listing numbers as `read` and `reply`.

## Boundaries

Reading attachments in has the mirror image of the risks #811 closed for sending them out, so it reuses the same rules:

- The destination directory goes through the project boundary — outside it raises `PermissionError` unless `allow_external_attachments` is set, exactly as the send path does.
- **The attachment name is sender-controlled.** It is reduced to its last path segment before joining, so a message whose attachment is named `../../.ssh/authorized_keys` still lands inside the chosen directory. A test asserts the file appears in `out/` and not in the parent.
- `itemAttachment` and `referenceAttachment` carry no `contentBytes`; they are skipped rather than written empty.

## Tests

Four cases in `tests/unit/test_outlook.py`: bytes land on disk, the traversal name stays contained, an out-of-project destination is refused, and byte-less attachments are skipped.

Note on running them locally: the suite's `_no_stray_project_above_the_test` teardown guard fails for every test in the file when the checkout sits under a directory that has its own `.co/` (a worktree under `$HOME`, for instance). The 41 tests themselves pass; the teardown assertions are about where the checkout lives, not about this change.

### DIFF
--- a/connectonion/cli/commands/outlook_commands.py
+++ b/connectonion/cli/commands/outlook_commands.py
@@ -209,6 +209,25 @@ def handle_outlook_read(email_id: str):
     console.print(f"\n[dim]{marked}Reply with:[/dim] [bold]co outlook reply <#> <message>[/bold]\n")
 
 
+def handle_outlook_download(email_id: str, out_dir: str = "."):
+    """Save an email's attachments to disk. Accepts the listing # or a full message id."""
+    outlook = _outlook()
+    resolved = _resolve_email_id(outlook, email_id)
+    if not resolved:
+        console.print(f"\n[yellow]No email #{email_id} in your last listing — run co outlook to refresh.[/yellow]\n")
+        raise typer.Exit(1)
+
+    saved = outlook.download_attachments(resolved, out_dir)
+    if not saved:
+        console.print("\n[yellow]No file attachments on that email.[/yellow]\n")
+        return
+
+    console.print()
+    for path in saved:
+        console.print(f"[green]✓[/green] {path}")
+    console.print()
+
+
 def handle_outlook_reply(email_id: str, message: str, at: str = None):
     """Reply to an email from the last listing (threaded via Graph). A message of '-' reads stdin."""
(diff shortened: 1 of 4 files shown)


wu-changxing:
Flagging a real blocker this PR would fix, in case it helps prioritise: today we needed to confirm whether an inbound background-check email carried the attachments its body claimed ("Any documentation that was supplied by X will be attached to this email"), and there was no way to answer that from the CLI — the attachment is invisible without `co outlook download`. The answer had to be left as unknown.

CI is green (11 checks) and the branch is mergeable, with no review yet.

Filed #840 for the neighbouring gap in the same command family: `co outlook read` strips hyperlink URLs, so verification and questionnaire links are unreachable too.

wu-changxing:
Security review found that the original download path could overwrite an existing project file (the CLI defaults to `--to .`) and could follow a pre-existing symlink. Fixed in `5adf7f96`:

- atomically create attachment files with `O_EXCL` / `O_NOFOLLOW` and mode 0600;
- refuse existing names instead of overwriting;
- sanitize terminal control characters in sender-controlled filenames;
- validate base64 before creating the output file;
- update the code-map side-effect contract and add regression tests.

Post-fix validation: 45/45 Outlook unit tests passed locally. GitHub Linux 3.10–3.13, Windows, Windows E2E, lockfile, Analyze Python, and CodeQL checks are all green. Final self-review: no remaining blocking findings.

PR #823, opened 2026-08-10
feat: add OrcaRouter as a named LLM provider

XiaoHuo888-hue:
## Description

This registers [OrcaRouter](https://www.orcarouter.ai) the same way the existing `OpenRouter` provider is wired, so the model picker, config reference and docs stay consistent. OrcaRouter is an OpenAI-compatible gateway: one `ORCAROUTER_API_KEY` (keys start with `sk-orca-`) unlocks 150+ models from OpenAI, Anthropic, Google, DeepSeek, Qwen, MiniMax and xAI behind a single `https://api.orcarouter.ai/v1` endpoint. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

I'm an engineer on the OrcaRouter team.

## Related Issue

None — new provider addition.

## How this was written

- [ ] I wrote this myself
- [ ] Mostly AI-generated

**Which model?** DeepSeek V4 Flash via Claude Code.

The exact set of models on the OrcaRouter gateway can grow, so the docs example pins `orcarouter/openai/gpt-4o-mini`, which is in the live `/v1/models` catalog today. The `sk-orca-` prefix check in `detect_api_provider()` had to be placed before the generic `sk-` OpenAI branch; the ordering is locked down by a new parameterized test.

**What did you *not* verify — which paths did you never actually run?**
I did not run the real-API test suite (`-m real_api`, needs keys) or the Windows `co browser` e2e suite (needs Chrome). A full `pytest tests/` on this Windows machine ends at 5755 passed / 145 failed / 59 skipped; every failure is in test files this PR does not touch (Windows-only issues: symlink privilege, `bash is Unix/Mac only` skips, GBK codec on CJK file reads, missing `docs-site` submodule). I verified all test files this PR touches pass, but I did not exhaustively diff the 145 unrelated failures against a pristine checkout.

If the prefix-ordering regressed, an `sk-orca-` key would be detected as a plain OpenAI user key and stored under `OPENAI_API_KEY`, so the agent would send it to OpenAI and get a 401. The new `TestDetectApiProvider` test guards exactly this.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Changes Made

- `connectonion/core/llm.py`: new `OrcaRouterLLM` class mirroring `OpenRouterLLM` — OpenAI-compatible `chat.completions` wire to `https://api.orcarouter.ai/v1`, `ORCAROUTER_API_KEY`, `orcarouter/` model-prefix stripping, JSON-mode `structured_complete()`. `create_llm()` factory gains an `orcarouter/` branch.
- `connectonion/cli/commands/project_cmd_lib.py`: `ORCAROUTER_API_KEY` added to the environment check, `configure_env_for_provider`, and `PROVIDER_TO_ENV`. `detect_api_provider()` now checks `sk-or-` (OpenRouter) and `sk-orca-` (OrcaRouter) before the generic `sk-` OpenAI prefix, so neither router key is misdetected as a plain OpenAI user key.
- `connectonion/cli/commands/create.py`, `init.py`, `doctor_commands.py`, `status_commands.py`: OrcaRouter key detection, `.env` template line, doctor credential guidance, and status credential table entry.
- `docs/concepts/models.md` and `README.md`: OrcaRouter env block, usage example, and provider notes mirroring OpenRouter.
- Tests: `tests/unit/test_llm_errors.py` (missing-key, endpoint, structured JSON mode, factory routing), `tests/unit/test_project_cmd_lib.py` (prefix detection ordering), `tests/unit/test_uniform_provider_errors.py` (uniform auth-error translation), `tests/e2e/cli/test_cli_create.py` and `tests/e2e/cli/test_cli_status.py` (CLI key detection and status rows).

No new dependencies.

## Testing

```
$ pytest tests/unit/test_llm_errors.py tests/unit/test_uniform_provider_errors.py tests/unit/test_project_cmd_lib.py tests/e2e/cli/test_cli_create.py tests/e2e/cli/test_cli_status.py -q
122 passed in 53.33s
```

Live test against the real OrcaRouter gateway (real key, same endpoint and model-name semantics this PR registers):

```
$ ORCAROUTER_API_KEY=... python _work/connectonion/l3_test.py
[1] picked model: openai/gpt-4o-mini
[2] create_llm -> OrcaRouterLLM, stripped model: openai/gpt-4o-mini
[2] complete() -> 'ok'
[2] usage: input_tokens=12 output_tokens=1 cached_tokens=0 cache_write_tokens=0 cost=1.5e-05 total_tokens=0
[3] structured_complete() -> ticker='AAPL' direction='up' score=8
[4] detect_api_provider -> orcarouter
L3 OK
```

## Scope

The change is confined to the LLM provider layer and its CLI plumbing, exactly where the existing OpenRouter integration already lives: the provider class and factory in `core/llm.py`, the shared provider registry in `project_cmd_lib.py`, the four CLI command modules that enumerate credentials, the two docs files, and the corresponding tests. Adding a named provider alongside OpenRouter keeps the model picker, `.env` templates, `co doctor`, and `co status` consistent rather than introducing a second pattern for routing gateways.

## Example Usage

```python

agent = Agent("assistant", model="orcarouter/openai/gpt-4o-mini")
response = agent.input("What is 2+2?")
```

Set `ORCAROUTER_API_KEY` in the environment or in `<project>/.env`.

## Checklist

Not applicable.

## Additional Notes

The `detect_api_provider()` reorder also fixes a pre-existing latent bug: an OpenRouter `sk-or-...` key was previously matched by the generic `sk-` OpenAI branch before its own dedicated branch, so `--key sk-or-...` was stored as `OPENAI_API_KEY`. Both router prefixes are now checked before the OpenAI fallback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -829,7 +829,7 @@ from connectonion.useful_tools.browser_tools import BrowserAutomation
 - `co deploy` for one-command deployment
 
 **Recently Completed:**
-- Multiple LLM providers (OpenAI, Anthropic, Gemini, Groq, Grok, OpenRouter)
+- Multiple LLM providers (OpenAI, Anthropic, Gemini, Groq, Grok, OpenRouter, OrcaRouter)
 - Managed API keys (`co/` prefix)
 - Plugin system
 - Google OAuth integration
--- a/connectonion/cli/commands/create.py
+++ b/connectonion/cli/commands/create.py
@@ -152,6 +152,8 @@ def handle_create(name: Optional[str], ai: Optional[bool], key: Optional[str],
                         detected_keys["grok"] = env_value.strip()
                     elif env_key_name == "OPENROUTER_API_KEY" and env_value.strip():
                         detected_keys["openrouter"] = env_value.strip()
+                    elif env_key_name == "ORCAROUTER_API_KEY" and env_value.strip():
+                        detected_keys["orcarouter"] = env_value.strip()
                     elif env_key_name == "OPENONION_API_KEY" and env_value.strip():
                         detected_keys["openonion"] = env_value.strip()
 
@@ -398,6 +400,7 @@ def handle_create(name: Optional[str], ai: Optional[bool], key: Optional[str],
                 "# GEMINI_API_KEY=",
                 "# XAI_API_KEY=",
                 "# OPENROUTER_API_KEY=",
+                "# ORCAROUTER_API_KEY=",
             ])
 
         env_content = "\n".join(env_lines) + "\n"
--- a/connectonion/cli/commands/doctor_commands.py
+++ b/connectonion/cli/commands/doctor_commands.py
@@ -392,6 +392,7 @@ def handle_doctor(*, fix: bool = False, yes: bool = False, json_output: bool = F
         "GROQ_API_KEY": "set GROQ_API_KEY in <project>/.env",
         "XAI_API_KEY": "set XAI_API_KEY in <project>/.env",
         "OPENROUTER_API_KEY": "set OPENROUTER_API_KEY in <project>/.env",
+        "ORCAROUTER_API_KEY": "set ORCAROUTER_API_KEY in <project>/.env",
         "MISTRAL_API_KEY": "set MISTRAL_API_KEY in <project>/.env",
     }
     api_key = None
--- a/connectonion/cli/commands/init.py
+++ b/connectonion/cli/commands/init.py
@@ -270,6 +270,7 @@ def handle_init(ai: Optional[bool], key: Optional[str], template: Optional[str],
 # GROQ_API_KEY=
 # XAI_API_KEY=
 # OPENROUTER_API_KEY=
+# ORCAROUTER_API_KEY=
 
 # Optional: Override default model
 # MODEL=co/gemini-3.6-flash
--- a/connectonion/cli/commands/project_cmd_lib.py
+++ b/connectonion/cli/commands/project_cmd_lib.py
@@ -4,7 +4,7 @@ LLM-Note:
   Dependencies: imports from [os, re, sys, time, shutil, rich.console, rich.prompt, rich.progress, rich.table, rich.panel, datetime, pathlib, __version__, address] | imported by [cli/commands/init.py, cli/commands/create.py] | calls LLM APIs for custom template generation | tested indirectly via test_cli_init.py and test_cli_create.py
   Data flow: provides utility functions called by init.py and create.py → validate_project_name() checks regex patterns → check_environment_for_api_keys() scans env vars for OpenAI/Anthropic/Google/Groq/Grok/OpenRouter keys → detect_api_provider() inspects key format to identify provider → api_key_setup_menu() displays interactive menu for key selection → generate_custom_template_with_name() calls LLM API with custom prompt to generate agent.py code → show_progress() displays Rich spinner → LoadingAnimation context manager for long operations → get_special_directory_warning() warns about home/root dirs
   State/Effects: no persistent state | reads from environment variables | writes to stdout via rich.Console | calls LLM APIs (OpenAI/Anthropic/Google) when generating custom templates | creates Rich UI elements (tables, panels, progress bars, prompts) | writes no files except create_host_yaml() (.co/host.yaml)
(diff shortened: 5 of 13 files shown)


wu-changxing (changes requested):
Thanks for the careful contribution. The endpoint, `sk-orca-` key prefix,
provider-prefixed fixed model IDs, and the need to put both router prefixes
before generic `sk-` are consistent with OrcaRouter's current official docs.
The targeted assertions also pass locally.

I am requesting changes for two contract blockers before this can merge:

1. `model.removeprefix("orcarouter/")` breaks OrcaRouter's own route namespace.
   The official automatic model is `orcarouter/auto`, and named routers also
   live under `orcarouter/...`; this implementation sends `auto` or the bare
   router name to the wire. Requiring users to write
   `orcarouter/orcarouter/auto` is not an acceptable public interface. We need a
   tested normalization rule that preserves native automatic/named route IDs
   while still turning `orcarouter/openai/gpt-4o-mini` into
   `openai/gpt-4o-mini`.
2. The PR description says tool governance is default-deny. OrcaRouter's
   official Firewall verdict documentation says a new policy defaults to
   `audit`, which allows and records calls; default-deny must be configured.
   Adding this provider also does not create or attach a Firewall policy, so the
   integration must describe these as optional upstream capabilities rather
   than an enabled ConnectOnion security boundary.

The design and proposed resolution are recorded in #858 before more code is
written. That issue also covers sharing the duplicated OpenAI-compatible
plumbing and making the fallback cost estimate explicit for gateway/named
routes.

Integration is not ready yet: GitHub has run no checks on this PR, and the
branch predates current `main`. In this workspace all 122 targeted assertions
passed, but the stale branch produced 120 teardown errors from the project-root
containment fixture that current `main` has since fixed. Please integrate
current `main` and run the full required matrix after the #858 contract is
implemented.

Official references:

- https://docs.orcarouter.ai/getting-started/models
- https://docs.orcarouter.ai/routing/auto-router
- https://docs.orcarouter.ai/security/firewall/verdicts
- https://docs.orcarouter.ai/operations/billing-and-usage

wu-changxing:
## Merge blocker review — provider semantics and policy claims need correction

This PR should not merge in its current Ready state.

- Model normalization strips or rewrites the provider-native OrcaRouter namespace. Values such as `orcarouter/auto` must reach the provider in the exact form its API expects; add tests covering native identifiers instead of normalizing them through another provider's rules.
- The documentation/marketing currently describes the integration as default-deny, but the upstream/default behavior is audit unless an explicit policy is attached. Either attach and test the claimed deny policy, or change the claim to match the actual runtime behavior.
- The branch has no trustworthy fresh check run against current `main`; rebase and rerun the complete provider/unit/CodeQL set after fixing the two issues above.

Please move the PR back to Draft until those semantics are proven end to end.

wu-changxing:
Thanks for this, and sorry for the silence — the delay was on our side, not yours.

Your PR was sitting with no checks because GitHub holds workflow runs from first-time contributors until a maintainer approves them. That is now done: **Tests** and **CodeQL** are running against `c9418cc`. Nothing was wrong with your branch.

I read the diff while waiting. The implementation is clean and follows the existing `OpenRouterLLM` pattern closely — prefix routing in `create_llm()`, `ORCAROUTER_API_KEY` with a clear error when it is missing, JSON-mode `structured_complete()` rather than assuming the beta parse endpoint, and the module header updated so the provider list stays accurate. The tests and the `docs/concepts/models.md` entry are appreciated; plenty of provider PRs arrive without either.

Two notes, neither of which is a change request:

1. `calculate_cost(self.model, ...)` is called with the model name after the `orcarouter/` prefix is stripped, so the cost table needs an entry for it to report anything other than zero. This is the same for `OpenRouterLLM` today, so it is pre-existing behaviour rather than something you introduced — just flagging it so the usage numbers do not surprise you.
2. On our side, this makes five near-identical OpenAI-compatible provider classes (Groq, Grok, Mistral, OpenRouter, OrcaRouter). That is our cleanup to do, not yours, and it should not block this.

One thing I want to be straight with you about: adding a named provider means it appears in the README, the docs, and the `co create` model list, so it is a product decision as well as a code one. That call belongs to the maintainer, not to me, and I have passed it on. I will follow up here either way — you will not be left guessing again.

If CI comes back red I will post exactly which job and which assertion, so you do not have to reproduce it blind.

PR #844, opened 2026-08-11
[1.9 Deferred] ask_user: reach the owner by email when nobody is connected

### LINKED ISSUE #843: [1.9] A deployed agent cannot reach its owner when nobody is connected
## What happens

An agent running anywhere except an attached terminal has **no way to reach a human**. When it calls `ask_user`, it gets a string back saying nobody answered, and the run ends with the question unanswered.

`connectonion/useful_tools/ask_user.py` says so in its own comment:

```python
if not agent.io:
    # One-shot mode (co ai "prompt", and every deployed agent) has nobody to
    # answer.
```

`agent.io` is a WebSocket (`connectonion/network/io/websocket.py`, assigned at `http_router.py:48`), so a human has to be **connected at that moment** for the question to go anywhere. That covers exactly one situation: someone sitting in front of an open session. It does not cover `co ai "prompt"`, cron runs, or any deployed agent.

The current behaviour is correct as far as it goes — #553 and #535 established that an unanswered question must not read as approval, and the returned string says so explicitly. The gap is not that it answers wrongly. **It is that there is no channel at all.**

## Who this affects

Every unattended run, which is most of them:

- `co ai "prompt"` one-shots — no `io` by construction
- Scheduled/cron agents — nobody is watching at 3am
- Deployed agents — the situation the comment names

Concretely: the run in #837 built an entire event listing across five ticketing platforms, then hit a verification step it could not pass alone and stopped. Account created, organiser page built, form complete, and no way to say *"I need six digits, send them to me."* A human was awake and willing. There was no path from the agent to that human.

## The channel that already exists

Email needs nothing bought and no new dependency. All three parts are in the repo today:

- `send_email` — the agent sends from its own address ✅
- `get_emails` / `mark_read` — the agent reads its own inbox ✅
- `ask_user`'s existing `fields` contract already supports structured input including `type: "password"` — "paste the code here" is already expressible ✅

Every agent already has an address (`0x{10}@mail.openonion.ai`, Cloudflare inbound → webhook → DB), and it costs nothing because we own the domain.

So the flow **agent gets stuck → emails the owner → owner replies → agent reads the reply → continues** is buildable with zero purchases.

Implementation for this half is written and passing on branch `feat/ask-user-email-fallback`: `ask_user` falls back to emailing `OWNER_EMAIL`, polls the agent's own inbox for a reply, strips the quoted question, and on timeout or missing config returns the same not-approval string as today with the reason appended. 5 new unit tests, no new dependencies.

## What email does not give you: attention

Email is asynchronous. Nobody is watching an inbox at 3am. A run that stops at midnight and mails a question gets its answer at 9am — by which time the run is long dead.

**A phone call or SMS is the only channel that interrupts a human who is not looking.** That is its whole value here, and it is a real one — but it is an escalation on top of the email path, not a replacement for it. The reply can still come back by email.

Provider selection for that escalation is decided in #359 (Twilio, one shared outbound number, ~$3–8/month total, AU rates verified). This is deliberately **outbound-only**, which is what keeps it clear of everything #837 documented — those blockers are all on the inbound/OTP side.

## Scope

1. **Email fallback** (branch exists) — `ask_user` reaches the owner by email when there is no `io`, waits, returns the reply
2. **Urgent escalation** — an urgent question additionally rings or texts the owner; the answer still comes back by email
3. Not in scope: SMS as an inbound reply channel (that is #358), live voice conversation (that is #359)

## Definition of done

- An agent with no `io` can ask a question, get an answer from a human, and continue — proven end-to-end on a real deployed agent, not only in unit tests
- A question that is never answered still returns the not-approval string, with the reason (no owner configured / send failed / timed out) — the guarantee from #553/#535 must not regress
- Where the owner's address comes from is documented; today nothing in `.co/` records who the human is

## Related

- #837 — why the agent cannot instead have a number of its own for receiving verification codes (the research stands; see its repositioning comment)
- #359 — the phone channel design and the provider decision for the escalation
- #358 — SMS as an inbound channel, separate build
- #553 / #535 — established that unanswered must never read as approval

wu-changxing:
## Outcome

Add a deliberately narrow email fallback for `ask_user` when no live `agent.io` is attached. The historical no-IO behavior remains immediate and fail-closed unless the operator explicitly sets `CONNECTONION_ASK_USER_EMAIL=1`.

This PR stays Draft. The code has been security-reviewed and tested, but it must not merge until the merged oo-api change is deployed and the real authenticated send/reply round trip passes.

## Design decision

DD-043 records the decision and rejected alternatives:

- live `agent.io` remains authoritative and unchanged
- merely defining `OWNER_EMAIL` does not create an outward email side effect or a 15-minute wait
- email answers ordinary questions only and never satisfy Host/ACP tool permission
- persistent email is not a password, token, OTP, verification-code, or secret channel
- the original OTP-relay use case is intentionally deferred until retention/redaction has its own design

## Safe fallback contract

- explicit opt-in via `CONNECTONION_ASK_USER_EMAIL=1`
- machine-global `~/.co/keys.env` `OWNER_EMAIL` wins over a conflicting project-loaded value; deployed processes may provide the ambient value
- only bounded, unique, non-sensitive choices are allowed; free-form fields are rejected
- choices are sent with stable numeric IDs, and replies must contain only valid IDs; arbitrary reply text cannot become an answer
- common secret terms, 4–8 digit PIN/OTP shapes, common token prefixes, private keys/JWTs, and opaque mixed tokens are rejected before send
- secret-pattern detection is defense in depth, not a general classifier; the opt-in caller contract still forbids secret-bearing question/choice data
- every question has a random 128-bit `CO-ASK` reference; sender plus exact tag are checked again client-side
- oo-api must echo the exact applied subject filter. An older deployment that silently ignores the query parameter fails closed
- the HTML body is escaped and the subject is a single printable line, preventing body/header injection
- timeout/poll settings are bounded; polling is immediate then interruptible, each inbox request is capped at two seconds and cannot exceed the remaining deadline
- one process permits one pending request per hashed owner key and applies a cooldown
- every failure returns `NOT ANSWERED` and explicitly says it is not approval

## Backend dependency

openonion/oo-api#140 filters inside the authenticated mailbox before `LIMIT` and returns the exact `subject_filter_applied` capability proof. It was green and independently reviewed at exact head `a642bf9`, then squash-merged to `main` as `d329cc98`. Merging does not auto-deploy this repository; production still runs the previous deployment until a separate reviewed workflow dispatch or version tag.

## Security and code review

Review was iterative rather than ceremonial. Findings fixed include:

- implicit email/wait behavior
- password/OTP/token persistence
- project config overriding global owner contact
- HTML and subject injection
- unrelated/concurrent reply association
- fixed recent-message window loss
- cancellation races and unbounded HTTP-over-deadline waits
- old backend silently ignoring the filter
- free-form reply injection, ambiguous comma choices, and oversized numeric replies

Final independent review of exact head `22e2fedb`: 0 P0/P1/P2 findings. No dependency was added.

## Validation

- focused email/ask_user suite: 88 passed
- full locked unit suite excluding the cross-repository docs version assertion: 6535 passed, 13 skipped, 13 deselected
- ACP/MCP real stdio unit file in the locked dev environment: 25 passed
- production source Ruff: passed
- compileall: passed
- `git diff --check`: passed
- remote Linux 3.10–3.13, Windows browser/full E2E, lockfile, static analysis, and CodeQL checks: all passed on exact head

The excluded cross-repository version file was run separately: 6 passed and 1 failed because the local docs-site checkout advertises `1.6.0` while this branch is `1.7.0a1`. That remains an explicit release/docs coordination blocker, not a hidden test failure and not part of this email change.

## Remaining before promotion/merge

- owner review
- deploy merged oo-api commit `d329cc98` through its reviewed workflow
- run a real authenticated send, server-filtered reply, cancellation/timeout, and mark-read round trip
- keep this PR Draft if any of those steps is incomplete

Closes #843 when the backend dependency and real round trip are complete.
Depends on deploying openonion/oo-api#140 (`d329cc98`).


wu-changxing:
## A local test-run discrepancy, investigated and not fully closed

Recording this rather than leaving it in a terminal, because CI being green does not by itself settle it.

**What happened.** A full local `pytest tests/unit` run on this branch came back **2 failed, 5954 passed**, while CI was green on all four Python versions. The two failures were `test_importing_connectonion_stays_cheap.py::TestThePublicSurfaceIsUnchanged::test_a_star_import_still_works` and `test_cross_worker_replay.py::test_claim_waits_for_a_short_lived_worker_lock`.

**Why the first one deserved real suspicion.** This PR adds imports to `ask_user.py` — `send_email` and `get_emails`, which import `requests`, `yaml` and `dotenv` at module level — and `test_importing_connectonion_stays_cheap` is the guard from #588 that exists precisely to stop the startup path getting heavier (`co --version` was 9.5s before that work). A change that adds imports failing the import-cost guard is exactly the shape of a real regression.

**It is not one.** Measured directly, comparing `import connectonion` on this branch against the same measurement on `main`:

| | top-level modules added | heavy deps present |
|---|---|---|
| main | 41 | `dotenv` |
| this branch | 41 | `dotenv` |

Identical. `useful_tools/__init__` already exports `send_email` and `get_emails`, so this PR adds an edge inside a graph that was already being pulled — no new module reaches the startup path.

**What else was checked**

- Both tests pass in isolation — three consecutive runs of the two files, 47 passed each time
- A full `tests/unit` run on this branch in random order (the same condition that produced the failures): **5956 passed, 0 failed**
- CI: 11/11 green, including the full suite on Python 3.10, 3.11, 3.12 and 3.13

**Where that leaves it.** The failures did not reproduce in five subsequent attempts, and the one mechanism that would have implicated this PR is measurably absent. Both failing tests are of the kind that is sensitive to ordering and machine load — one spawns a subprocess to compare import graphs, the other waits five seconds on a lock — and the suite runs under `pytest-randomly`, so each run takes a different order.

**But "could not reproduce" is not "proven unrelated",** and I would rather a reviewer know that than read a clean bill of health. If the same two ever fail again, the import-graph measurement above rules out this PR's mechanism and the investigation should start at ordering and load instead.

wu-changxing:
Review update: fixed a blocking authorization-correlation bug in commit 7a9afb9c. A reply is now accepted only when both the normalized owner address and a fresh 128-bit CO-ASK subject tag match this exact wait. Unrelated owner mail and concurrent questions cannot consume one another. Transient inbox failures remain within the bounded wait, empty correlated replies stay NOT ANSWERED, and read-state cleanup cannot cause a resend. Local validation: 92 ask_user, tool-executor, and approval tests passed; the focused ask_user file is 19/19. Keeping this PR Draft: issue #843 still requires a real deployed-agent end-to-end run and an explicit v1 statement for where contact identity is configured.

wu-changxing (commented):
## Review result: keep this Draft; not merge-ready

CI is green and the per-request 128-bit subject tag fixes the earlier cross-question correlation bug. Send failure, timeout, empty reply, and inbox failure also fail closed. The following issues still block promotion/merge.

### P1 — do not collect password/secret fields over persistent email

`ask_user` explicitly supports fields such as `{"type": "password"}`. `_body()` now renders every field as an instruction to reply by email, and the raw reply becomes the tool answer. The inbound service stores `text_body`/`html_body` unchanged; `mark_read` does not delete or redact it. This silently removes the UI password-control boundary and leaves OTPs/passwords in the external mail provider, server inbox, and potentially agent/tool traces.

Fail closed for password/secret fields in the email fallback, or first design a dedicated secret channel with retention/redaction rules and tests.

### P1 — a contact route is being treated as authorization

The code accepts a reply when `from == OWNER_EMAIL` and the bearer tag is present. But `OWNER_EMAIL` is not bound to the authenticated agent owner. It is also loaded from the project cwd `.env` before `~/.co/keys.env` (`connectonion/__init__.py`), so project-local configuration silently wins over the global owner address mentioned in the error text.

For questions gating send/post/delete/deploy/spend, document and enforce who is authorized to answer and the configuration precedence. A contact address alone cannot become approval authority by accident; unauthorized replies must fail closed.

### P1 — the claimed plain-text body is sent as HTML

`_body()` concatenates untrusted question, option, and field-label text without escaping. The existing backend sends `body` to Resend as `html`, not text (`oo-api/email_service/service.py`). A remote/tool-derived `<a>` or `` therefore becomes active content sent from the agent's official address. Escape content before a safe wrapper or add a true text-body backend contract; strip CR/LF/control characters from the subject and add hostile-input tests.

### P2 reliability/abuse gaps

- No-IO behavior changes from immediate fail-closed to an uninterruptible `time.sleep()` polling loop for up to 900 seconds. Make it explicit/opt-in, configurable, cancellable, and test worker shutdown.
- `get_emails(last=20)` can permanently miss the correlated reply if more than 20 newer messages arrive. Use tag filtering/server query or a cursor/pagination strategy and test a busy inbox.
- Every concurrent/repeated `ask_user` sends another outward email. Add per-owner/per-agent rate limits or a bounded pending-request policy and concurrency tests.

Issue #843 also still requires a real deployed-agent end-to-end test. Until these points and the owner/contact design are resolved, this should remain Draft and must not be merged.

wu-changxing:
Current review of Draft head `22e2fedb`: the earlier P1s are substantially addressed — fallback is opt-in, limited to non-secret choices, HTML is escaped, replies are correlated by a high-entropy tag, inbox filtering is server-side, and the wait is bounded/cancellable.

The remaining abuse control is only process-local. `_PENDING_OWNER_KEYS` and `_LAST_OWNER_ATTEMPT` cannot prevent two deployed workers or two separate agents from emailing the same owner concurrently, which is the normal topology this feature must survive. `_LAST_OWNER_ATTEMPT` also never evicts digests in a long-lived multi-tenant process. Enforce the cooldown/pending policy at the shared backend/account-recipient boundary (ideally with an idempotency key), or explicitly narrow the guarantee and add a bounded cache.

This should remain Draft until the shared-rate-limit decision and the real deployed-agent E2E in #843 are complete. The newly required `blog-gate` is also currently failing on this head.

wu-changxing:
Closing this Draft implementation rather than carrying a conflicted security-sensitive branch as active release work. The offline-owner problem remains valid in #843, but the review identified unresolved secret-field retention, authorization, HTML injection, cancellation, inbox-window, and abuse-control blockers. Redesign against the unified approval boundary before opening a fresh implementation PR.

wu-changxing:
Release target correction from Aaron: the offline-owner email fallback is **deferred to 1.9**, not cancelled. Keep #843 as the canonical requirement. This implementation PR remains closed because it must be rebuilt from current main against the unified approval/secret-retention design.

Issue #848, opened 2026-08-11
The account guard only covers the CLI; 17 direct reads of OPENONION_API_KEY bypass it


wu-changxing:
Follow-up to #845. Same shape as the bug that PR fixes, in the layer that actually spends money.

## The pattern

`_token_for_this_account()` (`project_cmd_lib.py:1090`) is the check that a token belongs to the account whose key this machine holds. It is reachable only through `load_api_key()`. #845 fixes the early return that made it dead code, but only for callers that go through `load_api_key()` at all.

Six files do:

```
cli/commands/{server,email,status,auth,deploy}_commands.py, project_cmd_lib.py
```

Meanwhile **17 call sites read the variable straight from the environment** and never see the guard:

```
useful_tools/get_emails.py:38,107,158,205    useful_tools/send_email.py:79
useful_tools/gmail.py:174                    useful_tools/outlook.py:132
useful_tools/gdrive.py:130                   useful_tools/google_calendar.py:122
useful_tools/microsoft_calendar.py:104       useful_tools/transcribe.py:71
core/llm.py:1059                             cli/browser_agent/agent.py:26,31
cli/commands/doctor_commands.py:415          useful_plugins/image_result_formatter.py:159
```

That is the tool layer — the code that sends mail, calls the model, and drives the browser. It is where a wrong account is billed and where a wrong mailbox is read.

## What this means concretely

Verified against the #845 branch, with a token for account B in the environment while the machine holds the key for account A:

- `co email inbox` — **refused.** `_require_auth()` → `load_api_key()` → guard fires.
- `get_emails()` called as a library function — **authenticates as B.** No guard on the path.

So #845 closes the CLI hole and leaves the library hole. Anyone using ConnectOnion as a library — which is the documented way to use it — is unprotected.

`cli/browser_agent/agent.py:24-32` deserves its own mention: `resolve_api_key()` is a second, independent implementation of `load_api_key()`'s file walk, without the guard. This is the path behind the "browser clicks report insufficient credit" reports — the NL click was billing a drained account that a stray `.env` had supplied.

## Suggested direction

One accessor that every caller uses, guard included, rather than 17 copies of `os.getenv`. `resolve_api_key()` in the browser agent should be deleted in favour of it.

The general lesson from #845 is worth stating: a guard reachable through only one of many call paths is not a guard, and it is harder to spot in review than a missing one, because the code plainly contains the check.

wu-changxing:
[...]

Issue #853, opened 2026-08-11
CI installs test dependencies from a hand-written list, not the dev extra — they drift


wu-changxing:
`.github/workflows/tests.yml` installs its test dependencies by naming them:

```yaml
# test job
pip install -e .
pip install pytest pytest-cov pytest-mock pytest-asyncio psutil

# windows-browser job
pip install pytest pytest-mock pytest-asyncio
```

`.github/workflows/release.yml` installs the declared set:

```yaml
python -m pip install --upgrade pip build twine ".[dev]"
```

So the repository already disagrees with itself about where test dependencies come from, and `pyproject.toml`'s `dev` extra is authoritative for exactly one of the three places pytest runs.

## Why this bites

A contributor adding a test dependency does the correct thing — adds it to the `dev` extra, updates `uv.lock`, verifies locally — and CI still does not have it.

Hit while adding `pytest-timeout` (#850 / PR #851). The resulting failure:

```
Run all tests    ->  failure in 44s
```

pytest.ini sets `timeout = 300` and the suite runs under `--strict-config`, so a runner without the plugin **exits on an unknown ini option before collecting a single test**. The log shows no test names and no assertion — it looks like the branch is catastrophically broken rather than like one missing package. The distance between the error and its cause is the expensive part.

Any future dev-extra addition fails the same way, or worse: a dependency that only *some* tests import produces a partial, confusing failure instead of a clean stop.

## Suggested fix

Have `tests.yml` install `".[dev]"` in both jobs, the way `release.yml` already does, so there is one list. If a job genuinely needs a narrower set, it should say so in a comment explaining why — `windows-browser` may be a real case, since it only runs two files.

Anything the extra does not cover (`psutil` is currently CI-only and not in the extra) should move into the extra, or the divergence just moves.

wu-changxing:
Adding a second failure mode to this issue, which I think is the more important one.

## "No CI" and "all green" look identical

Neither shows a red mark. Both let a PR sit there looking ready to merge.

PR #854 was opened with another branch as its base and reported:

```
no checks reported on the 'fix/browser-agent-key-guard' branch
```

Zero jobs ran. No warning, no annotation, nothing in the PR body — the checks section is simply absent, which reads as "not started yet".

[...]

PR #854, opened 2026-08-11
Route the browser agent's key through the account check

wu-changxing:
First reviewed slice of #848. This PR does not close the broader credential
accessor audit.

## Problem

The browser agent carried a second API-key lookup that read
`OPENONION_API_KEY` directly and bypassed the account check in
`load_api_key()`. Browser model calls therefore could bill a different account,
and the daemon cached that wrong token for its lifetime.

## Change

- Remove the duplicate environment/dotenv lookup.
- Delegate `resolve_api_key()` to the centralized `load_api_key()` path.
- Preserve the existing empty-string contract when no valid key is available.
- Keep failed reauthentication fail closed: the foreign token is never handed
  to the browser agent.

PR #845 is now merged, and this branch is integrated with current `main`, so
the diff contains only the browser entry point and its focused tests. It also
inherits the reviewed malformed-JWT claim handling from #845.

## Validation

- 26 browser-key, token-identity, and project-command tests passed locally.
- 87 browser-daemon E2E tests passed with local Unix sockets enabled.
- The previous head passed Linux Python 3.10–3.13, three Windows browser jobs,
  Windows E2E, lockfile, Python analysis, and CodeQL.
- Current integrated-head CI is required before merge.

## Remaining work

#848 remains open for the other direct credential readers. The durable next
step is one narrow credential accessor with explicit fail-closed semantics,
not sixteen copied lookups.


wu-changxing:
Retargeted from `fix/identity-resolution-2` to `main`, because **CI does not run on stacked PRs**: `.github/workflows/tests.yml` triggers on `pull_request: branches: [main, master]` only, so with the #845 branch as base this PR reported `no checks` — zero coverage, silently.

Consequence of the retarget: this PR now shows **two commits**, the second of which is #845's. Review only `Route the browser agent's key through the account check`; the other is #845 and is reviewed there. Once #845 merges, this diff reduces to its own commit automatically.

The tradeoff was no-CI-but-clean-diff versus CI-with-a-borrowed-commit. CI wins — this touches the path that spends money, and the whole point of the change is that unverified assumptions about which account is in play are expensive.

wu-changxing:
Second instance of the same silent failure, worth recording since #853 is about exactly this.

Retargeting this PR from the #845 branch to `main` did **not** trigger CI. Changing a PR's base fires the `edited` activity type, which is not in the default `pull_request` trigger set (`opened`, `synchronize`, `reopened`). So the PR sat with a corrected base and still zero checks — and reported it the same way as before: `no checks reported`.

I only caught it because I went back to verify the retarget had worked. Had I trusted the fix, this would have looked resolved while remaining completely untested.

Closed and reopened to fire `reopened`. Jobs are running now.

That is twice in one PR that the absence of CI was indistinguishable from its presence — first from the base branch filter, then from the trigger-type default. Both times the fix looked like it had worked.

Issue #816, opened 2026-08-10
Several tools have no command line: calendar, and outbound Discord/Telegram/WhatsApp


wu-changxing:
## The pattern that already exists

`co` gives a person terminal access to the accounts they already own:

```
co gmail    co outlook    co gdrive    co syno    co email
```

#250 closed on exactly this reasoning — *"Gmail has a full tool but no CLI surface"*. The same sentence is true of several things today.

## 1. Calendar — a full tool on both providers, and no way to reach it

The clearest case, because nothing needs building except the surface:

```
$ ls connectonion/useful_tools/ | grep calendar
google_calendar.py      microsoft_calendar.py
```

Both expose the same eight methods — `list_events`, `get_today_events`, `get_event`, `create_event`, `create_meet` / `create_teams_meeting`, `update_event`, `delete_event`, `get_upcoming_meetings` — and **neither has a single line of CLI**. `co auth google` and `co auth microsoft` already request the Calendar scope, so the credentials are sitting there unused by the terminal.

An agent can read your calendar. You cannot.

```
co calendar                      # today, both providers
co calendar week
co calendar new "Standup" tomorrow 9am --meet
co calendar cancel <#>
```

Provider selection should follow `co gmail` / `co outlook` rather than inventing a flag: `co gcal` and `co mscal`, or one `co calendar` that reads whichever is connected. Worth deciding in the issue, not in review.

## 2. Chat platforms — no outbound surface at all

`co` can send email three ways and cannot send a chat message anywhere.

This is **not** what #341 (Discord), #352 (Telegram) and #349 (WhatsApp) cover. Those are *inbound* — a platform triggering `agent.input()`. This is the opposite direction: a person or a script sending a message out.

```
co discord  send  <channel>  "message"
co telegram send  <chat>     "message"
co whatsapp send  <number>   "message"
```

### The evidence that this is missing is in our own release process

Announcing 1.6.0 to Discord tonight meant hand-assembling JSON and curling a webhook, and the release skill has accumulated workarounds for that path rather than a tool:

> Inline `curl -d "{…}"` breaks on the backtick/backslash escaping (Discord replies 50109 "invalid JSON"), and python urllib gets 403 (Discord rejects its default User-Agent). **json.dumps → file → curl is the path that works.**

[...]

Issue #352, opened 2026-07-31
Make Telegram a real inbound channel — the easy one (official Bot API, no business verification)


wu-changxing:
## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

## Motivation

Same gap as email (#336), Discord (#341), WhatsApp (#349): a messaging platform that could trigger `agent.input()` but doesn't. Confirmed nothing exists for Telegram anywhere in the repo (the only hit for "telegram" is #341 mentioning it in passing as a future candidate).

**This is the easiest of the four.** Telegram's official Bot API needs no business verification (unlike WhatsApp), is free, and — notably — supports **long-polling (`getUpdates`) as a first-class built-in mode**, not just webhooks. That's relevant beyond this issue: it means a bot can run with zero public-facing infrastructure at all, just a process that polls Telegram's own servers.

## Scope

Either transport mode is legitimate; pick whichever is cheaper to stand up first:

```
Telegram message to the bot
        │
        ▼
Telegram Bot API — webhook push OR long-polling (getUpdates), bot's choice
        │
        ▼
agent.input(message.text)
        │
        ▼
reply sent back via the Bot API to the same chat
```

- Identity: Telegram user IDs are platform-authenticated (same category as Discord — the question is which users/chats are allow-listed, not whether the identity claim can be forged)
- No threading question — Telegram chats are already per-conversation
- Media out of scope for v1, same as WhatsApp/Discord

## Relationship to other issues

- #336 (email), #341 (Discord), #349 (WhatsApp) — same category, standalone implementations by design (see #341's note on why not to abstract yet)
- Milestone 2.2 ("trigger runtime and capability discovery") groups this with the other inbound-trigger work

## Definition of done

- Sending a Telegram message to the bot causes `agent.input()` to run and a reply to come back in the same chat
- Only allow-listed users/chats can trigger it
- A test that doesn't require a live Telegram connection (mocked update payload → `agent.input()` call)

wu-changxing:
Sequencing: Wave 2 — no hard code dependency, but best done once #336 exists as the reference implementation to copy the shape from. This is the fastest of the channel issues (no business verification needed) — a good candidate for an early quick win within wave 2.

[...]

PR #866, opened 2026-08-11
telegram: send a message from the terminal and from an agent

wu-changxing:
## Summary

Implement the Telegram slice of #816 as one small outbound capability:

```bash
co telegram send <chat-id-or-@channel> "message"
```

The same operation is available to agents and Python callers as
`connectonion.send_telegram(chat, message)`.

## Design decision

The bot belongs to the user and is created through Telegram's @BotFather. Its
`TELEGRAM_BOT_TOKEN` therefore lives in the user's `~/.co/keys.env`, alongside
other user-owned credentials. ConnectOnion calls Telegram directly; no
OpenOnion carrier credential or credits are involved.

This PR deliberately supports plain-text outbound messages only. Inbound bot
events remain #352, and Discord/WhatsApp remain separate #816 slices so each
API and credential boundary can be reviewed independently.

## What changed

- add `send_telegram()` with a bounded success/failure result;
- add `co telegram send` with script-friendly exit status;
- expose the tool from `connectonion` and `connectonion.useful_tools`;
- discover the token in `co status`, `co doctor`, and `co keys`;
- mask the token by default and reveal it only through the existing explicit
  `co keys --reveal` flow;
- document BotFather setup, credential placement, CLI/tool usage, errors, and
  the plain-text boundary.

## Security and reliability review

- Telegram places the bot token in the request URL. Transport exceptions are
  reduced to their exception class; their string/URL is never returned.
- Non-JSON, non-object, refused, and malformed-success responses fail safely.
- Telegram descriptions are bounded and any echoed exact token is redacted;
  raw response bodies and URLs are never surfaced.
- Telegram/chat values are rendered as Rich `Text`, not parsed as terminal
  markup.
- No live bot or external message was used during tests.

## Validation

- `tests/unit/test_telegram.py` — 14 passed;
- CLI routing, credential, status, doctor, import-cost, and help-focused run —
  147 passed;
- `tests/unit/test_telegram.py tests/e2e/cli` — 428 passed, 3 headful-browser
  tests skipped because no display was present, 13 unrelated tests deselected;
- changed Python modules compile successfully;
- `git diff --check` passes;
- second simplicity/security review found no P1/P2 issue.

GitHub's full cross-version, Windows, lockfile, and CodeQL checks are the final
merge gate.


wu-changxing:
Second review is complete. The two material blockers from the first pass are
fixed:

1. request exceptions and raw responses can no longer expose Telegram's
   token-bearing URL;
2. Telegram/chat values are rendered as literal Rich `Text`.

Credential discovery/masking, the public export, CLI routing, and setup docs
are also covered. Local validation is 428 passed / 3 display-only skips for
the complete CLI regression plus the Telegram tool tests. The PR remains
Draft until all GitHub checks finish.

Issue #838, opened 2026-08-10
[RFC] Make backend event naming ACP-native


wu-changxing:
## Status and architecture

This RFC remains open. Seven additive Host/browser slices are implemented:

- tool notifications: Host #842 / React openonion/connectonion-react#12
- final agent messages: Host #870 / React openonion/connectonion-react#14
- permission requests: Host #873 / React openonion/connectonion-react#16 and #17
- cancellation: Host #880 / React openonion/connectonion-react#19
- mode observation: Host #882 / React openonion/connectonion-react#21
- mode transactions: Host #884 / React openonion/connectonion-react#23
- persisted public thoughts: Host #886 / React openonion/connectonion-react#27

The active frontend path is:

> ConnectOnion Host -> `@connectonion/react` -> O Chat

`@connectonion/react` owns browser connection, protocol selection/normalization, typed state, and hooks. O Chat consumes normalized state and performs presentation/product actions only. The standalone `connectonion` TypeScript client is retired: it is not an O Chat dependency or release prerequisite and receives no new parallel ACP work.

`@connectonion/react@0.4.2-alpha.2` is published and verified; O Chat #142 is merged and pins it exactly. React/O Chat native `/acp` migration is complete in #893; `/ws` remains the bounded fallback only when the discovery descriptor is absent.

This issue concerns **ACP — Agent Client Protocol**, not MACP. Official schema reference: https://agentclientprotocol.com/protocol/v1/schema

## Governing rule

> Use the ACP name directly when semantics are identical. Convert only when semantics are actually different.

Matching spelling is insufficient. Ordering, ownership, cardinality, cancellation, permission, replay, terminal behavior, and authority must match before concepts are declared equivalent.

1. Exact semantic match: use the ACP DTO at the public protocol boundary.
2. Stable existing API: introduce a versioned ACP form, retain a bounded compatibility reader, and document removal separately.
3. Different semantics: keep an explicit ConnectOnion domain name and translate through one small pure adapter.
4. No ACP peer: keep the event internal rather than forcing cosmetic parity.
5. Persisted state/traces: do not silently rewrite historical logs, eval inputs, or tool snapshots.
6. Authority: no protocol field, plan, mode, or restored value grants a capability by itself.
7. Browser ownership: O Chat calls typed React operations/state and never constructs or decodes ACP/legacy frames.
8. Stable vs unstable schema: version-pinned stable v1.19 forms ship independently of generated experimental fields.

## Completed slices

[...]

PR #870, opened 2026-08-11
feat: emit stable ACP agent messages

### LINKED ISSUE #867: ACP Host: emit final agent messages as stable agent_message_chunk notifications
## Parent and architecture

Child slice of #838. This issue covers the Python Host writer only.

The active browser release chain is:

`connectonion` Host → `@connectonion/react` → O Chat

`@connectonion/react` is the primary browser SDK and owns connection, ACP decoding, normalized chat state, and hooks. O Chat consumes that package and remains protocol-free. The retired standalone TypeScript SDK is not part of this slice or the frontend release plan.

Companion frontend issue: openonion/connectonion-react#13  
Merged frontend reader: openonion/connectonion-react#14  
Host implementation: #870

## Final design decision

Adopt ACP v1.19 `agent_message_chunk` when, and only when, the Host has a newly persisted final assistant text message with stable domain identity. Keep the authenticated ConnectOnion WebSocket and the versioned `ACP_NOTIFICATION` carrier from DD-035; the surrounding socket is not presented as a complete ACP transport.

Expert review found that reconstructing identity from `msg-{array index}` was unsafe because compaction can shift indexes and cause a later message to overwrite an earlier React card. The accepted design therefore persists a transport-neutral UUID on every new terminal assistant message.

| Concern | Decision |
|---|---|
| ACP shape | Exact schema-v1.19.0 `session/update` containing `agent_message_chunk` and a text content block |
| Cardinality | One complete persisted assistant answer is one ACP chunk; token streaming is out of scope |
| Identity | Persist one transport-neutral UUID and reuse it as both `ChatItem.id` and ACP `messageId` |
| Provider boundary | Shallow-copy provider-bound messages and remove only the top-level domain `id`; preserve nested tool-call IDs and canonical history |
| Ordering | Send the ACP notification immediately before the existing authoritative `OUTPUT` |
| Writer authority | Only the terminal `OUTPUT` path may mirror an agent message; the live loop mirrors tool lifecycle events only |
| Reconnect | Replay/re-delivery converges on the persisted UUID and React upserts the same item |
| Legacy compatibility | Historical assistant messages without UUIDs keep the index fallback for legacy display and receive no fabricated ACP identity |
| Failure behavior | Empty, unpersisted, mismatched, malformed, or conversion-failed candidates still drain through legacy `OUTPUT` |
| Rollback | Stop emitting the additive notification; stored domain IDs remain valid and no protocol migration is required |

DD-036 records this decision. It extends DD-028 (one complete provider response may be one legal ACP chunk) and DD-035 (versioned additive carrier, detached DTOs, bounded rollback).

## Implemented scope

- Serialize `AgentMessageChunk` with the official `agent-client-protocol==0.12.0` model.
- Persist a unique UUID on every new terminal assistant message.
- Reuse that UUID for reconstructed `ChatItem.id` and ACP `messageId`.
- Select only the final persisted assistant message that matches the completed result.
- Emit the ACP notification before both normal and stored-completion `OUTPUT` paths.
- Keep arbitrary/plugin-produced live assistant events from creating duplicate message mirrors.
- Sanitize provider-bound history in the main Agent loop and both debug preview paths.
- Share one message fixture with `@connectonion/react`.
- Validate the packed React reader against O Chat without adding protocol code to O Chat.

## Non-goals

- Token-by-token LLM streaming.
- ACP thought, mode, permission, plan, cancel, or stop-reason mapping.
- Parsing or rendering ACP in O Chat components.
- Updating the standalone TypeScript SDK.
- Publishing a release.

## Acceptance gates

- [x] Official Python models serialize the exact schema-v1.19.0 payload.
- [x] ACP and authoritative `chat_items` use the same persisted UUID and text.
- [x] Exactly one visible agent card results from ACP plus legacy `OUTPUT` in `@connectonion/react`.
- [x] Compaction, reconnect, and re-delivery do not duplicate or change the answer ID.
- [x] Empty, missing, unpersisted, or mismatched terminal text sends legacy `OUTPUT` without an ACP mirror.
- [x] A mirror conversion error cannot suppress `OUTPUT` or invite re-execution.
- [x] Top-level domain IDs never leak to model providers; nested tool IDs remain intact.
- [x] Full Host unit suite, package build, dependency audit, React/O Chat integration, and expert code review are green.

No release is performed by this issue.

wu-changxing:
## Summary

- persist a stable UUID on every assistant message and reuse it across session replay
- emit ACP v1.19 `agent_message_chunk` notifications immediately before the authoritative host `OUTPUT`
- mirror only persisted final assistant output; keep live ACP mirroring limited to tool events
- strip ConnectOnion's top-level message identity before all provider calls, including debug flows, without mutating canonical history or nested tool-call IDs
- document the decision and share one protocol fixture with `@connectonion/react`

This keeps the architecture boundary explicit: the Python host owns ACP transport, `@connectonion/react` is the browser SDK, and O Chat consumes React without parsing ACP itself.

## Design

See `docs/design-decisions/036-stable-acp-agent-message-identity.md`.

Legacy sessions without persisted assistant IDs continue to receive the existing `OUTPUT` carrier; they do not receive a fabricated ACP identity. A final ACP notification is emitted only when the completed result matches the persisted final assistant record.

## Validation

- full unit suite: `6311 passed, 13 skipped, 15 deselected`
- focused ACP/session/debug regression suite: `179 passed`
- isolated socket-heavy regression suite: `110 passed`
- production-file Ruff checks passed (existing repository-wide style debt excluded)
- `git diff --check` and `compileall` passed
- wheel and sdist built successfully; the wheel contains the new sanitizer and DD-036
- dependency audit: no known vulnerabilities in the exercised Python environment
- expert code review: no findings after provider-boundary and session-identity fixes
- paired React PR: openonion/connectonion-react#14 (merged, 67 tests); packed React integration in O Chat: 65 tests, lint, and production build passed

Closes #867
Related to #838

### DIFF
--- a/connectonion/core/acp_wire.py
+++ b/connectonion/core/acp_wire.py
@@ -11,7 +11,12 @@ from __future__ import annotations
 from typing import Any, Mapping
 
 from acp import text_block, tool_content
-from acp.schema import SessionNotification, ToolCallProgress, ToolCallStart
+from acp.schema import (
+    AgentMessageChunk,
+    SessionNotification,
+    ToolCallProgress,
+    ToolCallStart,
+)
 
 from .wire_events import normalize_wire_event
 
@@ -33,6 +38,20 @@ def map_tool_event(
     if event_type == "tool_call":
         return _tool_start(normalized)
     return _tool_update(normalized)
+
+
+def map_message_event(
+    event: Mapping[str, Any],
+) -> AgentMessageChunk | None:
+    """Map one complete Host assistant message to an ACP text chunk."""
+
+    if event.get("type") != "assistant":
+        return None
+    return AgentMessageChunk(
+        session_update="agent_message_chunk",
+        message_id=_required_string(event, "id"),
+        content=text_block(_required_string(event, "content")),
+    )
 
 
 def _tool_start(event: Mapping[str, Any]) -> ToolCallStart:
@@ -68,9 +87,9 @@ def _tool_update(event: Mapping[str, Any]) -> ToolCallProgress:
 def acp_notification_frame(
     event: Mapping[str, Any], session_id: str
 ) -> dict[str, Any] | None:
-    """Return a detached ConnectOnion carrier for one ACP tool update."""
+    """Return a detached ConnectOnion carrier for one supported ACP update."""
 
-    update = map_tool_event(event)
+    update = map_tool_event(event) or map_message_event(event)
     if update is None:
         return None
     notification = SessionNotification(session_id=session_id, update=update)
--- a/connectonion/core/agent.py
+++ b/connectonion/core/agent.py
@@ -22,6 +22,7 @@ from ..prompts import load_system_prompt
 from .events import EventHandler
 from .interrupt import run_interruptible
 from .llm import LLM, TokenUsage, create_llm
+from .provider_messages import messages_for_provider
 from .tool_executor import execute_and_record_tools, execute_single_tool
 from .tool_factory import create_tool_from_function, extract_methods_from_instance, is_class_instance
 from .tool_registry import ToolRegistry
@@ -547,7 +548,11 @@ class Agent:
             if response is not None:
                 if not response.tool_calls:
                     content = response.content or ""
-                    self.current_session['messages'].append({"role": "assistant", "content": content})
+                    self.current_session['messages'].append({
+                        "role": "assistant",
+                        "content": content,
+                        "id": self._next_trace_id(),
+                    })
                 else:
                     # Process tool calls
                     self._execute_and_record_tools(response.tool_calls)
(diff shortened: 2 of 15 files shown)


wu-changxing:
## CI follow-up

The first GitHub matrix run exposed a test-isolation bug in six new `test_asgi` assertions: the tests compared the entire WebSocket send list, while the full `tests/` ordering legitimately enabled an existing `DASHBOARD_SNAPSHOT` broadcast after `OUTPUT`. All Python versions failed on the same extra frame; the ACP/OUTPUT payloads and order were correct.

Commit `9a8f3287` scopes those assertions to the frame families owned by this feature, while still checking exact ACP cardinality, ACP-before-OUTPUT order, legacy-only fallback, and stray-assistant behavior. Production code is unchanged.

Validation after the fix:

- focused forwarder class: 11 passed
- exact GitHub Actions selection (`tests/`, same marker expression): 6,809 passed, 18 skipped, 194 deselected
- `git diff --check`: clean

Waiting for the restarted GitHub matrix before merge.

Issue #875, opened 2026-08-11
Real CLI smoke tests must use tool-specific config without exposing HOME


wu-changxing:
## Problem\n\nThe opt-in real Claude Code and Codex tests are collected under the repository's global autouse HOME isolation. Their module-level checks see the user's installed/login state before fixtures run, but the spawned CLI later sees an empty temporary HOME.\n\nObserved consequences:\n\n- Claude Code was reported as not logged in even though the real config was never presented to it\n- Codex HAS_AUTH was true from the real auth file, while app-server ran with an empty Codex home and returned 401\n- authentication failures can be misclassified because Claude currently reports expired OAuth as "OAuth access token has expired", not only "login" or "authentication"\n\n## Design discussion\n\nKeep the global HOME isolation for every normal test. Only tests explicitly selected with -m real_api may point each CLI at its own official config directory:\n\n- CODEX_HOME -> the pre-fixture Codex config directory\n- CLAUDE_CONFIG_DIR -> the pre-fixture Claude config directory\n\nDo not restore the entire HOME. Do not copy, parse, print, or fixture credential files. The child CLI remains the only credential consumer.\n\nAuthentication unavailable or expired is an environment skip for Claude's opt-in smoke test; protocol, parse, timeout, resume-identity, and provider errors still fail.\n\nOfficial references:\n\n- Codex app-server auth endpoints: https://github.com/openai/codex/blob/main/codex-rs/app-server/README.md#auth-endpoints\n- Claude CLAUDE_CONFIG_DIR: https://code.claude.com/docs/en/env-vars\n\n## Acceptance criteria\n\n- default suite retains isolated HOME\n- only real_api tests set tool-specific config roots\n- no credential content enters output or repository files\n- Codex real reply and resume run with the actual logged-in CLI\n- Claude missing/expired auth is reported as skip; authenticated reply/resume still must pass\n- unit and lint suites remain green\n\n## Release impact\n\nReal-client evidence must distinguish code failures from unavailable local authentication before 1.7.0 alpha decisions.

Issue #874, opened 2026-08-11
Codex app-server: refresh managed auth before starting or resuming a thread


wu-changxing:
## Problem\n\nThe real Codex E2E on Codex CLI 0.145.0 found a release-blocking integration regression:\n\n- codex login status reports ChatGPT login\n- direct codex exec in read-only mode returns pong\n- ConnectOnion's CodexAppServer completes initialize, thread/start, and obtains a real thread id\n- turn/start then fails with 401 Missing bearer or basic authentication\n- after initialize, calling account/read with refreshToken: true makes the same real app-server turn complete\n\nThis means the CLI and network are healthy; our app-server lifecycle is missing Codex-managed auth refresh.\n\n## Design discussion\n\nUse Codex's official app-server auth surface. After initialize/initialized and before thread/start or thread/resume, request:\n\n    account/read {"refreshToken": true}\n\nCodex owns credential storage and refresh. ConnectOnion must not read, copy, parse, log, or inject auth.json or access tokens.\n\nOfficial reference:\nhttps://github.com/openai/codex/blob/main/codex-rs/app-server/README.md#auth-endpoints\n\nWhy this design:\n\n- managed ChatGPT OAuth stays inside Codex\n- API-key and external auth modes retain Codex's own behavior; refreshToken is ignored where not applicable\n- the adapter detects auth problems before creating a turn\n- resume reapplies the same lifecycle because each tool call starts a fresh app-server process\n\nRejected:\n\n- reading auth.json: credential exposure and format coupling\n- copying bearer tokens into environment variables: expands secret lifetime and leak surface\n- falling back to codex exec: loses native step events, permission callbacks, and the app-server resume contract\n- swallowing account/read failure: defers a precise lifecycle error into a confusing model-turn 401\n\n## Acceptance criteria\n\n- unit contract proves account/read refresh occurs after initialized and before thread/start/resume\n- an account/read error returns a stable provider error without starting a thread\n- no token values enter application logs, envelopes, or test fixtures\n- real Codex app-server smoke returns pong\n- real multi-turn test resumes the same session\n- existing command/file event and permission tests remain green\n\n## Release impact\n\nBlocks the 1.7.0 alpha recommendation until fixed and real-binary E2E passes.

PR #876, opened 2026-08-11
fix: refresh Codex app-server managed auth

wu-changxing:
## Summary\n\n- refresh Codex-managed authentication through the official app-server account/read RPC after initialization and before thread start/resume\n- fail before creating a thread when account refresh itself fails\n- let explicitly selected real_api tests use CODEX_HOME and CLAUDE_CONFIG_DIR while preserving the default isolated HOME\n- classify Claude's current expired OAuth response as unavailable authentication, matching the existing skip policy\n\nCloses #874.\nCloses #875.\n\n## Design and security\n\nCodex remains the only credential owner. ConnectOnion never reads, copies, parses, logs, or injects auth.json or bearer tokens. The adapter only asks Codex to refresh its own managed authentication.\n\nNormal tests still receive an empty temporary HOME. Only opt-in real_api tests point each CLI at its tool-specific config directory; the rest of the user's HOME is not restored.\n\nOfficial references:\n\n- https://github.com/openai/codex/blob/main/codex-rs/app-server/README.md#auth-endpoints\n- https://code.claude.com/docs/en/env-vars\n\n## Verification\n\n- focused Codex/co-ai/Claude/paid-test gate suite: 131 passed\n- final Codex unit/co-ai rerun: 59 passed\n- real Codex CLI 0.145.0: 2 passed\n  - app-server handshake plus real pong\n  - resume of the same thread and remembered context\n- real Claude Code 2.1.227: correctly reached the actual config, then 2 skipped because the local OAuth token is expired\n- ruff: passed\n- git diff --check: passed\n\n## External follow-up\n\nA user must re-authenticate Claude Code before its real reply/resume checks can pass. This PR does not attempt login or weaken authentication.

### DIFF
--- a/connectonion/useful_tools/codex.py
+++ b/connectonion/useful_tools/codex.py
@@ -118,6 +118,7 @@ def codex(prompt: str, session_id: str = "", cwd: str = "",
     try:
         client.start()
         client.initialize(timeout=_remaining(deadline))
+        client.refresh_account(timeout=_remaining(deadline))
         approval_policy = "untrusted" if approval == "manual" else "never"
         if session_id:
             sid = client.resume_thread(
@@ -470,6 +471,10 @@ class CodexAppServer:
                                                             "item/reasoning/textDelta"]},
         }, timeout=timeout)
         self._notify("initialized", {})
+
+    def refresh_account(self, timeout=60):
+        """Let Codex refresh its managed auth without exposing credentials."""
+        return self.request("account/read", {"refreshToken": True}, timeout=timeout)
 
     def start_thread(
         self,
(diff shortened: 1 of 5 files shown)


wu-changxing (commented):
Self-review complete: no blocking findings. The production change uses Codex app-server's official account/read RPC with refreshToken=true and never reads, copies, or logs credentials. Refresh failure is fail-closed before thread start/resume. Real smoke tests restore only CODEX_HOME / CLAUDE_CONFIG_DIR under the explicit real_api gate; normal HOME isolation remains intact. Verified locally: 131 focused tests passed, 59 Codex-focused tests passed, real Codex reply + resume passed, ruff and diff checks passed. GitHub CodeQL, Python 3.10-3.13, Windows, and E2E checks are all green. Claude's local OAuth is expired and is correctly reported as an external auth skip.

PR #880, opened 2026-08-11
feat: accept negotiated ACP Host cancellation

### LINKED ISSUE #878: Host: accept negotiated ACP session/cancel without duplicate interrupts
Parent: #838

## Context

The hosted browser path still sends a ConnectOnion INTERRUPT frame directly from O Chat. The ACP stdio server already maps session/cancel to the generation-scoped interrupt lifecycle from DD-025, but the authenticated Host WebSocket has no exact ACP client-notification adapter.

This is a Host carrier slice, not a claim that the WebSocket is an ACP connection. The active frontend path is Host -> @connectonion/react -> O Chat. The retired standalone TypeScript SDK is out of scope.

## Best-practice decision before implementation

1. Keep the Host socket ConnectOnion-specific and accept one versioned ACP_NOTIFICATION envelope whose nested JSON-RPC notification is exactly session/cancel for schema-v1.19.0.
2. Add a small carrier capability to CONNECTED so a client chooses exactly one cancel vocabulary. Do not dual-send ACP cancel plus legacy INTERRUPT: a second queued interrupt could leak into a later turn.
3. Require an authenticated connection, exact schema/jsonrpc/method shape, and params.sessionId equal to the connection-owned session.
4. Translate a valid notification to the existing internal INTERRUPT lifecycle. Do not create a second cancellation engine or rewrite canonical traces.
5. Deliver at most one interrupt to one active WebSocketIO instance. Duplicate and completion-boundary cancels are idempotent and cannot enter the next turn.
6. Keep legacy INTERRUPT support for released clients. Unknown/malformed ACP frames fail closed with an explicit error.
7. If an ACP permission request is pending, the React client responds with RequestPermissionOutcome cancelled; the existing permission adapter becomes reject_hard and stops the turn, so it must not also enqueue a second cancel signal.

Preferred capability shape: CONNECTED.carrier_capabilities identifies the exact ACP schema and supported client notification methods. This is ConnectOnion rollout metadata, not ACP initialize negotiation.

## Acceptance criteria

- [ ] exact session/cancel for the connected session reaches DD-025 interruption within the existing bound
- [ ] mismatched session, schema, method, JSON-RPC shape, and unauthenticated frames execute no cancellation
- [ ] repeated cancel sends one internal interrupt for the active IO generation
- [ ] a cancel racing with OUTPUT cannot stop the next INPUT
- [ ] old clients using INTERRUPT keep working
- [ ] CONNECTED advertises capability only when the adapter is active
- [ ] unit and WebSocket E2E tests cover active LLM, active tool, pending gate, reconnect, duplicate, and completion races

## Rollout

Merge the Host reader/capability first, then publish the React writer, then let O Chat delegate its Stop action. Rollback removes the advertised capability and ACP adapter while leaving legacy INTERRUPT unchanged. No package release is part of this issue.

wu-changxing:
Closes #878.

## Summary

- accept an exact schema-v1.19.0 session/cancel notification inside the existing Host carrier
- advertise that client-notification capability in CONNECTED so React sends exactly one protocol form
- map ACP and legacy cancellation to one one-shot WebSocketIO interrupt generation
- reject wrong-session, malformed, unsupported, and idle cancellation without changing canonical traces
- record the lifecycle, compatibility, security, rollout, and rejected alternatives in DD-038

## Architecture

The Host WebSocket remains a ConnectOnion transport. The nested message is exact ACP; the outer carrier, authentication, session ownership, persistence, and DD-025 interrupt lifecycle remain ConnectOnion-owned. The frontend path is Host -> @connectonion/react -> O Chat. The retired standalone TypeScript SDK is not involved.

## Compatibility and safety

The Host reader lands first. A future React writer uses session/cancel only after CONNECTED advertises it and otherwise sends legacy INTERRUPT. It must never dual-send. The IO-scoped one-shot guard survives reconnects and prevents a repeated or completion-boundary signal from cancelling the next turn. Signed-command mode continues to authenticate and replay-check the entire carrier before this adapter runs.

## Verification

- 346 Host, ASGI WebSocket, IO, ACP, signature, interrupt, runtime-input, and session-ownership tests passed
- focused post-review run: 73 passed
- ruff, compileall, and git diff checks passed
- package sdist and wheel build passed
- Linus-style simplicity review removed an unnecessary one-use capability helper and found no remaining blocker

Bandit is not installed in the local environment; the repository CodeQL/security workflow remains a required merge gate.

### DIFF
--- a/connectonion/core/acp_wire.py
+++ b/connectonion/core/acp_wire.py
@@ -15,6 +15,7 @@ from acp import text_block, tool_content
 from acp.schema import (
     AgentMessageChunk,
     AgentRequest,
+    CancelNotification,
     PermissionOption,
     RequestPermissionRequest,
     RequestPermissionResponse,
@@ -32,6 +33,7 @@ ACP_RESPONSE_FRAME_TYPE = "ACP_RESPONSE"
 ACP_SCHEMA_VERSION = "schema-v1.19.0"
 ACP_SESSION_UPDATE_METHOD = "session/update"
 ACP_PERMISSION_METHOD = "session/request_permission"
+ACP_CANCEL_METHOD = "session/cancel"
 
 ACP_PERMISSION_OPTIONS = (
     PermissionOption(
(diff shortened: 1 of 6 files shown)


wu-changxing:
Self-review complete after the final push.

- Found and fixed one completed-turn edge: cancellation now requires the registry to still expose the same IO generation as `running`.
- Focused Host/ACP/session regression set: 220 passed.
- Earlier broad Host/ASGI/IO/ACP/signature/ownership set: 346 passed.
- Local full unit run: 6,332 passed; the residual failures were isolated to an absent optional `mcp` test dependency (10) and the pre-existing docs-site version mismatch (1). Sandbox-sensitive socket/browser/SSH tests passed when rerun with their required OS access.
- GitHub matrix is green on Python 3.10–3.13, native Windows browser/E2E, lockfile audit, static analysis, and CodeQL.
- `ruff` and `git diff --check` pass.

No remaining review blockers; the change stays within the Host adapter boundary and does not involve the retired standalone TypeScript SDK.

PR #884, opened 2026-08-11
feat: add durable ACP Host session mode transactions

### LINKED ISSUE #883: Implement durable ACP session/set_mode on the Host carrier
## Context

Parent RFC: #838. Governing decisions: DD-031 and DD-039.

The local stdio ACP adapter already implements exact session/set_mode semantics. The authenticated network Host does not: it forwards legacy mode_change only to a currently running Agent mailbox. An idle request has no receiver, a busy request can cross a model/approval/tool boundary, and the client receives no owned acknowledgement.

The active frontend path is Host -> @connectonion/react -> O Chat. The standalone TypeScript SDK is retired.

## Best-practice decision

This is a real policy transaction, not a wire rename.

1. Add an exact ACP v1.19 JSON-RPC request inside the versioned Host carrier:
   - outer type ACP_REQUEST and schema-v1.19.0
   - method session/set_mode
   - params are the official SetSessionModeRequest with sessionId and modeId
   - correlate one exact ACP_RESPONSE success or JSON-RPC error by request ID and session ID.
2. Advertise session/set_mode in CONNECTED only when this Host implements the full transaction. Include an exact SessionModeState so React learns currentModeId and availableModes instead of inventing them.
3. Authenticate first, bind to conn.session_id, and enforce the stored/active session owner. A caller-supplied session ID or _meta field grants nothing.
4. Linearize INPUT and mode changes across sockets and worker processes. The latest durable running record or active running registry state returns a retryable Session is busy error. An idle update copies the last durable snapshot, validates it, and appends the replacement under the storage file lock before any success response.
5. Persist a server-owned default snapshot for a new session so changing mode before the first prompt actually affects that prompt. Preserve requester ownership, TTL, history, trace, permissions, and unrelated fields.
6. Use the existing persisted IDs directly: safe, accept_edits, ulw. plan is a product/UI alias and is invalid ACP output/input.
7. Derive available modes from server authority:
   - safe is always available;
   - accept_edits requires the authenticated Host operator/admin, matching the existing approval hook;
   - ulw requires operator/admin plus an explicit server-side launch/configured turn ceiling. A client cannot choose or extend that ceiling through ACP because SetSessionModeRequest has no turns field.
8. On downgrade from ulw, remove skip_tool_approval, ulw_turns, and ulw_turns_used. On an authorized upgrade, initialize the bounded fields from the server ceiling, never client data.
9. Fail closed on unknown modes, malformed JSON-RPC, stale/duplicate IDs, wrong sessions, non-owners, storage failure, and lost ownership. Do not mutate conn.session or emit current_mode_update until persistence commits.
10. Keep legacy mode_change as a bounded compatibility path only; it must pass the same authority/mode validation and may not bypass the new ULW ceiling.

## Error ownership

Return an ACP_RESPONSE error, not an unrelated Host ERROR, for a syntactically owned request:

- -32602 for unsupported/malformed mode parameters
- -32002 for missing/wrong/unowned sessions
- -32000 for a busy session, with retryable metadata in the Host carrier
- -32603 for persistence/commit failure

Unknown schema/method remains unsupported and must not be treated as a mode request.

## Acceptance criteria

- [ ] Add a design decision extending DD-031 to the multi-socket/multi-worker Host transaction.
- [ ] Exact official SetSessionModeRequest/Response and SessionModeState serialization tests.
- [ ] CONNECTED capability/state advertisement is identity- and launch-authority-bounded.
- [ ] Mode change before first prompt persists and drives that prompt.
- [ ] Idle update survives reconnect/resume; failed writes expose no in-memory grant.
- [ ] Busy prompt, concurrent sockets, duplicate/stale request IDs, wrong owner/session, and worker-level storage races are covered.
- [ ] Safe/Auto/ULW transitions clean or initialize every bypass field correctly.
- [ ] Legacy mode_change cannot bypass the same authority checks.
- [ ] Python compatibility client consumes the acknowledgement/state without duplicate UI events.
- [ ] Security audit, dependency audit, full CI, and expert review are green before merge.

## Rollout

Land the React response/capability reader before enabling the Host writer. Then migrate O Chat to the acknowledged React operation. No package or application release is part of this issue.

wu-changxing:
## Summary

- implement exact ACP session/set_mode request and response handling on Host
- persist identity-bounded Safe, Auto, and ULW mode state through one cross-worker atomic transaction
- make prompt claims, CONNECT initialization, timeout behavior, and Agent failures fail closed
- expose the acknowledged operation in the Python client
- keep the browser boundary at @connectonion/react; O Chat remains protocol-free and the standalone TypeScript SDK remains retired

## Security and correctness

- session IDs remain correlation IDs; signature-verified requester identity owns the session
- durable JSONL state, not the process-local registry, is authoritative across workers
- no ACK is emitted before the durable append succeeds
- unexpected Agent and storage failures are logged server-side and sanitized on the wire
- hosted ULW cannot exceed the launch-time ceiling or be extended through the legacy mailbox
- client timeout means outcome unknown and requires reconnecting for authoritative state

## Validation

- 225 related unit tests passed
- changed production scope Ruff E/F/I checks passed
- Python compileall passed
- uv lock check passed
- package sdist and wheel build passed
- pip-audit against frozen runtime dependencies: no known vulnerabilities
- diff check clean
- no TypeScript or JavaScript files changed

Closes #883

### DIFF
--- a/connectonion/__init__.py
+++ b/connectonion/__init__.py
@@ -89,7 +89,8 @@ _FROM = {
         "read_file", "edit", "multi_edit", "glob", "grep", "write",
     )},
     **{name: ".network" for name in (
-        "connect", "RemoteAgent", "Response", "ExecResult", "host", "create_app",
+        "connect", "RemoteAgent", "Response", "ExecResult", "ACPModeError",
+        "host", "create_app",
         "IO", "relay", "announce", "HTTPRequest", "HTTPResponse", "HTTPRoute",
         "HTTPRouter",
     )},
@@ -198,6 +199,7 @@ __all__ = [
     "RemoteAgent",
     "Response",
     "ExecResult",
+    "ACPModeError",
     "host",
     "create_app",
     "IO",
--- a/connectonion/core/acp_wire.py
+++ b/connectonion/core/acp_wire.py
@@ -20,7 +20,11 @@ from acp.schema import (
     PermissionOption,
     RequestPermissionRequest,
     RequestPermissionResponse,
+    SessionMode,
+    SessionModeState,
     SessionNotification,
+    SetSessionModeRequest,
+    SetSessionModeResponse,
     ToolCallProgress,
     ToolCallStart,
     ToolCallUpdate,
@@ -35,7 +39,26 @@ ACP_SCHEMA_VERSION = "schema-v1.19.0"
 ACP_SESSION_UPDATE_METHOD = "session/update"
 ACP_PERMISSION_METHOD = "session/request_permission"
 ACP_CANCEL_METHOD = "session/cancel"
+ACP_SET_SESSION_MODE_METHOD = "session/set_mode"
 ACP_SESSION_MODE_IDS = frozenset({"safe", "accept_edits", "ulw"})
+
+ACP_SESSION_MODES = {
+    "safe": SessionMode(
+        id="safe",
+        name="Safe",
+        description="Ask before side effects.",
+    ),
+    "accept_edits": SessionMode(
+        id="accept_edits",
+        name="Auto",
+        description="Apply edits without asking; other tools still require approval.",
+    ),
+    "ulw": SessionMode(
+        id="ulw",
+        name="ULW",
+        description="Run without tool approvals within the Host launch ceiling.",
+    ),
+}
 
 ACP_PERMISSION_OPTIONS = (
     PermissionOption(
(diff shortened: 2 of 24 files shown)


wu-changxing:
Final review: no blocking findings. Two independent code reviews and the final self-review covered identity/ownership, cross-worker atomicity, ACK ordering, ULW ceiling enforcement, timeout ambiguity, failure recovery, and exception disclosure. The first CI run exposed one internal diagnostic compatibility regression; commit 4bed9abb preserves the original exception in-process while sanitizing only at the WebSocket boundary. The complete second CI matrix is green. Frontend ownership remains @connectonion/react -> O Chat; no TypeScript/JavaScript package files changed.

Issue #902, opened 2026-08-12
Stream Claude Code activity from co ai to O Chat


wu-changxing:
## User outcome

A user asks `co ai` to delegate one coding task to Claude Code. O Chat immediately shows the child tool lifecycle while the outer `claude_code(...)` invocation remains one normal ConnectOnion tool call.

## Architecture decision

This first product slice uses Claude Code documented `stream-json` CLI contract and translates child `tool_use` / `tool_result` blocks to existing native AgentIO `tool_call` / `tool_result` events.

The Python Agent SDK is not installed in-process: its current MCP 1.x dependency conflicts with ConnectOnion 1.7 MCP 2.x boundary. A future compatible SDK or isolated sidecar may add interactive child approvals without weakening the framework dependency decision.

The active browser path is Host -> `@connectonion/react` -> O Chat. React owns protocol decoding and typed session state; O Chat renders it. The standalone TypeScript SDK is retired and receives no work for this feature. Public ACP ingress remains the separate #900 / #901 compatibility track and is not a gate for this native product slice.

## Safety and ownership

- Permission mode and workspace root are operator-owned and absent from the model-visible schema.
- Claude safe mode disables ordinary `CLAUDE.md`, skills, plugins, hooks, MCP servers, commands, and custom agents for delegated runs.
- The child inherits only a small process/locale environment plus Claude authentication variables.
- Model-selected working directories must resolve inside the bound workspace; symlink escapes fail closed.
- Provider fields, stream lines, event counts, active correlations, output, usage, and request values are bounded.
- Resume accepts only the exact canonical UUID returned by an earlier invocation.
- Child intermediate text, thoughts, and plans do not become parent transcript or top-level O Chat state, consistent with DD-041 and DD-042.

## Acceptance criteria

- [x] One outer `claude_code` call streams inner tool starts and results as native tool events.
- [x] Stable provider tool IDs correlate starts and results without duplicate or permanently open cards.
- [x] Parent cancellation/timeout terminates the child process group and rejects late events.
- [x] A returned canonical session UUID resumes from the same bound workspace.
- [x] Final result, usage, cost, status, and session ID preserve the stable JSON envelope.
- [x] Missing CLI, authentication, provider, timeout, and malformed stream failures stay structured.
- [x] Provider-controlled payloads and collections are bounded; common secret-shaped arguments are redacted.
- [x] Unit tests use fake process transports and make no paid calls.
[...]

PR #901, opened 2026-08-12
Add acp_agent: one ACP client tool that drives Claude Code, Codex, and Gemini CLI

wu-changxing:
Advances #900. Implements the later downward-adapter edge of the #899 hourglass.

> **Priority:** DD-043 / #902 / #904 remains the first product slice: native Claude Code live activity for `co ai` and O Chat. This generic ACP adapter is a separately reviewed later-alpha track and is not a release gate for that work.

## What this is

One generic ACP **client** tool. ConnectOnion is the ACP Client and the selected coding engine is the ACP Agent.

```python
acp_agent(prompt, engine="claude-code")
acp_agent(prompt, engine="codex")
acp_agent(prompt, engine="gemini")
```

Claude Code and Codex use exact-version ACP adapters. Gemini uses exact `@google/gemini-cli@0.55.1 --acp`. Existing engine-specific `claude_code` and `codex` tools remain the preferred routes.

`co ai` registers a thin policy wrapper around this client. The model can select an engine, prompt, working directory, and an exact session for engines that passed cross-process conformance; command argv, approval policy, and workspace root remain operator-owned.

Companion documentation Draft: wu-changxing/connectonion-docs-site#58.

## Accepted boundary

DD-052 documents the downward edge:

- the official typed `agent-client-protocol` client owns framing and subprocess lifecycle;
- a model may select only a named engine; custom argv, approval policy, and workspace root are operator-owned;
- the final child message and bounded tool lifecycle IDs/titles cross the edge;
- child thoughts and plans do **not** become the outer session persisted thought or canonical TodoList plan;
- raw child tool inputs/outputs are not copied into ordinary progress events;
- final result is capped at 64 KiB, errors at 4 KiB, event text at 512 bytes, with bounded event/chunk/tool counts;
- exact Claude/Codex resume fails closed and never silently creates a fresh session;
- Gemini 0.55.1 is one-turn: real testing found that its advertised session cannot be loaded by the next child process, so named Gemini returns no session ID and rejects `session_id` before launch.

## Permission and process security

- Claude adapter: `@agentclientprotocol/claude-agent-acp@0.66.0`; persistent interactive CLI setting sources are disabled and the required ACP mode is reapplied on new/load.
- Claude receives only explicitly configured `CLAUDE_CONFIG_DIR` and `ANTHROPIC_API_KEY`; unrelated ambient credentials are excluded from the trimmed child environment.
[...]

Issue #900, opened 2026-08-12
acp_agent: one generic ACP client tool that drives Claude Code, Codex, Gemini CLI, and any ACP agent


wu-changxing:
Implements the **downward adapter** half of #899: one typed ACP client that lets `co ai` delegate a bounded turn to a reviewed coding agent without making that child the owner of the outer ConnectOnion session.

Merged implementation: #901 at `38014726`, merge commit `4105e41d`  
Release documentation: wu-changxing/connectonion-docs-site#58 at `4c4d910`, integrated into gated Alpha.2 docs PR #49 at `9c1a854`  
Architecture decision: DD-052

## Current engine contract

| Engine | Exact named route | Approval contract | Cross-process resume |
|---|---|---|---|
| Claude Code | `@agentclientprotocol/claude-agent-acp@0.66.0` | `manual`, `auto`, `deny` through advertised modes | yes |
| Codex | `@agentclientprotocol/codex-acp@1.1.14` | explicit operator-selected `auto` only | yes |
| Gemini CLI | `@google/gemini-cli@0.55.1 --acp` | `manual`, `auto`, `deny` through advertised modes; API-key, Vertex, or enterprise Code Assist authentication only | no at this pin |

The existing native `claude_code` and `codex` tools remain the preferred engine-specific routes. This tool adds a generic ACP edge; it does not replace them.

Google stopped serving Gemini CLI requests for individual/free/Pro/Ultra accounts on June 18, 2026. A legacy `~/.gemini/oauth_creds.json` file is therefore not a readiness signal. The exact route remains for explicit API-key, Vertex, and enterprise Code Assist authentication. Antigravity is not substituted until it has a documented ACP entry point and passes the same supply-chain, permission, environment, and real-provider review.

## Public model-facing API

```python
acp_agent(
    prompt: str,
    engine: str = "",       # claude-code | codex | gemini
    session_id: str = "",   # only for an engine with supports_resume=true
    cwd: str = "",          # operator workspace or a resolved descendant
    timeout: int = 600,
) -> str
```

A model may select only a reviewed engine, prompt, bounded working directory, and supported exact session. Process argv, approval policy, and workspace root belong to an operator-created `ACPAgent`; they are intentionally absent from the public tool schema.

## Native event boundary

Only the bounded final `agent_message_chunk` text and bounded tool lifecycle metadata cross the downward edge:

| ACP child signal | ConnectOnion view |
|---|---|
[...]

PR #904, opened 2026-08-12
feat: stream Claude Code tool activity to co ai

wu-changxing:
## Summary

- stream installed Claude Code CLI activity through native AgentIO tool cards
- preserve the existing single outer call, stable result envelope, canonical session resume, timeout, and cancellation behavior
- enforce operator-owned permission and workspace boundaries without exposing them in the model tool schema
- run the child in Claude `--safe-mode` with a minimal credential environment
- bound provider-controlled IDs, fields, collections, stream lines, event counts, active correlations, final output, and usage

## Product boundary

This is the native product path: Host -> `@connectonion/react` -> O Chat. React owns protocol decoding and typed state; O Chat renders it. The retired standalone TypeScript SDK is not changed. Public ACP ingress is separate work in #900 / #901 and is not required by this PR.

Child text, thoughts, and plans are not forwarded into the parent transcript or top-level O Chat plan. The parent agent remains the author of the final answer and owner of review.

## Security and permission boundary

- Claude safe mode disables ordinary `CLAUDE.md`, skills, plugins, hooks, MCP servers, commands, custom agents, and related customizations.
- Admin-managed policy and installed Claude authentication still apply.
- Unrelated parent provider, cloud, and GitHub credentials are not inherited.
- A requested `cwd` must resolve inside the operator-bound workspace; symlink escapes fail closed.
- Resume accepts only the exact canonical UUID returned by an earlier run, not CLI fuzzy search.
- Event capacity reserves a matching result for every emitted start so the UI cannot retain an orphaned in-progress card.

This adds observability, not an interactive child approval bridge. An unmatched permission request cannot round-trip through O Chat in this version and fails closed.

## Why CLI stream

`claude-agent-sdk==0.2.136` requires Python `mcp>=1.23,<2`, while ConnectOnion 1.7 requires `mcp>=2,<3`. The documented CLI stream provides the needed tool lifecycle without weakening that dependency boundary.

## Verification

- 138 related delegation, co ai assembly, import-surface, and tool-registry tests pass
- Ruff and `git diff --check` pass
- real installed-provider smoke: safe-mode completion succeeds
- real installed-provider smoke: native `tool_call` and `tool_result` are emitted
- real installed-provider smoke: unmatched Read permission fails closed and the canonical UUID resumes successfully
- one repository-existing `FileTools.permission` annotation warning remains unrelated

## Scope

- no public ACP ingress
- no interactive child approval bridge
- no TypeScript SDK changes
- no version bump, release, or package publication

Refs #902

### DIFF
--- a/connectonion/cli/co_ai/agent.py
+++ b/connectonion/cli/co_ai/agent.py
@@ -183,6 +183,7 @@ def create_agent(
         max_iterations=max_iterations,
         co_dir=co_dir,
     )
+    agent._delegation_workspace = Path.cwd().resolve()
     # This browser helper blocks on stdin, which is wrong for co ai's websocket
     # chat runtime. Use browser tools plus frontend-mediated user handoffs.
     agent.tools.remove("wait_for_manual_login")
--- a/connectonion/cli/co_ai/prompts/tools/claude_code.md
+++ b/connectonion/cli/co_ai/prompts/tools/claude_code.md
@@ -15,7 +15,7 @@ separate coding agent is better suited to implement or investigate it.
   exit is a result to handle; never claim the delegated work completed.
 - Claude Code runs headlessly here. Its inner tool starts and results appear
   automatically as live cards in co ai, so do not narrate or duplicate them.
-  Safe Mode can read and use actions already allowed by Claude's settings, but
+  Safe Mode can use actions allowed by its bound mode, but
   it cannot open an unmatched Claude permission prompt in the co ai UI. Accept
   Edits can edit in-scope files; other actions that need a prompt fail closed.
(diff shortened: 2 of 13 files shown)


wu-changxing:
CI follow-up at `a2021744`: all four Linux/Python jobs failed on the same shared interrupt regression because the test-created Agent did not bind the operator workspace that production `co ai` binds. The new fail-closed workspace check correctly prevented provider launch, so the test interrupt thread never observed its start marker.

The fixture now sets `_delegation_workspace = tmp_path`; production security behavior is unchanged. Verification: both Claude/Codex parameter branches passed five consecutive runs (10 executions), the 65 focused delegation tests pass, and Ruff/diff checks pass. Remote CI is rerunning.

Issue #903, opened 2026-08-12
[RFC] Align Connect controls with Codex modes and permission profiles


wu-changxing:
## Problem

Connect previously mixed workflow intent and approval authority in one mode vocabulary (`safe`, `plan`, `ulw`, and newer aliases). That made the Connect button harder to understand and diverged from Codex.

Codex models these as two independent controls:

1. **Collaboration mode** — how the agent works with the user
   - `default` → **Default**
   - `plan` → **Plan**
2. **Permission profile** — what the agent may do without asking
   - `:read-only` → **Read only**
   - `:workspace` → **Auto**
   - `:danger-full-access` → **Full access**

`Auto` is the product label for the workspace permission profile. `--yolo` remains a familiar CLI/compatibility alias for Full access, but it is not the primary UI label or canonical code identifier.

Official references:

- https://learn.chatgpt.com/codex/agent-approvals-security
- https://learn.chatgpt.com/codex/permissions
- https://learn.chatgpt.com/codex/config-file/config-reference

## Implemented change

- Split collaboration mode from permission profile in Host, React SDK, and O Chat.
- Use the Codex identifiers above on the wire and in public APIs.
- Present two independent controls in Connect: **Default / Plan** and **Read only / Auto / Full access**.
- Keep readers for persisted/wire values `safe`, `accept_edits`, `ulw`, `auto_approve`, and `full_access`, but immediately normalize them to the canonical Codex permission profiles.
- Keep Full access bounded by ConnectOnion's existing Host grant and turn-limit enforcement. This aligns vocabulary and contract shape; it does not claim that every provider has Codex's operating-system sandbox.
- Replace the old CUA vocabulary in current code, tests, comments, and documentation. Old names remain only in explicit migration/compatibility surfaces and the recognizable `--yolo` CLI shorthand.

## Mapping

| Legacy Connect value | Canonical value |
| --- | --- |
| `safe`, `default` | `:read-only` |
| `accept_edits`, `auto_approve` | `:workspace` |
| `ulw`, `full_access` | `:danger-full-access` |
| `plan` | collaboration `plan` + permission `:read-only` |

## Acceptance criteria

- [x] Connect shows independent collaboration and permission controls.
- [x] Public SDK APIs expose `collaborationMode` and `permissionProfile` separately.
- [x] ACP advertises and accepts the three canonical permission profile IDs.
- [x] Plan changes do not write Host permission authority.
- [x] Permission changes are Host-acknowledged before prompts can proceed.
[...]

PR #905, opened 2026-08-12
refactor: align permissions with Codex profiles

wu-changxing:
## Summary

- align Connect with Codex's two-axis model: collaboration is `default | plan`, while Host authority uses `:read-only | :workspace | :danger-full-access`
- advertise the three exact permission profiles as **Read only / Auto / Full access** through ACP
- make `permission_profile` the primary runtime/API vocabulary (`HostPermissionPolicy`, `set_permission_profile`, `available_permission_profiles`)
- translate the canonical profiles to Codex and Claude provider controls only at the delegation boundary
- update current docs and design decisions; keep `--yolo` only as the familiar CLI shorthand for Full access

## Compatibility and safety

- previous values (`safe`, `accept_edits`, `ulw`, `auto_approve`, `full_access`) are accepted only by explicit migration readers and immediately normalized
- malformed or unknown restored authority falls back to `:read-only`
- Full access requires a complete positive, unexpired Host turn grant; incomplete, exhausted, or third-party state loses bypass authority before model/tool execution
- Plan is collaboration intent, never Host permission authority
- this aligns vocabulary and contract shape; it does not claim that every provider has Codex's operating-system sandbox

## Checks

- 411 focused Host, ACP, approval, provider, resume, interrupt, and compatibility tests passed
- 117 approval/profile tests passed after the final public export rename
- `git diff --check`
- remote CI is running on `5ce3d45f`

## Delivery

This is the Host/protocol slice of #903. The browser rollout continues in:

- openonion/connectonion-react#34
- openonion/oo-chat#139

No package release is included in this PR.


wu-changxing:
## Coordination review — protocol contract is not frozen yet

CI is fully green and the reader/writer migration direction is sound, but this PR should remain Draft until #903 records one final public contract. The current PR uses `default / auto_approve / full_access`, while the active design discussion is separating collaboration mode (`default / plan`) from a permission profile (`read-only / workspace / danger-full-access`). React and O Chat must not be asked to implement a moving wire contract.

There is also one concrete state-cleanup blocker in the committed head. When a session leaves `full_access`, `connectonion/useful_plugins/tool_approval/approval.py` clears `skip_tool_approval` but can leave `full_access_turns`, `full_access_turns_used`, and `full_access_prompt` behind. Host validation rejects those fields outside full access, so a downgrade can preserve stale authority metadata or create a rejected session update.

Before marking Ready:

- update #903 with the final separation and exact wire values;
- update Core, React, and O Chat fixtures from that single contract;
- clear the complete full-access grant state on every downgrade, preferably through one shared cleanup path;
- add a regression covering `full access -> default/read-only` with no stale grant fields;
- document compatibility/rollback behavior for older React/O Chat clients;
- rerun the full matrix after the final contract lands.

Please also coordinate the design-decision numbering with #901: both branches currently introduce a DD-044.

wu-changxing:
Final local review on 1034069e: 446/446 focused Host, ACP, MCP, approval, provider-adapter, resume, interrupt, SDK, and stdio tests passed. The real MCP lifecycle file passed 26/26 after keeping startup bounded at 30 seconds for MCP 2.x cold imports. Final compatibility/static slice passed 92/92, Ruff passes for all touched files, and git diff --check is clean. I also restored deprecated tool_approval exports for one migration window; canonical writers remain permission-profile-only. Waiting for remote CI before merge.

PR #912, opened 2026-08-12
feat(cli): add co transfer for credit transfers

### LINKED ISSUE #65: CLI: Add transfer command for credit transfers
## Summary

Add a CLI command to transfer OpenOnion credits to another address.

## Command

```bash
co transfer <address> <amount> [--memo "message"]
```

## Examples

```bash
# Transfer $5 to an agent for onboard payment
co transfer 0x1234...abcd 5.00

# Transfer with memo
co transfer 0x1234...abcd 10.00 --memo "Payment for code review"

# Check balance first
co status
```

### 1. Add transfer command to CLI

```python
@app.command()
def transfer(
    address: str = typer.Argument(..., help="Recipient address"),
    amount: float = typer.Argument(..., help="Amount in USD"),
    memo: str = typer.Option(None, "--memo", "-m", help="Optional memo")
):
    """Transfer credits to another address."""
```

### 2. Call oo-api

- POST `/api/v1/transfers`
- Requires authentication (Ed25519 signature)
- Show balance before/after

### 3. Output

```
Transferring $5.00 to 0x1234...abcd

✓ Transfer complete
  From:    0xabcd...1234
  To:      0x1234...abcd  
  Amount:  $5.00
  Balance: $45.00 → $40.00
```

## Dependencies

- oo-api `/api/v1/transfers` endpoint (already implemented)
- Authentication via `address.load()` and signing

## Files to Modify

- `connectonion/cli/main.py` or new `commands/transfer_commands.py`

wu-changxing:
Closes #65.

`POST /api/v1/transfers` has existed in oo-api since the onboard-payment work, but nothing in the CLI reached it. Moving credits meant hand-rolling a curl with a bearer token — which is literally how a transfer was done earlier today.

## What this adds

```
co transfer <address> <amount> [--memo "..."]
co transfer --list [--type sent|received|all]
```

Output follows the shape in the issue:

```
Transferring $0.01 to 0xcb2f...8a2d

✓ Transfer complete
  From:    0x10e6...5508
  To:      0xcb2f...8a2d
  Amount:  $0.01
  Memo:    co transfer smoke test
  Balance: $329.02 → $329.01
```

## Guards before anything is posted

Money leaving the account is not reversible from the CLI, so three checks run first and each costs a round trip instead of a transfer:

- non-positive amount
- amount larger than the current balance (with the purchase URL)
- no API key — told to run `co auth`

## Two decisions worth review

**The balance read only decorates the output.** If `/auth/me` fails, the transfer still proceeds and the balance line is simply dropped. A decoration that fails must not fail the thing it decorates — the alternative is refusing to move money because we could not render a nicety.

**A network error on the POST does not claim nothing happened.** The request may have reached the backend. Rather than implying the transfer did not occur, it points at `co transfer --list` so the caller checks before retrying — a retry after a landed-but-unacknowledged POST sends the money twice.

## Verification

Against the live backend, not mocks:

- a real `$0.01` transfer through the command moved the money and reported `$329.02 → $329.01`
- `co transfer --list --type sent` showed both that transfer and the earlier `$5.00` one
- exit codes: bare invocation `1`, zero amount `1`, `--list` `0`

10 unit tests cover the guards, the posted body, the output, and the two failure decisions above.

**The tests were shown to fail first**: removing the amount and balance guards turns 3 of them red, and restoring the guards turns them green again. A test that has never failed proves nothing about what it guards.

### DIFF
--- a/connectonion/cli/commands/transfer_commands.py
+++ b/connectonion/cli/commands/transfer_commands.py
@@ -1,0 +1,158 @@
+"""
+Purpose: Transfer OpenOnion credits from this account to another address
+LLM-Note:
+  Dependencies: imports from [requests, rich.console, backend.backend_url, project_cmd_lib.load_api_key] | imported by [cli/main.py via handle_transfer()] | calls the configured backend /api/v1/transfers and /api/v1/auth/me | tested by [tests/e2e/cli/test_cli_transfer.py]
+  Data flow: receives address/amount/memo → load_api_key() resolves the account token → GET /api/v1/auth/me for the starting balance → POST /api/v1/transfers → GET /api/v1/auth/me again → prints from/to/amount and the balance before → after
+  State/Effects: moves credits between two accounts on the backend — not reversible from the CLI | no local file writes | two or three network calls
+  Integration: exposes handle_transfer(address, amount, memo) for CLI | reuses the same token resolution as `co status` so the account charged is the account shown there
+  Performance: 2-3 backend round trips, ~1-3s total
+  Errors: refuses a non-positive amount and a self-transfer before spending anything | surfaces the backend's own message for insufficient balance or an unknown recipient | a failed balance read never blocks the transfer, it only drops that line from the output
+"""
+
+from typing import Optional
+
+import requests
+from rich.console import Console
+
+from ...backend import backend_url
+from .project_cmd_lib import load_api_key
+
+console = Console()
+
+TIMEOUT = 30
+
+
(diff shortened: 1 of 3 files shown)


wu-changxing:
Superseded by #1023, which exposes the same live /api/v1/transfers API but is built against the current backend contract (field names verified line-by-line), follows the cli-skill-design methodology (#1008: confirm-before-irreversible, non-zero exits, tips that survive piping), and ships with live production verification. Thanks for the earlier groundwork — the command shape here informed the final one.

PR #920, opened 2026-08-12
feat(acp): preserve browser attachments

### LINKED ISSUE #919: Preserve browser image and file attachments over native ACP
## Problem

The supported browser path is `ConnectOnion Host -> @connectonion/react -> O Chat`. O Chat already accepts image and file attachments, but the native ACP adapter currently turns only `TextContentBlock` and `ResourceContentBlock` into prompt text. It rejects `ImageContentBlock` and `EmbeddedResourceContentBlock`, and it never calls the existing `Agent.input(..., images=..., files=...)` attachment path.

Selecting native ACP without closing this gap would either reject a previously supported input or tempt a client to silently drop content. Neither is acceptable for the atomic React migration in #893 / openonion/connectonion-react#32.

## Design constraints from existing decisions

- DD-045: native ACP remains behind authenticated direct admission; failures after native selection never downgrade to `/ws`.
- DD-047: the browser has one virtual workspace root `/`. Attachment metadata must not select or reveal a Host path.
- DD-033: browser content must not gain stdio MCP process-launch authority.
- ACP `_meta` is descriptive only and never an authority or filesystem channel.
- Existing Host file/image count and decoded-size ceilings remain authoritative. The native gateway must not bypass them.

## Proposed mapping

1. Advertise ACP `promptCapabilities.image=true` and `embeddedContext=true` only when the implementation below is active.
2. Map an exact `ImageContentBlock` to the existing Agent image data-URL input after strict MIME, base64, count, and decoded-size validation.
3. Map an exact embedded text/blob resource to the existing Agent file input after strict validation. Use one documented opaque upload URI convention to preserve a safe basename; never interpret it as a Host path. `Path(...).name` remains the final server-side filename boundary.
4. Continue to reject audio until the Agent input contract supports it.
5. Do not fetch arbitrary resource links. A link can remain textual context, but it is not an uploaded file and must not trigger server-side URL access.
6. Pass the same configured Host attachment limits into each network ACP adapter, or factor the validation into a shared neutral helper. Reject before Agent construction/turn mutation and clean partial uploads on failure.
7. Preserve ordered ACP updates and the existing `files_received` normalized UI event without adding parsing to O Chat.

## Tests / acceptance

- official-model image, embedded text file, and embedded binary file reach `Agent.input` exactly once;
- mixed text/images/files preserve order-independent meaning and safe names;
- malformed/oversized base64, unsupported MIME/audio, count overflow, traversal names, Host-path-like URIs, and `_meta` path hints fail closed before turn mutation;
- no SSRF from resource links and no Host path appears in ACP errors/updates;
- cancellation/failure removes partial upload files and leaves the last ACP snapshot resumable;
- stdio ACP behavior remains compatible, and network sessions still send/accept only virtual `cwd: "/"`.

## Cross-repository rollout

Land this Host capability before React enables atomic native selection for attachment-bearing prompts. React then maps its existing typed image/file inputs to official ACP content blocks; O Chat continues to call only the React API.

wu-changxing:
## Summary

- advertise native ACP image and embedded-context prompt capabilities
- map official ACP image/text-resource/blob-resource blocks into the existing `Agent.input(images=..., files=...)` contract
- validate MIME types, strict base64, decoded size, combined count, and portable opaque upload names before a turn
- isolate network upload staging per authenticated principal and roll back failed/cancelled turn files
- bound retained uploads per principal with atomic cumulative byte/file quotas
- document the boundary as DD-048 and extend official SDK subprocess coverage

## Design

This follows DD-029, DD-033, DD-045, and the accepted DD-047:

- browser ACP keeps virtual `cwd: "/"`; attachment metadata never selects a Host path
- resource links are not fetched
- `_meta` is not a content or authority channel
- browser/native ACP does not gain stdio MCP launch authority
- successful files remain available to resumable private Agent context, bounded by operator-configurable 100 MiB / 100 file defaults per authenticated principal
- the cross-process quota lock covers admission and physical staging only, never model work, approval, cancellation, or commit
- the one-MiB network JSON-RPC frame limit stays unchanged; larger files require a future authenticated streaming design

Closes #919

## Validation

- `ruff check` on all changed Python files
- 245 ACP lifecycle/resume/mode/approval/MCP/stdio/official-SDK and Agent regression tests
- 115 focused ACP attachment, Agent upload, and turn-outcome tests after quota hardening
- byte quota, file-count quota, concurrent connection reservation, unexpected entry, and pre-model lock release regression tests
- `git diff --check`

## Review status

- initial independent review: one P1 cumulative disk-exhaustion finding
- P1 addressed by commit `97bc1f71`; latest revision requires re-review and fresh CI before merge


wu-changxing:
## Review findings addressed in `ddad51d8`

The latest independent re-review found two distinct P2 concurrency races; both are fixed before merge:

1. Failed/cancelled rollback could unlink an upload between another connection's `scandir()` and `entry.stat()`, causing a false internal-storage error. The quota scan now ignores only `FileNotFoundError` for an entry that disappeared concurrently; deletion can only reduce usage. Unexpected surviving non-regular entries still fail closed.
2. `session/cancel` received while a prompt waited for the principal quota lock could be lost because the Agent generation had not started. The quota-wait phase is now part of the prompt-active lifecycle, with an explicit cancellation intent checked after acquisition and again after generation creation. Cancel or close in that window returns `cancelled` without calling `Agent.input` or writing a file.

New deterministic regressions cover stale scan entries, cancel while waiting, and close while waiting. The expanded ACP/Agent/gateway/official-SDK matrix passed 314 tests; the final focused suite passed 56 tests, plus Ruff and `git diff --check`.

Final reviewer sign-off and fresh CI are still required; this PR is not yet being merged.

PR #926, opened 2026-08-13
[1.9 Deferred] phone: give the endpoint a caller — co phone and notify_owner

wu-changxing:
Draft. The framework half of `openonion/oo-api#136` — that PR built `POST /api/v1/phone/notify` and **nothing in the framework could reach it.** The backend half of a channel is not a channel.

## What this adds

```
co phone                        what number your agents reach you on
co phone number +61435525634    set it
co phone notify "..." [--urgent]
```

and `notify_owner(message, urgent)` as an agent tool — so an agent blocked at 2am can make someone look, instead of dying quietly with its question unasked (#843).

`--urgent` places a call that speaks the message; without it you get an SMS. That is the whole distinction, because it is the only one that matters: a text is seen when someone next looks, a ringing phone gets through a silent one.

## Where the credential lives, and why it differs from #866

**No carrier credential comes down here.** These call oo-api, which holds it. The Telegram tool in #866 does the opposite — the bot is the user's own from @BotFather and the token sits in their `keys.env`.

Both are right, and the rule that decides it is **whose account it is and who pays** — not what the channel looks like. Twilio is our account and our bill, so a deployed agent must never hold that key. A Telegram bot is the user's and costs nothing, exactly like the Google OAuth token `co gmail` uses.

Worth writing down because "chat and phone are both messaging, put them together" is a plausible-sounding mistake, and the next channel added will face the same fork.

## The refusals are passed through, not flattened

The backend refuses for three different reasons and every one has a different fix: no number configured, inside the cooldown, out of credit. They come back with the server's own wording **and the `Retry-After` delay still attached**.

An agent that cannot tell *"wait 240 seconds"* from *"you have no number"* will either loop or give up — the two behaviours the cooldown in `openonion/oo-api#136` exists to prevent. Collapsing them into "failed" would throw away the instruction the agent needs.

## Tests

9 unit tests, no live backend:

- a normal notification goes out as a text, with the exact payload and bearer token asserted
- **`--urgent` escalates to voice** — the case the whole feature exists for
- **a rate-limited notification keeps the retry delay** (429 + `Retry-After: 240` → "240" survives into the error the agent reads)
- no number configured says so, rather than failing vaguely
- **no API key points at `co auth login` and never calls the backend**
- `co phone` with nothing configured prints the command to set one and says agents can only reach you by email until then — a person should leave with the next command, not a blank
- a failed notify exits non-zero, so a script can tell

`431 passed` across these, the import-cost guard, and the full `tests/e2e/cli` suite.

## Not yet real

`openonion/oo-api#136` has never talked to a live carrier — no Twilio account, no number bought. Until that exists and `scripts/add_phone_columns.py` has run, `co phone notify` will fail against production. **The first real call is the actual test of both PRs.**

### DIFF
--- a/connectonion/cli/commands/phone_commands.py
+++ b/connectonion/cli/commands/phone_commands.py
@@ -1,0 +1,55 @@
+"""
+Purpose: `co phone` — set the number your agents reach you on, and send a notification to it
+LLM-Note:
+  Dependencies: imports from [sys, rich.console, useful_tools.phone] | imported by [cli/main.py via phone_*()] | tested by [tests/unit/test_phone.py]
+  Data flow: handle_phone_number(phone) → set/get_owner_phone() | handle_phone_notify(message, urgent) → notify_owner() → prints the outcome
+  State/Effects: one HTTP call per command via the tool | no local state
+  Integration: same shape as handle_email_send — a thin handler over the tool an agent already has
+  Errors: any failure prints the server's own reason and exits 1
+"""
+
+import sys
+
+from rich.console import Console
+
+from ...useful_tools.phone import get_owner_phone, notify_owner, set_owner_phone
+
+console = Console()
+
+
+def handle_phone_number(phone: str = None) -> int:
+    """Show the configured number, or set it."""
+    result = set_owner_phone(phone) if phone else get_owner_phone()
+
+    if not result["success"]:
+        console.print(f"[red]{result['error']}[/red]")
+        sys.exit(1)
+
+    if phone:
+        console.print(f"[green]Your agents will reach you on {result['phone']}[/green]")
+        return 0
+
+    if not result.get("configured"):
+        console.print(
+            "No number set. [dim]co phone number +61435525634[/dim]\n"
+            "Until then your agents can only reach you by email."
+        )
+        return 0
+
+    console.print(f"Your agents reach you on [cyan]{result['phone']}[/cyan]")
+    return 0
+
+
+def handle_phone_notify(message: str, urgent: bool = False) -> int:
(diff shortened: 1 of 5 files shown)


wu-changxing:
Review of Draft head `62db2ebe`: the focused suite passes (**9 passed**), but the advertised error contract is not implemented for transport failures.

`notify_owner()`, `set_owner_phone()`, and `get_owner_phone()` call `requests` without catching `RequestException`. I reproduced an offline `notify_owner("wake me")` raising `requests.exceptions.ConnectionError` with a traceback instead of returning `{"success": False, "error": ...}`. Successful and error JSON decoding is also unguarded, so a proxy/gateway returning malformed JSON has the same result.

Wrap transport and decode failures for all three operations, return the stable failure shape, and add timeout/connection/malformed-JSON tests through both the tool and CLI paths.

The cross-repo backend dependency, openonion/oo-api#136, is still an open Draft. Keep this Draft until that contract is merged/deployed and a real SMS/voice smoke test confirms the client response shape and billing path.

wu-changxing:
Closing this Draft implementation as deferred. The PR states the backend has never completed a live carrier call and production would currently fail until carrier setup and schema deployment exist. Phone/voice remains tracked by #936/#938 and is outside the stable boundary in #1154; open a fresh PR only after a real end-to-end carrier path exists.

wu-changxing:
Release target correction from Aaron: phone notification/owner contact is **deferred to 1.9**, not cancelled. Keep #936/#938 as the design and product requirements. This PR remains closed until the backend has a real carrier E2E and the 1.9 channel architecture is active.

Issue #846, opened 2026-08-11
project_root() walks past the project into shared parents — $HOME and /tmp both resolve as projects


wu-changxing:
`project_root()` (`connectonion/project.py:26-35`) walks up until it finds a `.co/`, with no upper bound. Its own docstring already names the failure mode:

> the walk stops at the *nearest* `.co/`, so a stray one shadows the project's own for everything that comes after it.
> Two of those bugs failed open: a project configured `trust: strict` ran as `careful`, and an address on the blocklist read back as a stranger.

The unbounded walk means the "stray `.co/`" case is not rare — it is the default in two very common places.

## Reproduced

```
/tmp                       -> /private/tmp
/Users/me                  -> /Users/me
/Users/me/Desktop          -> /Users/me
/                          -> /
```

Two things worth separating:

**1. `~/.co` is the global config directory, not a project.** So *every* directory under `$HOME` that is not itself inside a project resolves to `$HOME` as its project root. `co` run from `~/Desktop` believes it is in a project rooted at `$HOME`, and reads `$HOME/.co` as that project's config — trust lists included. Given the docstring's own history (`trust: strict` degrading to `careful`), this is the failure that matters.

**2. Any writable shared parent becomes a project for everything beneath it.** On this machine another process had left a `/private/tmp/.co`, so every temp directory resolved to `/private/tmp`. Any user on the box can create that.

## Where it shows up first: tests in a git worktree

The repo's guideline is to develop in worktrees. `tests/conftest.py:405` (`_no_stray_project_above_the_test`) asserts the resolved project is the repo or the test's own tmp dir — and in a worktree it is neither, because a worktree has no untracked `.co/` (the main checkout does), so the walk escapes to `/private/tmp`.

Same file, same commit, same machine:

| checkout | result |
|---|---|
| main checkout | `10 passed` |
| git worktree | `10 passed, 10 errors` |

Every error is that assertion. The guard is correct — it is reporting real contamination — but it makes the suite unusable in the workflow the project recommends, and the failure reads as "your change broke something" rather than "your checkout has no `.co/`".

Workaround while developing: `mkdir .co` in the worktree root.

## Suggested direction

Stop the walk at a repository boundary — a directory containing `.git` — and treat `$HOME` as a boundary too. A project's `.co/` is inside the project; a `.co/` found only by climbing out of the repository is somebody else's.

That fixes the test symptom and the trust-list failure with one rule, rather than teaching conftest about worktrees.

PR #929, opened 2026-08-13
release: 1.6.1 — the CLI stops acting as the wrong account

wu-changxing:
First patch on the **1.6 stable line**. `main` is the 1.7 preview train (ACP), so this is cut from `v1.6.0` and cherry-picks fixes only — no feature train, per VERSIONING.md.

## What ships (PR #845, verified end-to-end this session)

- **The CLI refuses an `OPENONION_API_KEY` that belongs to an account this machine does not hold the signing key for.** A stray `.env` in the working directory could shadow the real identity in `~/.co/keys.env`, so a command run from the wrong directory silently read *that other account's* mailbox and spent its credit — the root of the wrong-sender / wrong-mailbox class of reports. Now it re-authenticates, or fails closed and says `co auth`.
- **`co outlook read` keeps the URLs in HTML mail** instead of stripping `<a href>` down to its link text, which had made per-recipient links (with tokens) unrecoverable.
- **A malformed JWT claim no longer crashes every CLI command** (`{"public_key": []}` etc. hit `.get()` on a non-dict).

## Why 1.6.1 and not 1.7

VERSIONING.md: patch releases carry "bug, security, documentation, compatibility fixes; they do not carry the next feature train." The ACP subsystem on `main` is that train and matures in `1.7.0aN`. This is the stabilised, non-ACP safety fix, so it belongs on the 1.6 line.

## Release checklist

```
Version
  [x] number chosen; patch (fixes only), reason above
  [x] pyproject.toml, connectonion/_version.py, VERSIONING.md (current + history), uv.lock
  [ ] docs-site/lib/version.ts — separate repo on another branch; bump on its own main (CI skips it here; does not block PyPI)
Gates
  [x] cherry-pick applied clean onto v1.6.0; feature tests green (21 passed)
  [x] release PR opened for the CI gate (8 jobs incl. Windows + e2e)
  [ ] CI green → merge
  [ ] build FROM THE MERGED COMMIT, twine check, wheel inspected, clean-venv `co --version` + `co outlook read --help`
Publish
  [ ] tag v1.6.1 (IRREVERSIBLE beyond here)
  [ ] twine upload, reinstall from PyPI, version confirmed
  [ ] GitHub release with artifacts
After
  [ ] Discord announce (feature framing)
```

Cut as the first of several small patches — subsequent fixes (project-root boundary #846/#852, the rest of the credential-guard cluster, gmail HTML #790, attachment boundaries) ship as 1.6.2 / 1.6.3, 1–2 verified per release.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,9 +2,9 @@ name: Tests
 
 on:
   push:
-    branches: [ main, master ]
+    branches: [ main, master, 'release/**' ]
   pull_request:
-    branches: [ main, master ]
+    branches: [ main, master, 'release/**' ]
   workflow_call:
 
 jobs:
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -38,9 +38,10 @@ patch releases; the minor is the statement that it no longer needs to be.
 - Reserved for breaking changes, or for a stable release worth naming
 - Same rule as any whole number: earned, not reached
 
-## Current Version: 1.6.0
+## Current Version: 1.6.1
 
 ### Version History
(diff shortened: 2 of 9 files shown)


wu-changxing:
Closing — wrong direction. A stable-line patch (release/1.6.1, cut from v1.6.0) sits *behind* the 1.7 `main`, so it can't PR into main meaningfully (main already carries #845). The branch is the 1.6 stable line itself; it's tagged and published directly. CI now runs on it via the push trigger on release/** (commit 8f1ab132). Milestone 1.6.1 stays.

PR #935, opened 2026-08-13
run the NL agent in the caller, not inside the daemon

### LINKED ISSUE #933: co browser do holds the single execution lane for its entire agent loop; the NL agent should be a client, not resident in the daemon
## The bug in one line

`co browser do` runs its **entire agent loop inside a single daemon request**, so it holds the one execution lane for the whole run instead of queueing one browser action at a time like every other caller.

```python
# cli/browser_agent/daemon.py
def _run_nl(self, command: str) -> tuple:
    """Hand the command to the NL agent driving the same live browser."""
    ...
    result = build_browser_agent(self.browser, api_key).input(command)   # whole loop, one request
    return True, _stringify(result)
```

`_run_nl` is called from `dispatch()`, and `serve()` is strictly serial:

```
accept → _read_request → dispatch → _send_reply → close
```

So `.input(command)` — up to 200 steps, each an LLM call plus a browser action — completes before `serve()` returns to `accept()`. No other client is served for the duration.

## Why that is the wrong shape

The NL agent is *resident inside* the daemon. It should be a *client of* the daemon:

| Expected | Actual |
| --- | --- |
| agent thinks → sends one command → waits its turn → runs a few seconds → releases → thinks again | agent runs entirely inside one request, holding the lane for all 200 steps |
| other sessions interleave between its steps | nothing interleaves until the whole run finishes |

The decisive detail: **most of each step is spent waiting on the model, which does not need the browser at all.** During the wait the browser sits idle and every other caller is still blocked. Measured on one afternoon, with the Gemini project over its spending cap so each call retried:

```
extract_data(selector="button")   ✓ 245.1s     # one command, one lane, browser idle
extract_data(selector="div")      ✓  61.0s
429 × 54 · "Your project has exceeded its monthly spending cap"
```

Three sessions were starved for ~40 minutes. One of them logged `client vanished before reply` after giving up.

## Suggested fix

Move the NL agent to the **client side**: `co browser do` runs the agent loop in its own process and issues each tool call as an ordinary `co browser <verb>` request.

- Each browser action becomes a normal few-second queue slot, so `do` interleaves with other sessions by construction.
- Model latency happens in the client process and holds nothing.
- **The daemon's concurrency model does not change at all** — no threads, no async rewrite. `_run_nl` goes away; the daemon keeps serving one short request at a time.
- Tab claims keep working unchanged, since each action still arrives with the same `caller`/`tab` envelope.

Trade-off to design for: the run is no longer atomic. Another caller can now act on the *same* tab between the agent's steps — which is what the existing `GUARD_WINDOW` claim mechanism is for, and it becomes load-bearing rather than decorative. A `do` that must not be interleaved should hold its own tab, exactly as the docs already tell human callers to do.

Related: #931 (serial daemon, tabs are not lanes), #932 (why one browser exists at all). This issue is the cheapest of the three and removes the worst offender without touching the daemon's threading model.

## Environment

macOS, `connectonion` at `f9e2f40`, evidence in `~/.co/browser.log`.

wu-changxing:
Closes #933.

## The problem

`co browser do` ran its **entire agent loop inside one daemon request**:

```python
def _run_nl(self, command: str) -> tuple:
    result = build_browser_agent(self.browser, api_key).input(command)   # whole loop
```

`serve()` is strictly serial (`accept → dispatch → reply → close`), so up to 200 steps — each an LLM call — completed before any other client was served. Three sessions were starved for ~40 minutes by one run, and the browser being held sat idle for nearly all of it, because most of each step is spent waiting on the model, which needs no browser.

## The change

The loop moves to the calling process. Each tool call becomes an ordinary `co browser <verb>` request carrying the same `caller`/`tab` envelope.

- Each browser action is a short queue slot, so other sessions interleave between the agent's steps.
- Model latency happens in the client and holds nothing.
- **The daemon's concurrency model is untouched** — no threads, no async, still one short request at a time.

New `cli/browser_agent/nl_client.py` builds the agent with one generated proxy per `BrowserAutomation` verb; each proxy sends a request and returns the reply as the tool result. The proxies carry the real verbs' signatures and docstrings, so the schema the model sees is unchanged.

## The trade-off, stated plainly

**A `do` run is no longer atomic.** Another caller can act on the same page between the agent's steps.

That is what the `GUARD_WINDOW` tab claim is for, and this change makes it **load-bearing rather than advisory**: a `do` that must not be interleaved has to hold its own tab (`-t <name>` on the `do` itself). Both `useful_skills/co-browser/SKILL.md` and `docs/useful_skills/co-browser.md` now say this outright, since the previous wording implied a `do` was indivisible.

## A problem that dissolves

`do` used to spend whatever key the *daemon's* environment held — the directory whoever started the browser happened to be in — which is why dispatch had to refuse callers whose account did not match (#728). Running in the caller's process, the caller pays. The guard stays for old clients; the class of bug it worked around is gone.

The daemon still accepts `do` and answers with an upgrade notice, because only a client older than the daemon still sends it over the wire.

## Verification

A test that **fails before the change and passes after** — it drives the real `co browser do` entry point with a slow stubbed model and asserts a second client's ordinary verb is served *while the agent is still thinking*, which is the moment that used to be stolen.

Before (with `nl_client.py` removed and sources reverted, so the loop really is in the daemon):

```
E   AssertionError: a second client was not served while a `do` was mid-run --
    the agent loop is still holding the single execution lane for its whole run
1 failed
```

After:

```
1 passed
```

Full suite: **7080 passed, 13 skipped, 2 failed**. Both failures are pre-existing on `main`, confirmed by running them there — a changelog-entry check and a CJK stealth test that needs a real browser.

`test_dispatch_do_routes_to_nl` asserted the old behaviour directly; it is rewritten to assert the new contract rather than deleted.

## Note for reviewers

Landing this needs a daemon restart to take effect, and a restart interrupts every session currently using the browser. Worth announcing before it goes out.

### DIFF
--- a/connectonion/cli/browser_agent/client.py
+++ b/connectonion/cli/browser_agent/client.py
@@ -181,6 +181,38 @@ def _ensure_browser_ready(line: str) -> bool:
     return True
 
 
+def run_verb(line: str, headless: bool = False, tab: str = None) -> str:
+    """Send one verb and return the daemon's reply text instead of printing it.
+
+    This is what makes the NL agent a *client* of the daemon rather than a
+    resident of it (#933). The agent needs each tool result back as a string to
+    feed its next turn; `send()` prints and returns an exit code, which is right
+    for a shell caller and useless to a caller that has to read the answer.
+
+    One request per tool call is the whole point: each is a short queue slot, so
+    other sessions interleave between the agent's steps instead of waiting out
+    its entire run.
+    """
+    # send() is the one place that knows the envelope, the transport, the
+    # cold-start provisioning and the old-daemon diagnostics. Reimplementing
+    # that here would mean two wire clients, and the one that drifts is the one
+    # nobody tests. So call it and capture what it writes.
+    #
+    # redirect_std*, not a patched builtins.print: patching the builtin would
+    # also swallow output from anything else running in this process, and the
+    # agent's own UI stream is running in it.
+    import contextlib
+    import io
+
+    out, err = io.StringIO(), io.StringIO()
+    with contextlib.redirect_stdout(out), contextlib.redirect_stderr(err):
+        exit_code = send(line, headless=headless, tab=tab)
+    body = out.getvalue().strip() or err.getvalue().strip()
+    if exit_code != 0 and not body:
+        body = f"command failed with exit code {exit_code}"
+    return body
+
+
 def send(line: str, headless: bool = False, tab: str = None,
          _provisioned: bool = False) -> int:
     """Send one request; print the reply; return the process exit code.
--- a/connectonion/cli/browser_agent/daemon.py
+++ b/connectonion/cli/browser_agent/daemon.py
@@ -1,11 +1,11 @@
 """
 Purpose: Persistent browser daemon — owns one BrowserAutomation and dispatches CLI requests to it over the platform transport (POSIX Unix socket / Windows named pipe), arbitrating concurrent agents through per-tab ownership.
 LLM-Note:
-  Dependencies: imports from [socket, os, sys, time, json, shlex, inspect, signal, atexit, threading, datetime, pathlib, useful_tools.browser_tools.BrowserAutomation, useful_tools.browser_tools.browser.driver_stealth_status, useful_tools.browser_tools.browser.installed_browser_path, browser_agent.agent (resolve_api_key, build_browser_agent), browser_agent.transport] | imported by [browser_agent/client.py (spawns via python -m), cli/commands/browser_commands.py (list_functions)] | tested by [tests/e2e/cli/test_browser_daemon.py]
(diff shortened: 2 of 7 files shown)


wu-changxing:
Review of `d6d7244c`: the client-side move is sound, but argument serialization is currently lossy.

`_format()` calls `shlex.quote()` only when a value contains a literal space. I reproduced:

```text
_format("type_text", {"text": "O'Reilly"})
=> "type_text --text=O'Reilly"
shlex.split(...) => ValueError: No closing quotation
```

Values containing backslashes, tabs/newlines, or other shell metacharacters can likewise be altered or rejected before the daemon sees them. These are normal browser inputs, not edge-only data. Quote every non-boolean value unconditionally with `shlex.quote(str(value))`, and add round-trip tests for an apostrophe, a Windows path, backslashes, and newline/tab content.

The existing CI is green because the tests cover space-containing values but not this serialization boundary.

wu-changxing:
Closing as superseded by #1139. Both move the `co browser do` model loop out of the serial daemon to resolve #933; #1139 is the newer, broader candidate and adds the current regression coverage plus the macOS wheel-install release gate. Please continue review and any follow-up fixes on #1139.

PR #943, opened 2026-08-13
fix(logger): scope metrics to the current turn

### LINKED ISSUE #942: Logger must scope per-turn tools and usage to the current trace boundary
## Problem

`Logger.log_turn()` writes one YAML entry per user turn, but it computes `tools_called`, tokens, and cost from the entire cumulative `session['trace']`.

In a continued session, turn 2 therefore repeats turn 1's tools and reports cumulative usage as if it belonged to turn 2. This conflicts with DD-017 (`One turn = one user input`) and DD-026's canonical per-turn trace outcome.

This was found while reviewing #941; it is separate from the eval verdict bug fixed there.

## Design options

1. **Keep cumulative metrics in every turn.** Rejected: fields named for one turn cannot be compared across runs if they include all earlier work.
2. **Subtract the previous YAML entry.** Rejected: restored sessions, partial logs, and provider totals make subtraction fragile.
3. **Slice the canonical trace at the current turn's `user_input` marker.** Recommended: it uses the same source of truth as Agent lifecycle and ACP outcomes, with a whole-trace fallback for legacy/test shapes.
4. **Read usage only from the terminal `turn_result`.** Useful for tokens/cost, but insufficient for `tools_called`; the logger would still need the turn slice, so using one slice for both is simpler.

## Acceptance criteria

- [ ] Multi-turn regression test: turn 2 YAML lists only turn 2 tools.
- [ ] Turn 2 token and cost metadata include only turn 2 `llm_result` entries.
- [ ] Single-turn and legacy traces without turn markers preserve existing behavior.
- [ ] The run YAML message history may remain cumulative; per-turn metrics must not be.
- [ ] Logger and usage tests pass.

## Related

- #940
- #941
- DD-017 Session Logging and Eval Format
- DD-026 Structured Turn Outcomes

wu-changxing:
## What changed

- centralize the canonical current-turn trace slice used by both evaluation and logging
- scope `tools_called`, tokens, and cost to the matching `user_input` boundary
- preserve whole-trace behavior for legacy or hand-built sessions without turn markers
- document the per-turn metric contract in DD-017 and the logging/eval guides

## Root cause

`Logger.log_turn()` read the cumulative session trace directly. On turn 2 it therefore credited turn 1's tools again and added turn 1's usage to the new turn. The evaluator had already learned the correct boundary in #941, but the invariant was private to that plugin.

## Impact

Multi-turn YAML now reports work and usage for the turn that owns it. Full messages remain cumulative for replay and conversation context.

## Validation

- `pytest -q tests/unit/test_logger.py tests/unit/test_eval_plugin.py tests/unit/test_the_token_count_matches_what_was_billed.py tests/unit/test_turn_outcome.py` — 88 passed
- regression proved red before the implementation and now verifies both eval summary and run YAML metrics
- `ruff check connectonion/core/trace.py`
- `ruff format --check connectonion/core/trace.py`
- `git diff --check`

Closes #942

### DIFF
--- a/connectonion/core/trace.py
+++ b/connectonion/core/trace.py
@@ -1,0 +1,22 @@
+"""
+Purpose: Select canonical per-turn slices from the cumulative Agent trace
+LLM-Note:
+  Dependencies: none | imported by [logger.py, useful_plugins/eval.py] | tested by [tests/unit/test_logger.py, tests/unit/test_eval_plugin.py]
+  Data flow: cumulative trace + current turn number -> trace beginning at that turn's user_input marker
+  State/Effects: pure; never mutates the trace
+  Errors: legacy or hand-built traces without a matching marker remain unchanged
+"""
+
+
+def current_turn_trace(trace: list[dict], turn: object) -> list[dict]:
+    """Return the canonical trace slice for one Agent turn.
+
+    Agent sessions retain earlier turns.  The latest matching ``user_input``
+    marker is the source-of-truth boundary; legacy traces without that marker
+    stay usable by falling back to the complete trace.
+    """
+    for index in range(len(trace) - 1, -1, -1):
+        entry = trace[index]
+        if entry.get("type") == "user_input" and entry.get("turn") == turn:
+            return trace[index:]
+    return trace
--- a/connectonion/logger.py
+++ b/connectonion/logger.py
@@ -272,17 +272,25 @@ class Logger:
         if self.eval_data is None:
             self._init_eval_file(user_input)
 
-        # Aggregate from trace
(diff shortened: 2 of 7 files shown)


wu-changxing (commented):
Self-review complete on exact head bf545041. No blocking findings. The shared pure function removes the already-demonstrated evaluator/logger drift without adding state or fallback complexity; matching user_input markers scope current-turn work, while markerless legacy traces keep existing behavior. The regression covers two cumulative turns, both YAML surfaces, tool isolation, token/cost isolation, and the existing eval boundary test. Focused result: 88 passed; new module Ruff and diff checks pass.

PR #947, opened 2026-08-13
test(relay): wait for reconnect behavior

### LINKED ISSUE #914: Make relay reconnect resilience test event-driven instead of sleep-based
## Problem

`TestRelaySupervisorResilience.test_supervisor_reconnects_after_non_connectionclosed_error` can fail on a busy CI runner even when the relay supervisor is correctly sleeping for its first one-second retry.

The test starts the supervisor, waits a fixed `asyncio.sleep(1.3)`, shuts it down, and expects at least two connection attempts. In ConnectOnion PR #913, Python 3.11 logged:

```text
[host] relay connection error (OSError('simulated network blip')); reconnecting in 1s
```

but `calls["connect"]` was still `1` when the 1.3-second assertion ran. The same unchanged relay code passed on Python 3.10, 3.12, and 3.13 in the same matrix. PR #913 only adds ACP parameters/mounting around `host()` and does not modify `_create_relay_lifespan` or this test.

## Best-practice fix

Synchronize on behavior instead of wall-clock slack:

- let the fake `connect` set an `asyncio.Event` on the second attempt;
- `await asyncio.wait_for(second_attempt.wait(), timeout=...)`;
- then shut down and assert the supervisor task was still alive/retrying.

This keeps the production one-second backoff testable without depending on CI scheduler latency. Do not shorten the production retry delay just to make the test fast.

## Acceptance criteria

- the regression still proves a non-`ConnectionClosed` exception does not kill the supervisor;
- the test does not use a fixed sleep to infer that a coroutine was scheduled;
- repeated runs are stable on supported Python versions;
- no behavior change to production relay backoff unless separately justified.

wu-changxing:
## Outcome

Make the relay supervisor resilience regression synchronize on the second connection attempt instead of guessing that CI scheduled it within a fixed 1.3-second sleep.

## Change

- the fake relay connection sets an `asyncio.Event` on attempt two
- the test waits for that observable behavior with a bounded `asyncio.wait_for`
- shutdown remains in `finally`, so a timeout or assertion cannot leak the supervisor task
- the production one-second first retry and later exponential backoff are unchanged

This keeps the test about its real contract: a non-`ConnectionClosed` exception does not kill the long-lived supervisor.

## Validation

- focused regression: passed
- focused regression repeated five consecutive times: 5/5 passed
- complete `tests/unit/test_relay.py`: 23 passed
- `git diff --check`: passed
- self-review: one test-only change; no production behavior or timing change

Fixes #914

### DIFF
--- a/tests/unit/test_relay.py
+++ b/tests/unit/test_relay.py
@@ -477,9 +477,12 @@ class TestRelaySupervisorResilience:
         from connectonion.network.host.server import _create_relay_lifespan
 
         calls = {"connect": 0}
+        second_attempt = asyncio.Event()
 
         async def boom(*args, **kwargs):
             calls["connect"] += 1
+            if calls["connect"] == 2:
+                second_attempt.set()
             raise OSError("simulated network blip")
 
         monkeypatch.setattr(relay_module, "connect", boom)
@@ -491,8 +494,10 @@ class TestRelaySupervisorResilience:
         )
 
         await on_startup()
-        await asyncio.sleep(1.3)   # initial attempt + at least one retry after the 1s backoff
-        await on_shutdown()
+        try:
+            await asyncio.wait_for(second_attempt.wait(), timeout=5)
+        finally:
+            await on_shutdown()
 
         assert calls["connect"] >= 2, (
             "supervisor died after the first error instead of reconnecting"


wu-changxing:
Self-review complete on exact head `ec4d1d5b`.

- Scope is one test-only synchronization change; production retry/backoff behavior is untouched.
- The assertion now waits for the observable second connection attempt instead of assuming a 1.3-second scheduler window.
- `try/finally` guarantees supervisor shutdown even if the bounded wait fails.
- Focused test passed 5 consecutive runs; `tests/unit/test_relay.py` passed 23/23; `git diff --check` passed.
- Exact-head GitHub matrix is green across Linux 3.10–3.13, Windows browser/E2E, lockfile, and CodeQL.

No correctness, security, or simplification findings remain. Leaving this PR open for maintainer review; it has not been merged.

wu-changxing (commented):
Exact-head maintainer-style review completed at `ec4d1d5b70fd0a80d2ad506292268827c6fec5da` against current `main@a89f6f47`.

The change is the smallest correct fix for #914: the test now synchronizes on the observable second connection attempt, bounds that wait, and always shuts the supervisor down in `finally`. It does not change production relay code, retry timing, or exception policy. A dead supervisor still fails by timeout, while a slow scheduler no longer creates a false negative.

Verification:
- exact PR head: `tests/unit/test_relay.py` 23/23 passed
- simulated merge with current `main@a89f6f47`: 23/23 passed
- merge is conflict-free; `git diff --check` passes
- all 11 GitHub checks are green

No blocking correctness, security, lifecycle-cleanup, or simplification finding remains.

Issue #894, opened 2026-08-11
Gate 1.7 beta on stdio, WebSocket, browser, and security conformance


wu-changxing:
Parent: #892

## Goal

Prove that stdio ACP, authenticated WebSocket ACP, and the bounded legacy browser adapter expose one coherent coding-agent product before 1.7 advances to beta.

## Scope

- Run one transport-independent lifecycle matrix against stdio and WebSocket: initialize, new, prompt, streamed message/thought/tool/plan, permission, mode, cancel, close, disconnect, resume, and errors.
- Add security vectors for identity, onboarding, replay, origin, ownership, approval correlation, authority ceilings, and cleanup.
- Run real-client smokes: O Chat browser, an external ACP WebSocket client, and at least one editor over stdio.
- Record known remote-transport v1 limits: no in-flight replay, host-owned reconnect/liveness, and WebSocket-only default server profile.
- Define fallback telemetry/support window and a separate evidence-based decision before legacy `/ws` removal.
- Update main docs, AI-readable docs, release checklist, and Design Journal.

## Beta gate

- [ ] planned Gateway, principal, relay-integrity, React, and O Chat slices are merged or explicitly deferred;
- [ ] the lifecycle and security matrices pass on supported Python/platform versions;
- [ ] real browser, external WebSocket, Zed/JetBrains, Codex, and Claude Code delegation/resume smokes pass;
- [ ] disconnect leaves no worker, lease, pending approval, or reusable ticket behind;
- [ ] old/new client compatibility and rollback are demonstrated;
- [ ] docs state exactly which transport/security features are preview, beta, or stable.

Related: #25, #298, #771, #792, #838.

wu-changxing:
## 1.7.0a2 Alpha conformance update — final candidate `77826c0d`

Draft release PR #891 now contains current `main@74d24357`, including #949/#951, plus candidate release notes. It remains Draft; no merge, tag, package publication, or deployment has occurred.

### Exact-head evidence now green

- 413/413 ACP unit tests;
- 9/9 stdio interoperability E2E (2 official Python ACP SDK + 7 raw subprocess framing/lease tests);
- real Codex app-server handshake/full turn and cross-process resume: 2/2;
- real candidate `co ai --json --max-iterations 1`: exact `ACP_E2E_OK`, JSON `error=null`, exit 0;
[...]

PR #949, opened 2026-08-13
fix: reject ambiguous ACP authority payloads

wu-changxing:
Closes #948
Related to #838 and #894.

## What changed

- reject ACP permission responses unless their JSON-RPC message contains exactly one success `result` shape;
- reject unknown non-`_meta` fields in the permission result and selected/cancelled outcome;
- reject unknown non-`_meta` fields in `session/cancel` params;
- keep standard ACP `_meta` extension objects opaque and non-authoritative;
- add regression coverage proving malformed owned permission responses are consumed once and become the existing hard rejection.

## Why

The pinned generated ACP Pydantic models ignore unknown object fields by default. Before this change, a malformed Host `ACP_RESPONSE` containing both `result` and `error` could retain an `allow_once` result and reach the permission mailbox as an approval. Unknown fields inside permission outcomes and cancellation params were also silently discarded.

Authority-bearing input should not inherit permissive parser behavior. DD-037 requires malformed permission responses to fail closed, while DD-038 defines an exact generation-scoped cancellation adapter. The strict checks stay local to those two Host inputs rather than forking the official SDK models or restricting unrelated protocol surfaces.

## Impact

Valid ACP v1.19 permission and cancel messages are unchanged. Standard `_meta` remains accepted, but cannot grant authority. Legacy Host events, canonical traces, session persistence, React, O Chat, and the SDK version are unchanged.

## Validation

- red-first reproduction: 5 new cases failed on current `main` while 37 existing focused cases passed;
- focused ACP wire/permission/cancel/mode/gateway matrix: 197 passed;
- all ACP-named unit files: 404 passed;
- official SDK and real stdio subprocess E2E: 9 passed;
- `ruff check` on all changed files: passed;
- `git diff --check`: passed;
- Linus-style simplicity review: no remaining findings;
- Aaron-style ConnectOnion/security review: no remaining findings.

## Review focus

- confirm the pinned v1.19 selected outcome permits `outcome`, `optionId`, and `_meta` while cancelled permits only its schema-defined discriminator;
- confirm invalid owned permission responses continue to be one-shot hard rejection rather than an ignored/retryable request;
- confirm no compatibility surface outside the two authority-bearing adapters is narrowed.

No merge, tag, package publication, or deployment is part of this Draft.

PR #951, opened 2026-08-13
fix: align ACP JSON-RPC IDs

### LINKED ISSUE #950: Align native ACP JSON-RPC ID validation across transports
Parent: #838

## Problem

The final ACP ID audit found an inconsistency between the two native transports.

DD-045 and the authenticated WebSocket gateway require an inbound JSON-RPC request ID to be a non-boolean string or integer before `initialize`. The strict stdio transport intends the same shape, but `_request_id()` uses `None` for both an invalid ID and the literal JSON value `null`. Its equality check therefore accepts all of these as valid inbound messages:

- a request with `"id": null`;
- a success response with `"id": null`;
- an error response with `"id": null`.

That gives stdio and WebSocket different correlation rules and lets an invalid ID reach the pinned SDK connection state.

## Best-practice decision

Use one explicit ID predicate at the strict stdio framing boundary:

- an inbound request or response ID is a string or integer, excluding booleans;
- JSON `null`, floats, arrays, objects, and booleans are invalid correlation IDs;
- a notification remains valid when the `id` member is absent;
- a notification with an explicit invalid `id` is invalid rather than being reclassified;
- parse/invalid-request error responses may still emit `"id": null` when no valid request ID can be reflected, as required by JSON-RPC error handling;
- do not change ACP session IDs, Host carrier UUIDs, SDK-generated outgoing integer IDs, or persistence.

Keep this validation in ConnectOnion's `_StrictNDJSONTransport`; do not fork the pinned ACP SDK or introduce a second ID type abstraction.

## Acceptance

- [x] stdio rejects request, success-response, and error-response frames whose explicit ID is null;
- [x] stdio rejects boolean and non-integer numeric IDs consistently;
- [x] valid string/integer request and response IDs continue to pass;
- [x] notifications without `id` continue to pass;
- [x] invalid-frame errors reflect a valid string/integer ID when available and otherwise use null;
- [x] strict transport, stdio subprocess, and official SDK tests pass;
- [ ] independent review confirms the stdio/WebSocket ID contracts now match.

## Rollback

Revert the predicate change. No stored sessions, traces, React/O Chat state, Host compatibility frames, or SDK version require migration.

Related: #894; DD-028, DD-032, DD-045.

wu-changxing:
Closes #950
Related to #838 and #894.

## What changed

- give the strict stdio transport one explicit ACP JSON-RPC correlation-ID predicate;
- accept only string or integer IDs, excluding booleans;
- reject explicit null, float, boolean, array, and object IDs on inbound requests/responses;
- keep notifications without an `id` valid;
- keep invalid-request error responses able to emit `id: null` when no valid ID can be reflected.

## Why

`_request_id()` previously returned `None` both for an invalid ID and for literal JSON `null`. `_is_json_rpc_message()` compared that result back to the original value, so request, success-response, and error-response frames with `id: null` all passed the strict stdio boundary.

The authenticated WebSocket gateway already requires a non-boolean string/integer ID before `initialize` under DD-045. This makes the stdio correlation contract match it without changing the pinned ACP SDK or session identity.

## Impact

Valid ACP clients are unchanged: the official SDK generates integer request IDs, string IDs remain valid, and notifications still omit `id`. Only malformed inbound correlation IDs are rejected before they can enter SDK connection state. Persistence, Host compatibility carriers, React, and O Chat are unchanged.

## Validation

- red-first reproduction: three null-ID request/response cases failed on current `main`;
- strict transport matrix: 13 passed;
- all ACP-named unit files: 403 passed before the final two existing invalid-ID assertions were added; the focused final matrix remains green;
- official SDK and real stdio subprocess E2E: 9 passed;
- `ruff check`: passed;
- `git diff --check`: passed;
- Linus-style simplicity review: no remaining findings;
- Aaron-style contract/security review: no remaining findings.

## Review focus

- confirm inbound correlation IDs intentionally match the WebSocket string/integer rule;
- confirm absent notification IDs remain distinct from explicit null IDs;
- confirm JSON-RPC error output may still use null when the invalid input supplied no reflectable ID.

No merge, tag, package publication, or deployment is part of this Draft.

### DIFF
--- a/connectonion/cli/co_ai/acp_transport.py
+++ b/connectonion/cli/co_ai/acp_transport.py
@@ -154,14 +154,18 @@ class _StrictNDJSONTransport:
             chunks.append(exc.partial)
             return b"".join(chunks)
 
-    @staticmethod
-    def _request_id(message: Any) -> str | int | None:
+    @classmethod
+    def _request_id(cls, message: Any) -> str | int | None:
         if not isinstance(message, dict):
             return None
         request_id = message.get("id")
-        if isinstance(request_id, bool):
-            return None
-        return request_id if isinstance(request_id, (str, int)) else None
+        return request_id if cls._is_request_id(request_id) else None
+
+    @staticmethod
+    def _is_request_id(value: Any) -> bool:
+        return isinstance(value, str) or (
+            isinstance(value, int) and not isinstance(value, bool)
+        )
 
     @classmethod
     def _is_json_rpc_message(cls, message: Any) -> bool:
@@ -170,13 +174,13 @@ class _StrictNDJSONTransport:
         if "method" in message:
             if not isinstance(message["method"], str):
                 return False
-            if "id" in message and cls._request_id(message) != message.get("id"):
+            if "id" in message and not cls._is_request_id(message["id"]):
                 return False
             params = message.get("params")
             return params is None or isinstance(params, (dict, list))
         if "id" not in message or (("result" in message) == ("error" in message)):
             return False
-        return cls._request_id(message) == message.get("id")
+        return cls._is_request_id(message["id"])
 
 
 def _reject_json_constant(value: str) -> None:
--- a/tests/unit/test_co_ai_acp.py
+++ b/tests/unit/test_co_ai_acp.py
@@ -1581,6 +1581,41 @@ async def test_strict_ndjson_returns_errors_and_keeps_reading():
     assert [item["id"] for item in errors] == [None, None]
 
 
+@pytest.mark.parametrize(
+    "message",
+    [
+        {"jsonrpc": "2.0", "id": None, "method": "initialize", "params": {}},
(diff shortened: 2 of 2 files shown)


wu-changxing:
Pre-merge review at exact head `96f33e5114b8ab57eb921e76e481412987be34d7`:

- GitHub: all 11 checks pass, including Python 3.10-3.13, Windows E2E/browser, and CodeQL.
- The explicit ID predicate matches the authenticated WebSocket boundary: string or integer, excluding booleans; an absent notification ID remains distinct from an explicit null ID.
- JSON-RPC parse/invalid-request errors still emit `id: null` when no valid correlation ID can be reflected.
- Linus simplicity review and Aaron contract/security review: no actionable findings.
- Combined with #949 on current `main`: 413 ACP unit tests and 9 official-SDK/real-stdio E2E tests pass; Ruff and `git diff --check` pass.

No GitHub review from another human is recorded. The author-side review and automated evidence are clean.

PR #961, opened 2026-08-13
fix: reject mixed ACP WebSocket envelopes

### LINKED ISSUE #960: Reject mixed JSON-RPC envelopes at the native ACP WebSocket boundary
Parent: #838
Beta conformance evidence: #894
Precedent: #956 / #957 and #948 / #949

## Reproduction

The native ACP WebSocket boundary and strict stdio boundary currently classify the same JSON-RPC envelope differently.

On current `main`, an authenticated `/acp` connection accepts and executes both of these mixed envelopes:

```json
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1},"result":{"foreign":true}}
{"jsonrpc":"2.0","id":2,"method":"session/resume","params":{"sessionId":"saved-session","cwd":"/tmp"},"error":{"code":-32000,"message":"foreign"}}
```

The first frame creates an Agent and returns a successful initialize response. After a valid initialize, the second frame executes `session/resume` and returns success. The pinned generated ACP models ignore the response-only field.

PR #957 already makes `_StrictNDJSONTransport` reject equivalent request/response-mixed frames with JSON-RPC `-32600`. The WebSocket Gateway validates only selected properties of the first initialize request, then `_ACPTransport.feed()` forwards every later dictionary directly to `AgentSideConnection`. This leaves the two native transports with different framing contracts and lets an ambiguous authority-bearing request execute remotely.

### 1. Let the generated ACP router ignore unknown envelope fields

Rejected. Generated models validate ACP semantics but do not own exact JSON-RPC envelope families. #949 and #957 already showed that ignored extras can preserve authority or side effects from an ambiguous payload.

### 2. Copy the stdio predicate into `acp_gateway.py`

Rejected. Two local classifiers have already drifted. Copying the same rules creates another future mismatch across IDs, request/notification fields, response families, and structured params.

### 3. Globally set every generated ACP model to `extra=forbid`

Rejected. This forks upstream behavior, changes far more than the transport invariant, and risks rejecting standard opaque `_meta` extensions inside ACP params/results.

### 4. Share one small pure JSON-RPC envelope classifier across both native transports

Recommended.

- Accept request/notification keys `jsonrpc`, `method`, optional `id`, optional structured `params`.
- Accept response keys `jsonrpc`, `id`, and exactly one of `result` or `error`.
- Accept only string/integer non-boolean correlation IDs; notifications omit `id`.
- Keep ACP `_meta` and all method-specific schema validation untouched inside params/results.
- Before Agent creation, a mixed `initialize` fails the existing WebSocket first-frame gate.
- After initialization, an invalid WebSocket envelope is not routed and receives the same `-32600 Invalid Request` shape as stdio through the existing bounded outbound queue.
- Keep the stdio class methods as thin compatibility wrappers if tests or local call sites rely on them; do not expose the whole transport implementation to the Gateway.

This follows DD-032: raw framing evidence owns facts hidden by SDK helpers. It also restores DD-045's requirement that the first frame be a JSON-RPC 2.0 initialize request before Agent construction and advances #894's cross-transport conformance gate.

## Acceptance criteria

- [ ] Red-first WebSocket vectors prove mixed initialize and post-initialize request envelopes currently execute.
- [ ] The shared classifier is the single source for native stdio and WebSocket exact envelope families.
- [ ] Mixed initialize fails before Agent construction.
- [ ] Post-initialize request+result, request+error, response+method, and response+params are rejected with `-32600` and no method side effect.
- [ ] Valid requests, notifications, success/error responses, string/integer IDs, and method-level `_meta` extensions remain supported.
- [ ] Existing WebSocket authentication, ticket, origin, lifecycle, cleanup, and backpressure behavior remains unchanged.
- [ ] Focused Gateway/stdio/official-SDK suites, Ruff, diff check, package build, and locked dependency audit pass.
- [ ] Exact-head simplicity/security review records any findings before maintainer review.

## Rollback

Revert the shared classifier and Gateway feed guard. No stored sessions, traces, React/O Chat state, SDK version, browser migration, or legacy Host event needs migration.

No merge, tag, package publication, React/O Chat change, or retired TypeScript SDK work is part of this issue.

wu-changxing:
## Summary

- share one exact JSON-RPC envelope classifier between ACP stdio and native WebSocket transports
- reject mixed request/response envelopes before SDK routing
- return `-32600` through the existing bounded WebSocket outbound queue after first-frame admission
- keep ACP method schemas and `_meta` extensions owned by the pinned official SDK

## Security behavior

Before this change, generated SDK models could ignore foreign envelope members on the native WebSocket path. A mixed `session/resume` request could therefore execute even when it also carried a response `error` or `result` member. Mixed `initialize` frames could construct an Agent.

After this change:

- a mixed first `initialize` frame closes with `4400` before Agent construction
- a mixed later envelope returns JSON-RPC `-32600`
- the mixed message is never delivered to `AgentSideConnection`
- error responses use the same bounded queue and backpressure as normal ACP output
- valid responses, notifications, method payloads, and SDK-owned `_meta` remain supported

## Design

This follows DD-032's rule that raw framing owns facts hidden by generated SDK models and extends DD-045 so stdio and WebSocket cannot drift again. The shared module validates envelope families only; it intentionally does not duplicate ACP method schemas.

## Verification

- red-first reproduction: 6 focused cases failed before implementation, then passed
- `143 passed`: full Gateway, stdio, and official-SDK ACP suites
- original end-to-end reproduction now reports no Agent for mixed initialize and no resume side effect for mixed resume
- Ruff lint passed on every changed Python file
- `uv lock --check` passed
- locked production dependency audit: no known vulnerabilities
- wheel and sdist built as `1.7.0a1`; Twine checks passed
- local Linus-style simplicity/security review completed with no blocking findings
- external-model review was not run because the environment's data-egress policy rejected sending the unpublished security diff

## Scope

No merge, release, tag, React/O Chat change, or retired TypeScript SDK change is included.

Closes #960
Related to #838 and #894

### DIFF
--- a/connectonion/cli/co_ai/acp_transport.py
+++ b/connectonion/cli/co_ai/acp_transport.py
@@ -12,6 +12,12 @@ from typing import Any
 from acp import RequestError, stdio_streams
 from acp.core import DEFAULT_STDIO_BUFFER_LIMIT_BYTES
 
+from ...core.acp_jsonrpc import (
+    acp_request_id,
+    is_acp_json_rpc_message,
+    is_acp_request_id,
+)
+
 
 class _BoundStdoutWriter:
     """Windows writer bound to original stdout with blocking write-all drain."""
@@ -156,41 +162,15 @@ class _StrictNDJSONTransport:
 
     @classmethod
     def _request_id(cls, message: Any) -> str | int | None:
-        if not isinstance(message, dict):
-            return None
-        request_id = message.get("id")
-        return request_id if cls._is_request_id(request_id) else None
+        return acp_request_id(message)
 
     @staticmethod
     def _is_request_id(value: Any) -> bool:
-        return isinstance(value, str) or (
-            isinstance(value, int) and not isinstance(value, bool)
-        )
+        return is_acp_request_id(value)
 
     @classmethod
     def _is_json_rpc_message(cls, message: Any) -> bool:
-        if not isinstance(message, dict) or message.get("jsonrpc") != "2.0":
-            return False
-        if "method" in message:
-            # Generated ACP models ignore unknown fields, so reject mixed
-            # request/response envelopes before authority-bearing routing.
-            if not set(message).issubset({"jsonrpc", "id", "method", "params"}):
-                return False
-            if not isinstance(message["method"], str):
-                return False
-            if "id" in message and not cls._is_request_id(message["id"]):
-                return False
-            return "params" not in message or isinstance(
-                message["params"], (dict, list)
-            )
-        has_result = "result" in message
-        has_error = "error" in message
-        if "id" not in message or has_result == has_error:
-            return False
-        expected = {"jsonrpc", "id", "result" if has_result else "error"}
-        if set(message) != expected:
-            return False
-        return cls._is_request_id(message["id"])
+        return is_acp_json_rpc_message(message)
 
 
 def _reject_json_constant(value: str) -> None:
--- a/connectonion/core/acp_jsonrpc.py
+++ b/connectonion/core/acp_jsonrpc.py
@@ -1,0 +1,47 @@
+"""Exact JSON-RPC envelope rules shared by native ACP transports.
+
+Method-specific ACP schemas remain owned by the pinned SDK. This module only
+keeps stdio and WebSocket from disagreeing about request, notification, and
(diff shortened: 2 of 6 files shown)


wu-changxing:

PR #891, opened 2026-08-11
release: prepare ConnectOnion 1.7.0a2

wu-changxing:
## Outcome

Prepare the second opt-in 1.7 alpha from reviewed current `main`. Stable users remain on `1.6.4`; this candidate is `1.7.0a2`.

This PR does not tag, publish, merge, or deploy anything. It remains Draft until every hard gate below is cleared and a maintainer reviews the exact final head.

## Product and design boundary

- Native browser ACP: authenticated `/acp` discovery, signed one-use tickets, caller-bound workspaces, bounded private sessions/storage/attachments, verified payment onboarding, permissions, cancellation, modes, public thoughts, canonical plans, tool activity, reconnect, and resume.
- Generic downward ACP: merged #901 adds `acp_agent` through the official typed ACP client and exact-version Claude Code, Codex, and Gemini routes. Model-visible choices stay bounded; argv, approval policy, workspace authority, child thoughts, and child plans stay operator/parent-owned.
- Pinned Codex ACP supports explicit Full Access only because its read-only mode cannot reliably gate shell or outbound network work; approval-aware calls continue to use the native `codex` tool.
- Pinned Gemini is one-turn and requires API-key, Vertex, or enterprise Code Assist authentication. Retired individual OAuth is not treated as readiness.
- `@connectonion/react@0.4.2-alpha.2` remains the browser protocol owner and O Chat remains its thin renderer. The retired standalone TypeScript SDK is not on this train.
- Direct loopback or TLS/WSS is the preview transport boundary; no relay end-to-end-encryption claim.
- Publication remains the immutable exact-tag workflow through PyPI Trusted Publishing and a protected GitHub environment. There is no workstation upload path.

## Candidate contents since `1.7.0a1`

- native ACP Gateway, fail-closed discovery, browser admission, workspace/session/storage/attachment boundaries, and payment onboarding
- permission/collaboration vocabulary, public plan/thought/tool events, cancellation, reconnect, and resume
- bounded native Claude Code live events and generic `acp_agent` delegation through Claude Code, Codex, and Gemini
- isolated `co ai --acp --state-dir PATH` mutable state and current-turn log/evaluation accounting
- JSON-RPC correlation-ID and authority-payload hardening
- restored 1.6.1–1.6.4 release history and fail-closed preview/stable documentation checks
- Alpha rather than Production/Stable PyPI metadata
- fail-closed real-provider CLI release gates

## Exact-head validation at `2a75a651`

[...]

PR #970, opened 2026-08-13
Let real_api tests see the operator's real credential store

wu-changxing:
Unblocks the real-Codex hard gate on #891. **Two real Codex provider tests go from 1 failed to 2 passed.**

## The problem

`tests/conftest.py`'s autouse `_never_touch_the_real_home` repoints `HOME` / `USERPROFILE` / `Path.home()` at a temp dir for every test. That fixture earns its place — a mocked-HTTP run of `co auth microsoft` once wrote fake tokens into the operator's real `~/.co/keys.env` and surfaced days later as "Microsoft session expired".

But provider CLIs keep their credentials under `$HOME` as well. `codex` reads an OAuth session from `~/.codex/auth.json`; with HOME repointed it finds none and the turn dies at the provider:

```
tests/e2e/real_api/test_real_codex.py::test_real_codex_session_resume
turn failed: unexpected status 401 Unauthorized:
Missing bearer or basic authentication in header, url: https://api.openai.com/v1/responses
```

## Root cause, by controlled experiment

Identical `codex()` call, one variable changed:

| condition | result |
|---|---|
| operator HOME, cwd = scratchpad | ✅ inference returned `CODEX_INFER_OK` |
| operator HOME, cwd = repo root | ✅ inference returned `REPO_CWD_OK` |
| **HOME → temp dir** (what the suite does) | ❌ **the exact same 401** |

Credentials were valid throughout: `auth_mode: chatgpt`, access token good until 2026-08-24. The harness was hiding them, not the provider rejecting them.

Only the resume test trips it because it runs a genuine inference turn. The handshake test exercises protocol alone and passed either way — which is why this looked like a partial provider failure rather than a harness bug.

## The change

Five lines: `real_api`-marked tests keep the operator's HOME; everything else is isolated exactly as before.

The rule is not "isolate HOME" but **"isolate HOME for tests that must not touch it."** `real_api` is the stated exception — reaching the operator's real account is its entire purpose, it is opt-in behind a marker, and it is deselected from the default run. The alternative (seeding a fake HOME with real credentials per provider test) copies secrets around, which is strictly worse.

## Tests

`tests/unit/test_real_api_tests_can_reach_real_credentials.py` asserts **both halves from inside real tests** rather than inspecting the fixture:

- an ordinary test still cannot see the operator's home
- a `real_api` test can

Verified red-first: without the change it is **1 failed / 1 passed** — the `real_api` half fails while the isolation guard keeps passing, which is exactly the half that must not move.

Real gate result with the change: `tests/e2e/real_api/test_real_codex.py` → **2 passed**.

One implementation note worth keeping: the operator's home is read from `pwd.getpwuid(os.getuid())`, not `Path.home()`, because the test asserting the contract is itself subject to it.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -45,7 +45,7 @@ load_dotenv()
 
 
 @pytest.fixture(autouse=True)
-def _never_touch_the_real_home(monkeypatch, tmp_path_factory):
+def _never_touch_the_real_home(request, monkeypatch, tmp_path_factory):
     """No test may read or write the operator's real ~/.co.
 
     `CliRunner.isolated_filesystem()` isolates the working directory and
@@ -58,7 +58,18 @@ def _never_touch_the_real_home(monkeypatch, tmp_path_factory):
     Isolating HOME is the only guard that holds: the paths are resolved deep
     inside the commands, from Path.home() and from AGENT_CONFIG_PATH, and a
     per-test patch has to be remembered by every future test.
+
+    `real_api` is the one exception, because provider CLIs keep their
+    credentials under $HOME as well: `codex` reads an OAuth session from
+    ~/.codex/auth.json, and with HOME repointed it finds none and the turn
+    dies at the provider with `401 Missing bearer`. Reaching the operator's
+    real account is the entire purpose of those tests, they are opt-in behind
+    a marker, and they are deselected from the default run — so they keep the
+    real HOME while every other test is isolated.
     """
+    if request.node.get_closest_marker("real_api"):
+        return
+
     home = tmp_path_factory.mktemp("home")
     monkeypatch.setenv("HOME", str(home))
     monkeypatch.setenv("USERPROFILE", str(home))       # Windows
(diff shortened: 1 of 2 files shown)


wu-changxing:
Review of `cd5e39db`: the exception is much broader than the credential access the Codex tests need.

`request.node.get_closest_marker("real_api")` disables HOME isolation completely. That exposes writable `~/.co/keys.env`, `~/.ssh`, cloud credentials, and every other home-scoped file to every present and future `real_api` test. A marked test that calls an auth command can therefore recreate the exact live-token corruption incident this fixture was introduced to prevent.

Keep the temporary HOME and expose only the required provider credential through a narrow, explicit fixture for the Codex tests (preferably read-only or copied into the isolated home). A generic marker should not grant arbitrary code read/write access to the operator’s entire home.

There is also a portability regression: the new unit module imports `pwd` unconditionally, so a normal Windows collection fails before fixtures run. The green Windows jobs do not collect this file. Please add a cross-platform way to capture the original home and a Windows collection test.

wu-changxing:
Update required before merge. The current patch disables HOME isolation entirely for every `real_api` test, exposing the operator's full `~/.co`, `~/.codex`, and `~/.claude` trees to opt-in tests. That solves provider login but also reopens the class of live-token mutation the fixture was created to prevent. Please narrow this to an explicit credential-staging/read-only mechanism (or a dedicated opt-in fixture requested only by provider-login tests), document which paths may be read, and add a regression proving unrelated real_api tests cannot write live OAuth/config state.

PR #972, opened 2026-08-13
fix: block ACP metadata parameter shadowing

### LINKED ISSUE #971: Reject ACP _meta keys that shadow request parameters
Parent: #838

## Problem

The final ACP `_meta`/extension audit found that the pinned `agent-client-protocol==0.12.0` router expands every `_meta` entry into the Python handler keyword arguments **after** it has populated the official request fields.

That means metadata can replace a parsed protocol value when its key matches the generated Python field name. A minimal router reproduction sends:

```json
{
  "sessionId": "visible-session",
  "cwd": "/visible",
  "mcpServers": [],
  "_meta": {
    "session_id": "hidden-session",
    "cwd": "/hidden"
  }
}
```

The handler receives `session_id="hidden-session"` and `cwd="/hidden"`.

This affects both native stdio and WebSocket because both eventually use the same SDK `build_agent_router`. It creates an ambiguous request boundary for `initialize`, `session/new`, `session/resume`, `session/set_mode`, `session/prompt`, `session/close`, and `session/cancel`.

The same defect exists when ConnectOnion acts as the client for the generic `acp_agent` tool: `build_client_router` expands `_meta` the same way for `session/update` and `session/request_permission`. A real callback reproduction with visible `sessionId="visible-wrong-session"` is cancelled normally, but adding `_meta.session_id="active-session"` changes the outcome to a selected permission option. The child process can therefore make the visible permission request disagree with the session checked by `ToolClient`.

Existing workspace binding, principal-scoped state, MCP policy, and Full-access launch authority still fail closed; this reproduction is not evidence of a cross-principal privilege escalation. It is nevertheless a protocol-integrity defect: the authenticated/logged wire fields need to be the fields the handler executes.

## Best-practice decision

Keep ACP extensibility, but reject metadata that can shadow an official handler parameter before the SDK router sees it:

- retain unrelated `_meta` entries unchanged;
- derive the reserved Python parameter names from the pinned official request models rather than maintaining a second camelCase conversion table;
- apply one shared model-derived rule to native stdio, native WebSocket, and the guarded child-agent stdio transport;
- return `InvalidParams` (`-32602`) with fixed public details and the valid request ID;
- for `session/cancel` notifications, drop the invalid notification; do not invent a response to a notification;
- keep the WebSocket first-message rule fail closed: a shadowed `initialize` closes with `4400` before Agent construction;
- do not fork/patch the installed SDK and do not reject all `_meta` values.

### Alternatives rejected

1. **Ignore the collision:** leaves visible protocol fields different from executed handler arguments.
2. **Reject all `_meta`:** breaks the ACP extension point and future signed/encrypted metadata work.
3. **Fork the SDK:** creates an unnecessary generated-schema maintenance branch for a small boundary compatibility guard.
4. **Match only keys present beside `_meta`:** misses optional official fields that the SDK materializes as `None` before metadata overwrites them.

## Acceptance

- [ ] stdio and WebSocket reject `_meta` keys that shadow official handler arguments;
- [ ] the generic `acp_agent` client rejects shadowed `session/update` and `session/request_permission` callbacks before `ToolClient` routing;
- [ ] post-admission request failures return exact `-32602` responses with fixed details and a valid reflected ID;
- [ ] a shadowed first WebSocket `initialize` closes before Agent construction;
- [ ] invalid notifications do not execute and do not receive JSON-RPC responses;
- [ ] unrelated metadata still reaches `**kwargs` unchanged;
- [ ] the reserved-key set is checked against the pinned official request models;
- [ ] initialization order, session ownership, workspace binding, permissions, and extension methods remain unchanged;
- [ ] strict transport, Gateway, official-SDK, native real-stdio, and child-process real-stdio tests pass;
- [ ] author-side simplicity/security review and independent maintainer review complete.

## Rollback

Revert the shared pre-router guard. No session snapshot, trace, React/O Chat state, ACP carrier, or release artifact requires migration.

Related: #894, #960, #962, #964; DD-032, DD-045.

wu-changxing:
## Summary

- reject ACP `_meta` keys that the pinned Python SDK router would promote over official request parameters
- share the same model-derived guard between native stdio, authenticated WebSocket, and generic `acp_agent` child callbacks
- preserve unrelated metadata and extension methods
- return fixed `InvalidParams` responses for requests, drop invalid notifications without a response, and keep the WebSocket first-frame gate fail closed before Agent construction

## Design

`agent-client-protocol==0.12.0` validates the official request/callback model, then expands `_meta` into Python handler kwargs. Because metadata is applied last, a raw peer can replace values such as `session_id`, `cwd`, `mode_id`, `prompt`, `update`, or permission options after validation.

The guard derives reserved names from the pinned official models. The strict NDJSON transport now lives once in `core` and is used for both Agent and downward Client process roles. It does not duplicate camelCase aliases, inspect metadata values, reject unrelated metadata, or patch the SDK. DD-032, DD-045, and DD-052 record the compatibility boundary and rollback.

This PR is stacked on #961 because that PR introduces the shared native JSON-RPC boundary module. Retarget it to `main` after #961 merges.

## Verification

- red-first reproduction proved both stdio and WebSocket executed shadowed values before the guard
- 294 native ACP/Gateway/official-SDK/real-stdio tests passed
- 66 generic child-agent/tool tests passed, including real subprocess permission and update shadow attempts
- method matrix covers the supported Agent methods plus `session/update` and `session/request_permission`
- Ruff and `git diff --check` passed
- exact lock check passed; locked production dependency audit found no vulnerabilities
- `connectonion-1.7.0a1` wheel and sdist built; both passed Twine validation
- author-side protocol/security/simplicity review found no blocking issue

Closes #971
Related: #838, #894, #960, #961

### DIFF
--- a/connectonion/cli/co_ai/acp_transport.py
+++ b/connectonion/cli/co_ai/acp_transport.py
@@ -3,22 +3,16 @@
 from __future__ import annotations
 
 import asyncio
-import json
 import platform
 import sys
-from contextlib import suppress
 from typing import Any
 
-from acp import RequestError, stdio_streams
+from acp import stdio_streams
 from acp.core import DEFAULT_STDIO_BUFFER_LIMIT_BYTES
 
-from ...core.acp_jsonrpc import (
-    ACP_META_SHADOW_ERROR_DETAILS,
-    acp_meta_shadows_request_params,
-    acp_request_id,
-    is_acp_json_rpc_message,
-    is_acp_request_id,
-)
+from ...core.acp_transport import StrictACPTransport
+
+_StrictNDJSONTransport = StrictACPTransport
 
 
 class _BoundStdoutWriter:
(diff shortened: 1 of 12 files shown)


wu-changxing:
Exact-head real-provider follow-up for 21cc0f99cff53d75871132b347d4e39faafccd9b:

- Direct ACP child adapters: Codex first turn plus resume and Claude Code first turn plus resume both passed (2/2, 43.36s).
- Shipped co ai path: co ai -> acp_agent -> Claude Code and co ai -> acp_agent -> Codex both returned their exact expected child results (2/2, 67.46s).
- These selected tests executed; none were skipped.

Together with the recorded 294 native ACP/Gateway tests, 66 generic child-agent tests, Ruff, lock check, package build/Twine, and zero-known-vulnerability locked dependency audit, no author-side correctness or security blocker is known on this exact head.

This remains a stacked Draft based on #961, so it has no main-target GitHub matrix yet and has no maintainer approval. Do not merge, tag, or publish from it.

wu-changxing:

PR #974, opened 2026-08-13
fix: require ACP wire field names

### LINKED ISSUE #973: Reject Python field aliases on ACP wire parameters
Parent: #838
Beta conformance evidence: #894
Related hardening: #948, #950, #960, #971

## Reproduction

The pinned agent-client-protocol 0.12.0 models enable populate_by_name for ergonomic Python construction. The SDK router validates raw JSON params with those same models, so native ACP currently accepts Python field names on the wire:

- initialize accepts protocol_version instead of protocolVersion;
- session/new and session/resume accept mcp_servers and additional_directories instead of mcpServers and additionalDirectories;
- session/resume, session/set_mode, session/prompt, session/close, and session/cancel accept session_id instead of sessionId;
- session/set_mode accepts mode_id instead of modeId;
- generic child callbacks likewise accept session_id and tool_call instead of sessionId and toolCall.

When both spellings are present, Pydantic silently selects the alias value and ignores the Python-name value. With only the Python name, the actual lifecycle handler runs. This creates a Python-specific ACP wire dialect and makes ambiguous payloads appear valid on both stdio and WebSocket.

The official ACP v1 extensibility contract says custom information belongs in _meta and implementations MUST NOT add custom root fields to protocol types. React already writes the standard sessionId and modeId forms; O Chat is protocol-free. No frontend migration is required.

### Accept and document both spellings

Rejected. Python model-construction ergonomics must not become a public wire dialect, and other ACP SDKs serialize the standard camelCase names.

### Disable populate_by_name globally or fork generated models

Rejected. That changes normal Python SDK construction and forks the pinned upstream package far beyond the transport boundary.

### Hand-write camelCase conversions

Rejected. Guessed string conversion drifts on names such as _meta and on future acronym or explicit-alias fields.

### Validate routed top-level params from pinned model aliases

Recommended.

- Keep one shared routed-method model registry for the ConnectOnion Agent handlers and implemented generic Client callbacks.
- Derive permitted top-level wire keys from each pinned model field alias, including _meta; never guess camelCase.
- Reject Python field names and other non-schema top-level params before the SDK router.
- Preserve arbitrary content inside _meta and preserve extension methods whose method name starts with underscore.
- Leave nested schema validation, optional/null semantics, enums, and result serialization to the official SDK.
- Requests receive fixed -32602 Invalid params without reflecting raw values; notifications are dropped without a response.
- A nonstandard first WebSocket initialize closes with 4400 before Agent construction.
- Apply the same rule to native stdio, native WebSocket, and the guarded generic child transport.

This extends the model-derived guard introduced by #971 instead of adding a second parser.

## Acceptance

- [ ] Red-first tests prove snake_case-only initialize and lifecycle params currently reach real handlers.
- [ ] Alias plus Python-name duplicates are rejected atomically with no side effect.
- [ ] Every currently routed Agent method and implemented Client callback has model-derived wire-key coverage.
- [ ] Requests return stable -32602; malformed notifications produce no response.
- [ ] WebSocket rejects nonstandard initialize before Agent construction and rejects later params without lifecycle side effects.
- [ ] Standard camelCase, optional/null forms, opaque _meta data, and underscore extension methods remain compatible.
- [ ] React ACP writers remain unchanged because they already use standard wire names.
- [ ] Native, Gateway, official-SDK, real subprocess, Ruff, lock, build, dependency-audit, and self-review gates pass.
- [ ] Maintainer reviews before merge.

## Rollback

Revert the model-derived top-level wire-key guard. No session, trace, React/O Chat state, SDK version, or legacy Host carrier needs migration.

No merge, tag, package publication, frontend release, or retired TypeScript SDK work is part of this issue.

wu-changxing:
Closes #973
Related to #838 and #894.

## What changed

- derive accepted top-level ACP wire parameter names from the explicit aliases on the pinned official request models;
- reject Python construction names such as `session_id`, `mode_id`, and `mcp_servers` before SDK routing;
- reject alias/name duplicates and non-schema root fields on the ConnectOnion-routed methods;
- apply one shared rule to native stdio, authenticated WebSocket, and guarded generic child callbacks;
- preserve standard optional/null values, opaque `_meta`, nested SDK validation, responses, and underscore extension methods.

## Why

The pinned Python SDK enables `populate_by_name` so local Python callers can construct generated models with snake_case names. Its raw JSON router validates with those same models, which unintentionally turns the Python names into a second wire dialect. A raw stdio or WebSocket peer could therefore send `session_id` instead of standard ACP `sessionId`; generic ACP children could send the same nonstandard spelling into `ToolClient`. When both spellings were present, the model silently selected the standard alias and discarded the Python-name field.

ACP reserves protocol-type root fields and requires custom data to live under `_meta`. Python construction ergonomics should not change that public wire contract.

## Impact

Standards-compliant clients are unchanged. React already emits `sessionId` and `modeId`, and O Chat remains protocol-free. No standalone TypeScript SDK work is involved. Invalid requests receive fixed JSON-RPC `-32602` details without reflecting input; invalid notifications are dropped without a response. No session, trace, or frontend-state migration is required.

## Validation

- red-first reproduction: six selected paths failed before the guard, proving stdio forwarding, WebSocket resume side effects, child result injection, and a real child permission prompt; the existing WebSocket first-frame `protocolVersion` check was already fail-closed;
- implementation-focused matrix: 18 passed;
- native ACP/Gateway/official-SDK/real-stdio matrix: 174 passed;
- all ACP-named unit and CLI E2E files: 537 passed;
- generic child/tool matrix: 68 passed;
- Ruff, `git diff --check`, and `uv lock --check`: passed;
- frozen production dependency audit: no known vulnerabilities;
- wheel and sdist built as `1.7.0a1`, Twine accepted both, and the wheel contains the shared guard modules;
- author-side protocol/security/simplicity review: no blocking finding.

## Review focus

- confirm the canonical names come from pinned model aliases rather than guessed camelCase conversion;
- confirm the registry contains exactly the methods ConnectOnion routes to implemented Agent handlers or Client callbacks;
- confirm `_meta`, extension methods, nullability, and nested validation remain owned by the official protocol layers;
- confirm invalid child permission input cannot reach the operator approval gate.

This is a stacked Draft based on #972 (`fix/acp-meta-shadow-971`). It should be retargeted to `main` only after its parent stack lands. No merge, tag, package publication, frontend release, or deployment is part of this PR.

### DIFF
--- a/connectonion/core/acp_jsonrpc.py
+++ b/connectonion/core/acp_jsonrpc.py
@@ -1,8 +1,9 @@
-"""JSON-RPC boundary rules shared by native ACP transports.
+"""JSON-RPC boundary rules shared by ConnectOnion ACP transports.
 
 Method-specific validation remains owned by the pinned SDK. This module keeps
-stdio and WebSocket aligned on envelope families and blocks metadata keys that
-the pinned SDK router would otherwise promote over validated request fields.
+stdio and WebSocket aligned on envelope families, preserves canonical wire
+names, and blocks metadata keys that the pinned SDK router would otherwise
+promote over validated request fields.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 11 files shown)


wu-changxing:
Exact-head real-provider validation for commit 9f172a69: 4/4 passed in 215.22s with no selected skips. This covered direct Codex first/resume, direct Claude first/resume, and co ai delegation through acp_agent to both providers. Official camelCase ACP traffic remains accepted. Author review found no blocking issue. This PR remains a stacked Draft; no merge or release action was taken.

wu-changxing:

Issue #975, opened 2026-08-13
Validate ACP JSON-RPC error responses without response loops


wu-changxing:
## Problem

The strict ACP envelope classifier is not self-consistent for JSON-RPC error responses.

On the exact stacked head at `9f172a69`:

- parse and invalid-request handling emits an error response with `id: null`, as JSON-RPC 2.0 requires when the request ID cannot be detected;
- `is_acp_json_rpc_message()` rejects that same response because every response ID currently has to be a string or non-boolean integer;
- an error object with `code: true` and `message: 42` is accepted because only result/error exclusivity is checked.

A strict ConnectOnion peer can therefore answer another peers valid null-ID error with a new Invalid Request error. More generally, malformed responses are currently treated like malformed requests even though JSON-RPC clients must not reply to responses. This can create a response loop and lets malformed error values reach the pinned SDK response path.

## Existing decisions

DD-032 keeps shared typed and framing evidence. DD-045 requires one exact request/response envelope classifier across stdio and WebSocket, but currently says every invalid later envelope receives `-32600`. JSON-RPC response semantics need a narrow clarification: invalid requests receive errors; responses never receive responses.

The JSON-RPC 2.0 response contract requires exactly one of `result` or `error`. Error must be an object with an integer `code`, string `message`, and optional arbitrary `data`. A response ID is normally the request ID, while an error whose request ID could not be detected uses `null`.

## Proposed design

1. Classify response candidates separately from requests and notifications before deciding whether an error reply is legal.
2. Keep ACP request IDs restricted to strings and non-boolean integers. Do not broaden request correlation IDs to null.
3. Accept `id: null` only on a structurally valid error response so parse/invalid-request errors are self-consistent.
4. Validate error objects at the shared envelope boundary: object, non-boolean integer `code`, string `message`, optional arbitrary `data`.
5. Never send a JSON-RPC response to a response candidate. A malformed response closes/fails the transport so outstanding work settles through the existing disconnect path; it is not silently converted into a request error.
6. Preserve valid result responses, valid error responses with string/integer IDs, request/notification behavior, `_meta`, official field aliases, and underscore extension methods.
7. Apply the same rule to native stdio, authenticated WebSocket, and the generic child ACP transport. No React, O Chat, or retired TypeScript SDK conversion is needed.

## Rejected alternatives

- Allow `None` in the shared request-ID predicate: this would also admit null request IDs and weaken the first initialize invariant.
[...]

PR #976, opened 2026-08-13
fix: validate ACP error responses

wu-changxing:
## Summary

- validate JSON-RPC error responses as `{code, message, data?}` and reject boolean/non-integer codes or non-string messages
- accept `id: null` only for error responses while keeping request and success IDs restricted to strings/non-boolean integers
- distinguish malformed response candidates from invalid requests so stdio and WebSocket fail the connection without ever answering a response
- document the response-loop boundary in DD-032, DD-045, and the `co ai --acp` guide

Addresses #975.

## Why

Before this change, ConnectOnion rejected the null-ID Parse Error it emitted itself, but accepted malformed error values such as `code: true` and `message: 42`. Echoing one emitted Parse Error into another strict transport produced a new `-32600`, which can become a response loop between strict peers.

The boundary now follows the narrow JSON-RPC rule: invalid request candidates may receive errors; response candidates never receive responses. A malformed response closes/fails the transport so existing disconnect cleanup settles pending work.

## Design

- one shared response-candidate predicate; no transport-specific protocol dialect
- one small error-object validator; ACP method/result schemas remain owned by the pinned official SDK
- no broadened null request IDs
- no React or O Chat conversion; the retired TypeScript SDK is untouched
- request/notification, `_meta`, official field aliases, arbitrary error `data`, and underscore extension behavior remain unchanged

## Verification

- red evidence on parent commit `9f172a69`: own null-ID error rejected, malformed error object accepted, and real stdio subprocess echoed a second `-32600`
- focused response matrix: 25 passed
- complete ACP unit + CLI E2E selection: 614 passed, 6261 deselected
- Ruff: passed
- `git diff --check`: passed
- `uv lock --check`: passed
- wheel and sdist build: passed
- Twine checks: passed
- wheel contains both shared ACP boundary modules
- frozen production dependency audit: no known vulnerabilities

## Author review

No blocking finding. Production behavior adds one shared classifier, one error-object validator, and the same short fail-closed branch in both transports. No new class, config layer, copied protocol schema, authority expansion, or raw peer value is exposed.

## Stack

This Draft is based on #974 (`fix/acp-wire-alias-973`). It should remain Draft until the parent chain lands, then be retargeted to `main` and run through the full main CI/reviewer gate. No merge, tag, or publish action is part of this PR.

### DIFF
--- a/connectonion/core/acp_jsonrpc.py
+++ b/connectonion/core/acp_jsonrpc.py
@@ -64,6 +64,21 @@ def acp_request_id(message: Any) -> str | int | None:
     return request_id if is_acp_request_id(request_id) else None
 
 
+def is_acp_json_rpc_response_candidate(message: Any) -> bool:
+    """Return whether an invalid envelope is recognizably a response.
+
+    A JSON-RPC peer must never answer a response. Keeping this classification
+    separate lets transports fail a malformed response without creating an
+    error-response loop.
+    """
+
+    return (
+        isinstance(message, dict)
+        and "method" not in message
+        and ("result" in message or "error" in message)
+    )
+
+
 def acp_meta_shadows_request_params(message: Any) -> bool:
     """Detect metadata keys that the pinned SDK would promote over ACP fields."""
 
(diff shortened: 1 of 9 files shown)


wu-changxing:
Exact-head real-provider validation for e9a5c1c0: 4/4 passed in 117.56s with no selected skips. Covered direct Codex first/resume, direct Claude first/resume, and co ai delegation through acp_agent to both providers. The response-loop regression also passed through a real stdio subprocess, and the complete ACP selection passed 614/614. Author simplicity/security review found no blocking issue. This PR remains a stacked Draft; no merge, tag, or publish action was taken.

wu-changxing:

PR #982, opened 2026-08-14
fix: normalize null ACP capabilities

wu-changxing:
## Summary

- normalize a schema-valid `agentCapabilities: null` initialize result to the pinned SDK's empty `AgentCapabilities` model
- preserve capability-driven continuation: prefer advertised `session/resume`, otherwise `session/load`, and never guess or fall back after a request
- extend the real stdio capability matrix and document the null-as-empty decision in DD-052 and the public tool guide

## Root cause

The official typed schema accepts an explicitly null `agentCapabilities` value. The continuation path assumed the decoded object was always present and dereferenced `.session_capabilities`, leaking a `NoneType` implementation error before it could issue the existing clear capability error.

The fix uses the existing pinned ACP model as the default. It does not add a local schema, exception catch, lifecycle retry, or new-session fallback.

## Design review

- **Chosen:** treat null as an empty optional-capability set and fail closed before a lifecycle request.
- **Rejected:** declare the valid initialize response malformed.
- **Rejected:** try resume/load without an advertisement.
- **Rejected:** catch `AttributeError` at the public envelope boundary.

Full alternatives and acceptance criteria are recorded in #981 before implementation.

## Validation

- reproduced red against the previous exact head: `ACP agent-null: 'NoneType' object has no attribute 'session_capabilities'`
- focused continuation/version matrix: 8 passed
- ACP unit + CLI selection: 628 passed, 6678 deselected
- authenticated real provider E2E: 4 passed
  - Claude Code first turn + cross-process resume
  - Codex first turn + cross-process resume
  - `co ai -> acp_agent -> Claude Code`
  - `co ai -> acp_agent -> Codex`
- `git diff --check` and Ruff lint pass
- `uv lock --check` passes
- wheel + sdist build as `1.7.0a1`; Twine checks both artifacts
- frozen production dependency audit: no known vulnerabilities
- Linus-style review: no helper, catch-all, fallback, or duplicate schema introduced

## Stack

Draft stacked on #980. Keep this Draft and do not merge directly until its parent stack has been reviewed, merged, and this PR is retargeted to `main` for full CI.

Closes #981

PR #984, opened 2026-08-14
fix(acp): reject coerced protocol versions

### LINKED ISSUE #983: acp_agent: reject coerced protocolVersion values before session creation
## Problem

`acp_agent` now compares the typed `InitializeResponse.protocol_version` to ACP v1 before creating or resuming a session. The pinned official Python SDK uses Pydantic's normal coercive validation, however, so wire values such as the JSON string `"1"` and boolean `true` both decode to Python integer `1`.

On the current exact stack head `23b6c4ff`, a real stdio ACP peer returning either invalid value receives `session/new`, `session/prompt`, and completes successfully. This means the client treats a schema-invalid version value as an agreed protocol major.

ACP initialization defines `protocolVersion` as an integer major and requires the peers to agree before using other methods. JSON booleans and strings are not integers even though Python and coercive Pydantic models can convert them.

## Reproduction

Return either initialize result from a real stdio peer:

```json
{"protocolVersion": "1", "agentCapabilities": {}}
```

```json
{"protocolVersion": true, "agentCapabilities": {}}
```

Then call a new `ACPAgent(...).acp_agent(...)`. Both currently return a successful session result instead of stopping after initialize.

## Design decision

Capture the raw `protocolVersion` for the correlated initialize response through the official connection's existing raw stream observer, then require `type(value) is int` before comparing its value to `acp.PROTOCOL_VERSION`.

- **Strict-check only the negotiated authority field — chosen.** The raw observer already receives the exact JSON-RPC request/response pair. Tracking the initialize request ID avoids assuming it is always `0`, and checking one field prevents schema duplication.
- **Make every ACP model globally strict — rejected.** That changes the pinned SDK's compatibility behavior across every request, response, metadata field, and third-party adapter without a protocol-wide migration plan.
- **Copy/validate the whole InitializeResponse schema locally — rejected.** The official SDK remains the typed schema owner; a second schema would drift.
- **Check only the typed value — rejected.** Coercion has already erased whether the peer sent an integer, string, or boolean.
- **Add method state to the shared transport — rejected for this fix.** The generic client already owns the initialize call and raw observer. Making the shared transport correlate method-specific response schemas would broaden both ACP process roles.

Missing/malformed initialize fields remain the SDK's validation job. This fix only preserves the raw type of the field that decides which protocol is safe to speak next.

## Acceptance criteria

- [ ] Real stdio fixtures prove string `"1"` and boolean `true` stop before every session request.
- [ ] Integer `1` continues normally; integer `2` retains the clear unsupported-version error.
- [ ] Correlation uses the actual initialize request ID rather than assuming a fixed ID.
- [ ] Error text reports the invalid wire value without exposing unrelated response data.
- [ ] Null/omitted capabilities, resume/load selection, and no-fallback behavior remain covered.
- [ ] Relevant ACP unit/CLI suites, authenticated Claude/Codex and `co ai` E2E, build, lockfile, and frozen dependency audit pass.
- [ ] DD-052 and the public tool guide record strict raw version-type negotiation.

## Release impact

Include in the 1.7.0 ACP preview hardening stack. Keep the issue open until its PR is merged.

wu-changxing:
Closes #983

## Decision

- capture the raw `protocolVersion` from the response correlated to the actual initialize request ID;
- require the wire value to be a JSON integer before comparing it with ACP v1;
- fail closed before `session/new`, `session/resume`, or `session/load`;
- keep the pinned official ACP models instead of copying the initialize schema or enabling global strict coercion.

## Stack

- base: #982 (`agentCapabilities: null` normalization)
- this PR contains only the protocol-version wire-type delta and its docs/tests

## Validation

- red proof: the parent accepted both `"1"` and `true`, created `sess-known`, and completed a prompt;
- focused version/correlation matrix: 4 passed;
- complete `test_acp_agent.py`: 73 passed;
- full ACP selection: 631 passed, 6935 deselected;
- authenticated real provider E2E: 5 passed (Claude first/resume, Codex first/resume, `co ai` to Claude, `co ai` to Codex, Gemini one turn);
- `uv lock --check`, Ruff, and `git diff --check` pass;
- 1.7.0a1 wheel and sdist build; both pass Twine;
- frozen production dependency audit: no known vulnerabilities.

Draft only. No merge or release is requested.

### DIFF
--- a/connectonion/useful_tools/_acp_agent_client.py
+++ b/connectonion/useful_tools/_acp_agent_client.py
@@ -214,10 +214,15 @@ async def run_agent(
                 startup_deadline,
                 timeout,
             )
-            if initialized.protocol_version != acp.PROTOCOL_VERSION:
+            if type(client.raw_protocol_version) is not int:
+                raise RuntimeError(
+                    f"{engine} returned ACP protocolVersion as "
+                    f"{type(client.raw_protocol_version).__name__}; expected integer"
+                )
+            if client.raw_protocol_version != acp.PROTOCOL_VERSION:
                 raise RuntimeError(
                     f"{engine} selected unsupported ACP protocol version "
-                    f"{initialized.protocol_version}; client supports "
+                    f"{client.raw_protocol_version}; client supports "
                     f"{acp.PROTOCOL_VERSION}"
                 )
             agent_capabilities = (
@@ -423,13 +428,24 @@ class ToolClient:
         self._in_prompt = False
         self._updates_seen = 0
         self._updates_handled = 0
+        self._initialize_request_id: str | int | None = None
+        self.raw_protocol_version: Any = None
 
     def observe_stream(self, event) -> None:
         """Count updates before their independently scheduled callbacks run."""
-        if (
-            getattr(event.direction, "value", "") == "incoming"
-            and event.message.get("method") == "session/update"
+        direction = getattr(event.direction, "value", "")
+        message = event.message
+        if direction == "outgoing" and message.get("method") == "initialize":
+            self._initialize_request_id = message.get("id")
+        elif (
+            direction == "incoming"
+            and self._initialize_request_id is not None
+            and message.get("id") == self._initialize_request_id
         ):
+            result = message.get("result")
+            if isinstance(result, dict):
+                self.raw_protocol_version = result.get("protocolVersion")
+        if direction == "incoming" and message.get("method") == "session/update":
             self._updates_seen += 1
 
     async def drain_updates(self, deadline: float, timeout: int) -> None:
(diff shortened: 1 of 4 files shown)


wu-changxing:
Exact-head review for `761f008cab514e3ebea54bc074602ef9c74dff4c`:

- parent red proof: wire `protocolVersion: "1"` and `true` were both coerced to v1, created `sess-known`, and completed the prompt;
- focused version/type/request-ID matrix: 4 passed;
- complete adapter unit file: 73 passed;
- full repository ACP selection: 631 passed, 6935 deselected;
- authenticated real-provider E2E: 5 passed in 128.56s (Claude first/resume, Codex first/resume, `co ai` to each, Gemini one turn);
- Ruff, `git diff --check`, and `uv lock --check` pass;
- 1.7.0a1 wheel and sdist build and both pass Twine;
- frozen production dependency audit reports no known vulnerabilities;
- Linus-style self-review found no copied schema, new helper, catch-all, retry, or lifecycle fallback.

Docs-site Alpha candidate is synchronized at `086eba727ee2fa23ee7800a17c8e42b17919b562` in Draft PR wu-changxing/connectonion-docs-site#49. Core PR remains Draft, mergeable, clean, and unmerged.

wu-changxing:

PR #993, opened 2026-08-14
docs: explain agent identity, and add an agent-identity skill

wu-changxing:
## Preview publication decision — 8 September 2026

The owner explicitly authorized technical-article model review and Gmail/Drive fixture writes, then requested publishing **1.8.4a1 preview first**, followed by local installed-package acceptance. The approved journal is now included. Final 1.8.4 stable still needs the live Gmail/Drive and physical NAS journeys; they do not block this explicitly requested opt-in preview. The earlier pending-export statements below are historical. All required current-head checks must pass before merge; none is waived.

Explain the live signing address, selected billing token and env metadata through the current global-env contract. The skill and guide now resolve the configured backend, define the token before using it, use a read-only account GET, and explain preserved server keys and filtered deployment credentials.

Refs #1444 and #1445. Foundation `3b5401f7` merged; candidate `09195364297aaa20daf5a70f247e08c3d99856e1`.

- **Proposed target version:** 1.8.4a1 preview; 1.8.4 stable after acceptance
- **Estimated release window:** next stable, after coordinated acceptance

Validation: 38 focused copy/identity/deploy tests pass. Both documented Python diagnostics execute against an isolated explicit env and mocked custom backend without printing the token. The combined candidate (including this change) passes **8555 tests, 23 skipped, 184 deselected**. No live credentials, deployed service or account was modified.

Explicit env selection may use an existing adjacent `.co/` key; otherwise the global identity remains the fallback. Default cwd changes select neither env nor identity. This is a documentation/skill repair, not a new authentication endpoint.


wu-changxing:
Review of `d1ea3fa2`: this is blocked by both CI and a broken diagnostic command.

All four Linux test jobs fail because the packaged skill hardcodes `https://oo.openonion.ai`; `test_no_packaged_network_client_hardcodes_the_production_origin` names `useful_skills/agent-identity/SKILL.md` as the offender. The skill must resolve the configured backend rather than bypassing `backend_url()`.

Separately, Step 2 decodes the token from `.env`, then runs:

```bash
curl -s -H "Authorization: Bearer $KEY" ...
```

but `KEY` is never assigned. Following the procedure literally sends an empty bearer and cannot confirm the payer. Make token acquisition/validation one explicit command (without printing the token), use the configured backend, and add an executable documentation test for the full sequence. Until then the advertised troubleshooting workflow does not work.

wu-changxing:
Update required before merge. CodeQL passed but the Tests workflow failed. Please rebase onto current `main`, inspect the failing job, and verify that the bundled skill registry, copied skill, docs links, and platform-specific commands pass the current suite. The identity documentation is valuable, but a 502-line docs/skill addition should not bypass a failing required check.

PR #995, opened 2026-08-14
fix: a deployed agent runs on its own account, not its author's

### LINKED ISSUE #994: co deploy --to ships the operator's identity, so the deployed agent bills its author
## What happens

`co deploy --to` writes the project `.env` verbatim to `/etc/connectonion/<agent>.env`, rewriting only `AGENT_CONFIG_PATH` (`deploy_to_server.py:_sync_env`). That file carries four lines that describe the **operator**, not the agent:

```
AGENT_ADDRESS
AGENT_EMAIL
IS_EMAIL_ACTIVE
OPENONION_API_KEY
```

They get there legitimately — `co init` copies them from `~/.co/keys.env` on purpose (`init.py:189`), so a project under development runs on its author's account with no second setup step. That is right on a laptop.

On a server it is wrong, and two of the four are not inert:

- **`AGENT_EMAIL` overrides the derived mailbox.** `address.load()` reads `os.getenv("AGENT_EMAIL", f"{address[:10]}@…")` (`address.py:305`), so the deployed agent sends mail as its author.
- **`OPENONION_API_KEY` decides who pays.** Billing keys off the `public_key` inside the JWT (`llm/billing.py:236`), not off `.co/keys/agent.key`. Every model call the agent makes is charged to the person who deployed it.

`AGENT_ADDRESS` sets nothing — the address always comes from the keypair — but a stale value there is what people read when they want to know who an agent is.

The agent is therefore cryptographically itself and financially its author, and **nothing reports an error**. `systemctl is-active` is green, `/info` shows the right address, `co status` shows a balance.

## Why it matters more than a wrong label

The spend cannot be separated afterwards. `usage_logs` records `(public_key, timestamp, model, tokens, cost_usd)` and no column naming the machine or the agent; `prompt`/`response` exist but are never written. Two processes sharing one API key produce **indistinguishable rows**.

Observed: one deployed agent ran this way for nine days. Its host's `.co/keys/agent.key` was correct the whole time, its `/info` reported its own address, and $1,350 of spend accumulated on a personal account that also served a laptop. Reconstructing the agent's share from run records recovered $1.94 of a $438 window — the rest is unattributable by construction.

Server rental is separable (each charge also lands in `payments` as `server:<machine_type>`); token spend on a shared key is not.

## Why it is not obvious

`env_inheritance.py` already names half of this problem — `PERSONAL_ACCOUNT_PREFIXES` keeps Google/Microsoft OAuth out of a new project, and its comment says *"`co deploy` then delivering them to a server made for an unrelated agent"*. The identity half was never covered, because propagating it is the **intended** behaviour of the command that writes it.

## Proposed fix

`co deploy --to` withholds the four identity keys and substitutes the agent's own, authenticated from the key the server holds:

- the deploy already derives that key (`derived_agent_identity`, SLIP-0013 `agent://<name>`) in order to install it, so it can sign `ConnectOnion-Auth-…` and call `POST /api/v1/auth` for a token belonging to the agent
- take the mailbox from the auth response rather than computing it — the backend routes the mail and the local `address[:10]` fallback has not always agreed with what it issues
- `--own-identity` mints on the machine, so this cannot be done from the operator's side. Write no account and say so, pointing at `co auth` in `/srv/<agent>`. An agent with no account fails visibly on its first model call; an agent quietly spending its author's credits does not fail at all
- name the key set once, in `env_inheritance.py`, and have `co init` and `co deploy --to` both read it — two lists of "what counts as identity" drift, and the one that drifts is the one nobody is looking at

Everything else in `.env` — third-party API keys, database URLs — must keep travelling unchanged.

## Ships in

1.6.5 (patch). This is a correctness fix on the stable line, not 1.7 preview work. There is no 1.6.x maintenance branch, so the PR is based on `main`; it needs cherry-picking for the tag.

wu-changxing:
Fixes #994. Target release: **1.6.5** (patch on the stable line — not 1.7 preview work). Based on `main` because there is no 1.6.x maintenance branch; needs cherry-picking for the tag.

## The bug

`co deploy --to` wrote the project `.env` to `/etc/connectonion/<agent>.env` verbatim. That file carries four lines describing the **operator**, put there deliberately by `co init` so a project runs as its author locally:

| | effect on the server |
|---|---|
| `AGENT_EMAIL` | overrides the mailbox `address.load()` derives — the agent sends mail as its author |
| `OPENONION_API_KEY` | billing keys off the JWT's `public_key`, so every model call charges its author |
| `AGENT_ADDRESS` | sets nothing, but it is what people read to learn who an agent is |
| `IS_EMAIL_ACTIVE` | travels with the above |

The agent was cryptographically itself and financially its author, with nothing failing. One ran that way for nine days; `usage_logs` has no column naming the machine, so the spend cannot be separated afterwards.

## The fix

`_env_for_server` withholds the identity keys and substitutes the agent's own, from `_agent_account()` — which signs `ConnectOnion-Auth-…` with the key the deploy already derived (`derived_agent_identity`, SLIP-0013 `agent://<name>`) and calls `POST /api/v1/auth`.

- **The mailbox comes from the auth response**, not from `address[:10]`. The backend routes the mail and the local fallback has not always matched what it issues — a computed-but-wrong mailbox fails by silently not receiving.
- **`--own-identity` writes no account and says so**, naming `co auth` in `/srv/<agent>`. The operator cannot sign as an agent whose key was minted on the machine, and falling back to their own account is the bug being fixed.
- **The key set is named once** in `env_inheritance.py`; `co init` and `co deploy --to` both read it. Propagating it is the point in one and a bug in the other, so neither keeps its own copy.

Third-party keys and database URLs travel unchanged.

## Tests

`tests/unit/test_a_deploy_runs_as_the_agent.py` — 10 tests that intercept the real `_sync_env` and decode the base64 body actually sent over ssh, rather than asserting on argv:

- the operator's address, mailbox and token do not reach the server
- the agent's own do
- application secrets and the `AGENT_CONFIG_PATH` rewrite are untouched
- **with no account, the operator's identity is still dropped** — this is the shipped bug's exact path
- `_agent_account` signs as the agent, takes the mailbox from the response, and returns `None` on a refused or unreachable backend instead of falling back
- `co init` and the deploy read one definition of identity

**Verified red first.** With `is_operator_identity` stubbed to `return False`, `test_without_an_account_the_operators_identity_is_still_dropped` fails with the operator's address and `aaron.xie@…` present in what was sent.

```
$ pytest tests/unit -k "deploy or env or init" -q
455 passed
```

One pre-existing test changed: `test_local_config_path_is_rewritten_for_the_server` asserted `OPENONION_API_KEY` survived `_env_for_server`. That was incidental to what it tests — that only `AGENT_CONFIG_PATH` is rewritten — so it now uses an application key, which still does pass through.

## Docs

`docs/network/deploy.md` gains "The agent runs as itself, not as you", including the `--own-identity` path and why a missing account is the safer failure. The concept doc is in #993.

### DIFF
--- a/connectonion/cli/commands/deploy_to_server.py
+++ b/connectonion/cli/commands/deploy_to_server.py
@@ -22,6 +22,7 @@ import yaml
 from dotenv import dotenv_values
 from rich.console import Console
 
+from .env_inheritance import is_operator_identity
 from .server_commands import SSH_TIMEOUT_SECONDS, derived_agent_identity, load_server
 
 console = Console()
@@ -640,7 +641,55 @@ RSYNC_FILTERS = [
 ]
 
 
-def _env_for_server(env_vars: dict, agent: str) -> dict:
+def _agent_account(agent_identity: dict) -> Optional[dict]:
+    """Authenticate as the agent, so the server bills the agent.
+
+    `/api/v1/auth` creates an account for any key that authenticates, so this
+    both mints a token and, on a first deploy, opens the agent's account. The
+    email comes from the response rather than being computed here: the backend
+    is authoritative about the mailbox it actually routes, and the local
+    fallback format has not always agreed with it.
+    """
+    import time
+
+    import requests
+    from nacl.signing import SigningKey
+
+    from ...backend import backend_url
+
+    public_key = agent_identity["address"]
+    timestamp = int(time.time())
+    message = f"ConnectOnion-Auth-{public_key}-{timestamp}"
+    signature = SigningKey(agent_identity["key_bytes"]).sign(
+        message.encode()).signature.hex()
+
+    try:
+        response = requests.post(
+            f"{backend_url()}/api/v1/auth",
+            json={"public_key": public_key, "signature": signature,
+                  "message": message},
+            timeout=15,
+        )
+    except requests.exceptions.RequestException:
+        return None
+    if response.status_code != 200:
+        return None
+
+    data = response.json()
+    token = data.get("token")
+    if not token:
+        return None
+    email = ((data.get("user") or {}).get("email") or {}).get("address")
+    return {
+        "OPENONION_API_KEY": token,
+        "AGENT_ADDRESS": public_key,
+        "AGENT_EMAIL": email or f"{public_key[:10]}@mail.openonion.ai",
+        "IS_EMAIL_ACTIVE": "true",
+    }
+
+
+def _env_for_server(env_vars: dict, agent: str,
+                    agent_account: Optional[dict] = None) -> dict:
     """The project's `.env`, corrected for the machine it is going to.
 
     `AGENT_CONFIG_PATH` is written by `co init` as an absolute path to the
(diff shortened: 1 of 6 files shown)


wu-changxing:
Post-merge review found a path that recreates the identity/account mismatch this PR was meant to remove.

`_ensure_setup()` deliberately preserves an existing remote `.co/keys/agent.key`:

```bash
if [ ! -f .../agent.key ]; then
  # install the locally derived key
fi
```

But the new code later calls `_agent_account(agent_identity)` using the **local derived key**, without establishing which key the server actually kept. On a server created with `--own-identity`, or any legacy server whose existing identity differs from derivation, deploy now writes `OPENONION_API_KEY`, `AGENT_EMAIL`, and `AGENT_ADDRESS` for one identity beside a different live private key — the exact silent drift described in this PR.

This needs an urgent follow-up on `main`: read the actual remote public identity first and authenticate with that identity (on the server when its private key is server-only). If it cannot be authenticated, deploy without an account rather than substituting the derived account. Add a regression where an existing remote key differs from `derived_agent_identity(agent)` and assert the emitted env never names the derived account.

Issue #938, opened 2026-08-13
[1.9] co voice: let the agent phone the owner and return a decision


wu-changxing:
## What this is, and how it differs from what already exists

`co phone notify` (#926, `openonion/oo-api#136`) shouts once: it rings a phone and reads a sentence. It cannot hear the answer. When an agent is blocked on a *decision* — should I publish this, which of these two, is this price right — a notification only tells the human to go and read their email.

**This is the two-way version: the agent calls you, explains what it is stuck on, listens to what you say about it, and hands the outcome back to the agent as something it can act on.**

Aaron's framing:

> 语音 Agent 要有整个的上下文，告诉它现在 AI 遇到了什么需求 … 如果用户同意的话，就把用户的需求整理好，发给 Agent，Agent 就知道了下一步应该怎么去做。

And the shape it should take:

> 我觉得这应该是一个命令行工具，这样的话别的也可以调用，就跟 co email 发送邮件是一样的，只需要提示，随便跟它说清楚就可以了。

## The interface

```
co voice call "I am listing the workshop on AllEvents. It wants a mobile
               verification I cannot pass. Options are: ask Aaron for a code,
               skip AllEvents, or try the voice-verification path. I need a
               decision before I can continue."
```

One argument: **the context, in plain language.** Same contract as `co email send` — you say what you mean and the tool carries it. No schema for the agent to fill in, because the thing on the other end is a conversation.

It returns when the call ends, with what was decided. The agent is blocked on this answer anyway; making it asynchronous would just mean inventing a second channel to deliver the result to a process that is sitting there waiting.

## Where it runs

Both credentials it needs are ours, so the session itself runs in oo-api:

```
co voice call "<context>"          the agent's machine
        │  POST /api/v1/voice/call
        ▼
oo-api                             holds both credentials
        ├── Twilio Media Streams (WSS)  ←→  the phone call
        └── Gemini Live (WSS)           ←→  the conversation
        │
        ▼  returns when the call ends
{ outcome, decision, transcript }
```

[...]

Issue #341, opened 2026-07-30
Make Discord a real inbound channel — not just a docs snippet


wu-changxing:
## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

## Motivation

`docs/examples.md:519-543` already sketches this: a Discord bot that calls `agent.input(message.content)` on incoming messages. But it's illustrative doc code — no test, not shipped as a tool/integration, not referenced from anywhere else in the repo. Right now the only way to get an agent responding in Discord is to hand-copy that snippet yourself, same as the situation email was in before #336.

Slack, Telegram, and SMS have the same shape of gap (a chat/messaging platform that could trigger `agent.input()` but doesn't, today), but **this issue is scoped to Discord only, on purpose** — CLAUDE.md's own principle is "avoid premature abstractions, wait until you need it 3 times." Email (#336) is the first real inbound channel; Discord would be the second. A shared "channel adapter" abstraction shouldn't be designed until a third one is actually being built — building it now, speculatively, for platforms nobody's implementing yet is exactly the premature abstraction the project says not to do.

## Scope

Turn the existing doc snippet into a real, tested, maintained integration:

```
Discord message (DM or @mention)
        │
        ▼
Discord bot process calls agent.input(message.content)
        │
        ▼
reply sent back to the same channel/DM
```

- Ship as a copyable tool/integration under `useful_tools/` or a CLI-adjacent piece, following the same pattern as `Gmail`/`Outlook`
- Trust: unlike raw email `From:` headers, Discord user IDs are authenticated by Discord itself — the open question is which users/roles/servers are allowed to trigger the agent, not whether the identity claim can be forged. Simpler question than #338's, don't block on that issue's outcome
- No reply-threading question the way email had — a channel reply or DM reply is the natural unit already

## Relationship to other issues

- #336/#337/#338 — the email precedent this follows the shape of, minus the threading and unauthenticated-sender problems (Discord doesn't have either in the same form)
- If/when a third channel (Slack, Telegram, SMS) becomes real work, that's the point to extract a shared inbound-channel abstraction from email + Discord's actual implementations — not before

## Definition of done

[...]

PR #1004, opened 2026-08-14
[1.9 Deferred] discord: post from the terminal and from an agent

### LINKED ISSUE #1003: [1.9] Add co discord send for release and agent notifications
## The gap

`co` can send email three ways and, since #866, a Telegram message. It still cannot post to Discord — where our own community lives and where our releases are announced.

#816 documents what that costs, and it is not hypothetical:

> Announcing 1.6.0 to Discord tonight meant hand-assembling JSON and curling a webhook … Inline `curl -d "{…}"` breaks on the backtick/backslash escaping (Discord replies 50109 "invalid JSON"), and python urllib gets 403 (Discord rejects its default User-Agent). **json.dumps → file → curl is the path that works.**

There is also a shipped incident: the download-count parse silently returned an empty string and the release post went out reading `• downloads` with no number (#300).

Both are the shape of thing a command owns once so nobody rediscovers it. That is the same argument that produced `co gmail` rather than everyone writing their own Gmail calls, and it is why #816 puts Discord third after calendar and Telegram.

## Scope

```
co discord send <channel> "message"
```

where `<channel>` is **either** a channel id (posts as the bot) **or** a webhook URL (posts as the webhook). Supporting both is not indulgence — the webhook is exactly what the release path already uses and what the escaping incident happened on, and the bot is what a general-purpose agent tool needs. The difference is one branch on whether the target starts with `https://`.

Plus `send_discord(channel, message)` as an agent tool, so an agent can post as well as a person.

## Where the credential goes

**Local, in `keys.env`** — same as Telegram in #866, and for the reason established there and in `openonion/oo-api#135`: the bot is the user's own, created in their own Discord developer portal, costs nothing, and a leaked token reaches only their bot.

That is the opposite of `openonion/oo-api#136`, where the Twilio account is ours, every call is on our bill, and a leaked credential reaches every user's call logs.

**The test is blast radius plus whether a policy exists that only we can enforce — not what the channel looks like.** "Chat and phone are both messaging, put them together" is the plausible-sounding mistake, and Discord is the third channel in a row where getting it right matters.

## Not in scope

- **Inbound** — a Discord message triggering `agent.input()` is #341, a different direction
- **Voice** — Discord bots genuinely can hold live voice conversations in a channel (people already run Gemini Live this way), which makes it the only free live-voice option we have. It cannot ring anyone, so it does not replace #938. Worth its own issue rather than being smuggled in here.

## Definition of done

- `co discord send` posts to a channel id and to a webhook URL, with the same command
- The escaping that broke the 1.6.0 announcement is handled by the tool, with a test carrying the exact characters that broke it
- Discord's own error text survives to the caller — 50109 and "unknown channel" are different problems with different fixes
- Exit non-zero when the message did not arrive, so the release path can tell
- No live Discord needed to run the tests

wu-changxing:
Closes #1003. Third of the outbound chat commands in #816, after Telegram (#866).

```
co discord send <channel-id> "message"
co discord send https://discord.com/api/webhooks/... "message"
```

plus `send_discord(channel, message)` as an agent tool.

## Scoped to notification, deliberately

**This PR is the outbound notification channel and nothing more.** A webhook and a bot are not two ways of saying the same thing — a webhook is a one-way pipe with no identity and no ears, while a bot can read, reply, be @-mentioned, join voice, and hold a role in a server. They overlap on exactly one verb: post a message.

For notification, **the webhook is not a compromise — it is the complete answer.** No bot, no developer application, no permission scopes, no token to leak: a URL is the whole configuration. The channel-id path is here because the release process and a general agent both benefit from posting as a real bot identity, and it is one branch on whether the target starts with `https://`.

Everything else a bot can do is a larger surface with a different setup cost and belongs in its own issue. #1003 has the capability table and a review of `discli`, an existing MIT-licensed Discord CLI aimed at agents that already covers much of that surface — worth reading before designing it, and too young (27 stars) to depend on.

## The two incidents this is built around

#816 recorded both, and both are pinned by a test rather than a comment:

**Escaping.** `curl -d "{…}"` broke on backtick and backslash quoting and Discord answered **50109 invalid JSON**, so the release skill accumulated a `json.dumps → file → curl` workaround. The test feeds the shape that broke it — backticks, `C:\Users\x`, quotes, a percent sign — and asserts it arrives byte-for-byte. Serialising in the library cannot make that mistake.

**403.** python urllib was rejected because Discord refuses its default User-Agent. There is a test asserting ours is never the default.

And the reason exit codes matter here: **#300 shipped a release post reading `• downloads` with no number and nobody noticed.** A refused post exits non-zero so the release path can tell.

## Three things the API will not do for us

- **A webhook URL is itself the credential.** It never appears in a result, in an error, or in a transport exception's string form — `requests` puts the URL in the exception, so the failure is reported by exception class instead. Tested with a URL containing a recognisable secret.
- **A bot token appearing in Discord's own error text is redacted.**
- **Over 2000 characters is refused here.** Discord rejects the whole post rather than truncating, so a long release note would silently not appear. The error names both the limit and the actual length.

## Where the credential lives

Local, in `keys.env`, same as Telegram in #866. The bot is the user's own Discord application, costs nothing, and a leak reaches only their bot.

That is the opposite of `openonion/oo-api#136`, where the Twilio account is ours and a leaked credential reaches every user's call logs. **Third channel in a row on the same rule: blast radius plus whether a policy exists that only we can enforce — not what the channel looks like.**

## Not in scope

- **Inbound** (a Discord message triggering `agent.input()`) is #341
- **Voice.** Discord bots genuinely can hold live voice conversations in a channel, and people already run Gemini Live that way — it is the only free live-voice option available to us. It cannot ring anyone, so it does not replace #938. Noted in #1003 as deserving its own issue.

## Tests

12 unit tests, no live Discord. `471 passed` across these, the Telegram tests, the import-cost guard, and the full `tests/e2e/cli` suite.

### DIFF
--- a/connectonion/cli/commands/discord_commands.py
+++ b/connectonion/cli/commands/discord_commands.py
@@ -1,0 +1,30 @@
+"""
+Purpose: `co discord send` — post to Discord from the terminal
+LLM-Note:
+  Dependencies: imports from [sys, rich.console, rich.text, useful_tools.discord.send_discord] | imported by [cli/main.py via discord_send()] | tested by [tests/unit/test_discord.py]
+  Data flow: handle_discord_send(channel, message) → send_discord() → prints the outcome → exit 0 on success, 1 on failure
+  State/Effects: one HTTP POST via the tool | no local state
+  Integration: same shape as handle_telegram_send — a thin handler over the tool an agent already has
+  Errors: a refused post or a missing token prints Discord's own reason and exits 1
+"""
+
+import sys
+
+from rich.console import Console
+from rich.text import Text
+
+from ...useful_tools.discord import send_discord
+
+console = Console()
+
+
+def handle_discord_send(channel: str, message: str) -> int:
+    """Post one message and say what happened."""
+    result = send_discord(channel, message)
+
+    if not result["success"]:
+        console.print(Text(str(result["error"]), style="red"))
+        sys.exit(1)
+
+    console.print(Text(f"Posted to {result['channel']}", style="green"))
+    return 0
--- a/connectonion/cli/main.py
+++ b/connectonion/cli/main.py
@@ -158,6 +158,7 @@ def _show_help():
     console.print("  [green]email[/green]             Send and read agent email")
     console.print("  [green]gmail[/green]             Send and read Gmail (co auth google)")
     console.print("  [green]telegram[/green]          Send a message from your Telegram bot")
+    console.print("  [green]discord[/green]           Post to Discord, as your bot or via a webhook")
     console.print("  [green]gdrive[/green]            List and transfer Google Drive files (co auth google)")
     console.print("  [green]syno[/green]              Browse and transfer Synology NAS files (co syno login)")
     console.print("  [green]outlook[/green]           Manage Outlook email and contacts (co auth microsoft)")
@@ -779,6 +780,22 @@ def email_upgrade(
 
 # Telegram command group. The bot is the user's own (@BotFather), so the token
 # lives in their keys.env -- no OpenOnion credential and nothing billed.
+# Discord command group. The bot is the user's own application, so the token
+# lives in their keys.env -- same as Telegram. A webhook URL needs no token.
+discord_app = _typer_app(help="Post a message to Discord, as your bot or through a webhook.")
+app.add_typer(discord_app, name="discord")
+
+
+@discord_app.command("send")
+def discord_send(
+    channel: str = typer.Argument(..., help="Channel id, or a webhook URL"),
+    message: str = typer.Argument(..., help="The text to post (max 2000 chars)"),
+):
+    """Post a message to Discord."""
+    from .commands.discord_commands import handle_discord_send
+    handle_discord_send(channel, message)
+
+
 telegram_app = _typer_app(help="Send a message from your Telegram bot.")
 app.add_typer(telegram_app, name="telegram")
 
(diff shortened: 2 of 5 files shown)


wu-changxing:
Review of Draft head `05a24eeb`: one security boundary is still missing.

`send_discord()` treats **every** string beginning with `https://` as a Discord webhook. I reproduced `send_discord("https://example.com/capture", "private release note")` posting to that URL and returning `success: True` on a 204 response. Because this is agent-facing, the function currently provides a generic outbound POST/exfiltration primitive while reporting the destination as “webhook”.

Parse and validate the URL against the documented Discord webhook host/path shape before sending; reject userinfo, unexpected ports/hosts/paths, and redirects (or revalidate every redirect hop). Add hostile-host and redirect tests. The focused suite is green (**13 passed**), but it only tests a genuine Discord-looking URL and therefore does not cover this boundary.

wu-changxing:
Closing this Draft implementation as deferred from the current stable release boundary. The capability remains tracked by #1003/#816, but outbound Discord is not part of the reliable coding-session release in #1154. Reopen or create a fresh current-main PR when channel expansion is the active train.

wu-changxing:
Release target correction from Aaron: outbound Discord is **deferred to 1.9**, not cancelled. Keep #1003 as the canonical requirement. This implementation PR remains closed so the 1.9 work starts from current main and the then-current channel contract.

PR #1009, opened 2026-08-14
release: 1.6.5 — a deployed agent runs on its own account

### LINKED ISSUE #994: co deploy --to ships the operator's identity, so the deployed agent bills its author

wu-changxing:
Cuts **1.6.5** on the stable line, carrying the fix from #995 (closes #994).

## Why this number

Latest stable is **1.6.4**; `VERSIONING.md` rule 1 is "increment by 1 for each release", so this is 1.6.5. It is a patch, not a `.0` — nothing here is a milestone being claimed.

## Why this branch

Cut from **`v1.6.4`, not `main`.** `main` is at `1.7.0a2` and carries the whole 1.7 preview train (ACP, the `co ai` orchestrator rewrite, plan-mode removal). Releasing 1.6.5 from `main` would ship all of that as a stable release. `v1.6.4` is not an ancestor of `main`; the 1.6.x line is maintained as successive `release/1.6.N` branches, and this follows that.

The fix is a clean cherry-pick of `c18548bb` — one commit, no conflicts.

## What it fixes

`co deploy --to` copied the project `.env` to the server verbatim, including the four identity lines `co init` writes so a project runs as its author locally. The deployed agent was therefore cryptographically itself and financially its author: it sent mail from the operator's mailbox and charged every model call to the operator's balance. Nothing failed, so nothing surfaced it — one agent ran that way for nine days, and `usage_logs` has no column naming the machine, so that spend cannot be separated after the fact.

## Verification

- Cherry-pick applied with no conflicts.
- **The regression test was proven red without the fix.** Reverting only the three source files and keeping `tests/unit/test_a_deploy_runs_as_the_agent.py` fails 10 tests; restoring the fix passes them. The test protects the behaviour rather than merely accompanying it.
- `test_the_version_agrees_with_itself.py` — 7/7. It caught `uv.lock` still naming 1.6.4; `uv lock` refreshed it to 1.6.5.
- Full offline suite (`-m "not slow and not real_api and not network"`) running locally alongside CI.

## Docs site

`docs-site/lib/version.ts` gains the stable bump to `1.6.5` (preview stays `1.7.0a2`). Shipped on its own branch in the docs repo, built with `next build --webpack` before pushing.

> **Correction to an earlier revision of this description.** It claimed the docs site had been advertising `1.6.0` while PyPI carried `1.6.4` — "four patch releases behind". That was wrong. I read `lib/version.ts` from a feature branch that happened to be checked out, not from `origin/main`, which was already correct at `1.6.4` / `1.7.0a2`. There was no drift. The irony is not lost: the mistake was reading whatever state was in front of me instead of the current one.

## Release checklist

- [x] Version chosen and the patch-vs-minor reason stated
- [x] `pyproject.toml`, `connectonion/_version.py`, `VERSIONING.md` (current + history entry)
- [x] `uv.lock` refreshed
- [x] `docs-site/lib/version.ts` (stable channel)
- [ ] CI green
- [ ] Design Journal post drafted (stable release — required by `docs/releases.md`)
- [ ] Tag and publish from the merged commit


wu-changxing:
v1.6.5 is released: tagged from this branch tip (3ffe0199), published to PyPI (https://pypi.org/project/connectonion/1.6.5/), GitHub Release up with both artifacts. Closing unmerged per the 1.6.x precedent — every stable-train release is tag-only from its release branch; main stays on the 1.7 pre-release track (v1.6.4 was shipped the same way). The branch carries your deploy-identity fix plus the three mail-CLI changes (#1010/#1016/#1017 cherry-picked, CI green, 1113-test slice green). Thanks for prepping the branch — the deploy-identity fix is the headline of the release notes.

Issue #1007, opened 2026-08-14
send_email tool: add optional from_address parameter


wu-changxing:
Companion to openonion/oo-api#150 (backend half — ships first, is backward compatible).

`send_email(to, subject, message, idempotency_key=None)` in `cli/co_ai/useful_tools/send_email.py` has no way to choose which of the account's addresses to send from — it always sends as whatever `get_sender_email()` resolves server-side. Once accounts can own multiple addresses (oo-api#148), add an optional `from_address` parameter here that is passed through to the `/api/v1/email/send` POST body. Omitted → current default behavior, unchanged.

**Target release**: v1.6.5 (next available stable patch — latest tag is v1.6.4). Ship only after oo-api#150 is live in production.

**Label**: enhancement (new capability, not a bug fix).

PR #1010, opened 2026-08-14
docs: cli-skill-design methodology, and the mail skill rewritten against it

### LINKED ISSUE #1008: Generalize browser-workflow-skill-builder into a CLI-skill design methodology, apply it to the email skill
We already have `useful_skills/browser-workflow-skill-builder/` — the methodology behind how `co-browser` was designed. Generalize it into a repo-wide "how to design a good CLI skill" methodology, then use it to rewrite `useful_skills/co-mail-and-drive/` (the email skill).

Two properties every skill built this way must have:

**(a) Tip-tested discoverability**: every command execution surfaces a next-step tip, and the methodology defines how to evaluate whether a given tip is actually useful (does the next call an agent makes, using only that tip, do the right thing?) — not just that a tip exists.

**(b) Self-diagnosing, self-correcting execution**: the skill's `help`/usage output must enumerate every capability (no hidden commands an agent has to guess at), and every error path must double as a fix-it guide — exit codes plus error text tell the agent what to do next, the same contract `co-browser`'s exit 0/1/2/3/4 table already follows.

Deliverable: a new `useful_skills/cli-skill-design/SKILL.md` (or a generalized rename of `browser-workflow-skill-builder`, TBD which reads better once written) capturing (a) and (b) as checkable criteria, then `co-mail-and-drive` rewritten against it as the first real case — including covering the multi-address / from-address changes from oo-api#148/#150 and connectonion#1007 once those ship.

**Target release**: v1.6.5 (next available stable patch — latest tag is v1.6.4). Sequence last, after the email feature work it documents actually exists.

**Label**: enhancement (methodology + doc work, not a bug fix).

wu-changxing:
## What changed

Two deliverables from #1008.

**1. New `connectonion/useful_skills/cli-skill-design/SKILL.md`** — the methodology
behind how `co-browser` was designed, generalized to any `co <thing>` command surface,
with the two required properties written as checks you run rather than advice you agree with.

**2. `connectonion/useful_skills/co-mail-and-drive/SKILL.md` rewritten against it** as the
first real case, covering `co gmail`, `co outlook`, `co email` and `co gdrive`.

Plus a `docs/useful_skills/cli-skill-design.md` page + README row (mirrors how
`co-browser` and `browser-workflow-skill-builder` are listed), and a one-line
cross-link from `browser-workflow-skill-builder` to the new skill.

Target release: **v1.6.5** (latest published tag is v1.6.4). No version file touched,
nothing tagged, nothing published.

## Naming: new skill, not a rename

The issue left this TBD. I wrote it first and then compared: `browser-workflow-skill-builder`
is about building a skill that drives a *website* through `co browser` — DOM analysis,
selector heuristics, hash-verified one-shot submits, `run_page_script` contracts. Almost none
of that generalizes, and the parts that do (dry-run first, evidence gates) are a small
fraction of its 617 lines. Renaming it would have meant either dragging 500 lines of DOM
advice into a general methodology, or deleting hard-won browser lessons to make the title
true. They are different axes: one is *how to design a command surface*, the other is
*how to build a workflow on top of one*. So: a new skill, ~185 lines, with a cross-link in
each direction. Neither is in `DEFAULT_LIBRARY_SKILLS` — both are builder tools, not
customer-facing workflow.

## How (a) is made checkable

**The tip test.** A tip is good iff an agent holding *only that tip* makes the right next
call. The skill specifies the harness: feed the command's output and the goal to a
**text-only** `llm_do` call (pinned model), no SKILL.md, no `--help`, no history, and ask
for exactly one shell command. Pass = the reply exists and advances the goal.

Text-only matters, and I learned it by getting it wrong: my first attempt used `co ai`,
which has a shell, so it *executed* the command it picked (`co gmail read 1`, then
`co auth google`) against a real account. That warning is now in the skill.

Run on 8 real tips from this surface (`co/gemini-2.5-flash`), **5/8 passed**:

| output given | goal | model replied | pass |
|---|---|---|---|
| `co gmail inbox` piped (no tip, just `ID:` lines) | read the newest email | `readmail 18f2a` | ❌ |
| `Marked read. Reply with: co gmail reply <#> <message>` | reply "thanks" | `co gmail reply 1 "thanks"` | ✅ |
| `Download one with: co gdrive get <#>` | download #1 to ~/Downloads | `co gdrive get 1 ~/Downloads` | ⚠️ names the right command, invents a positional for `--to` (would exit 2) |
| `❌ Google account not connected … co auth google` | check the inbox | `co auth google` | ✅ |
| `Retry the same command with --idempotency-key <key>` | resend without double-sending | `!! --idempotency-key k-123` | ❌ |
| `No email #3 in your last listing — run co gmail to refresh.` | read the third email | `co gmail && co gmail 3` | ❌ |
| `Claim it: co email name aaron --buy` | claim it | `co email name aaron --buy` | ✅ |
| `Cancel one with: co outlook cancel <#>` | cancel the first | `co outlook cancel 1` | ✅ |

The three failures are exactly the three rules the skill states: the tip must contain the
literal command name, it must survive piping, and it must reach the goal rather than stop
one step short. Each failure is a real gap in today's CLI, not in the doc — see "left for
later".

## How (b) is made checkable

**`--help` enumerates everything** — diff it against the skill, both directions:

```bash
g=gmail
co $g --help | sed -n '/─ Commands/,$p' | grep -oE '^│ [a-z-]+' | awk '{print $2}' | sort -u
grep -oE "co $g [a-z-]+" SKILL.md | awk '{print $3}' | sort -u
```

Ran for `email`, `gmail`, `outlook`, `gdrive`, `outlook contact` and `email sent`: after
adding an explicit `co outlook` example block (the old skill covered `inbox/read/search/sent`
only in prose — "takes the same shape" — so a grep for `co outlook read` found nothing),
**both directions are empty for all six groups**. No CLI command is undocumented, and no
documented command is missing from `--help`.

**Every exit code provoked, not asserted.** Run against a throwaway `HOME` so no credentials
were present:

| exit | provoked by | printed | names a next command |
|---|---|---|---|
| `0` | `co email inbox`, `co email send a@b.com S B` (unauthenticated) | `❌ No API key found` … `co auth` | yes — but exits **0** |
| `1` | `co gmail inbox` / `co outlook inbox` / `co gdrive list` | `❌ … not connected` + `co auth google` / `co auth microsoft` | yes |
| `2` | `co gmail bogus` | `No such command 'bogus'.` + `--help` hint | yes |
| `2` | `co gmail send` | `Missing argument 'TO'.` | yes |

The `co email` row is the headline finding: every `co email` failure path `return`s instead
of `raise typer.Exit(1)`, so an agent chaining `co email send … && next` walks straight past
a failed send. The rewritten skill therefore opens with co-browser's rule verbatim —
**read the output, not just the exit code** — and the exit table has a dedicated
"exit 0, `❌` on stdout" row.

## What I verified about the email surface, and how

Every command/flag below came from running `--help` on this branch
(`PYTHONPATH=/tmp/co-1008 python -m connectonion.cli.main …`, since the installed `co` is
1.5.10 and would have lied) plus reading the handlers.

- `co email`: `send` (`--idempotency-key`), `inbox` (`-n/-u`), `read`, `sent` (`-n/--to`,
  and `sent read`), `name` (`--buy`), `upgrade` (`-d/-a/--keep-address`); bare = inbox.
  **No `reply`, no `search`, no attachments, no scheduling** — the skill says so explicitly.
- `co email read` takes the **server id printed in the `#` column**, not a row position, and
  only searches the last 100 received (`email_commands.py:196-220`). Different contract from
  gmail/outlook/gdrive, which resolve row numbers through `~/.co/{gmail_last_inbox,outlook_last_inbox,gdrive_last_list}.json`.
- `co email inbox -u` filters **client-side after** fetching `-n`, so `-n 10 -u` means
  "unread among the last 10" (`email_commands.py:76`, with the comment saying the endpoint
  ignores the param).
- `co email sent` can return "not available on this backend yet" on a 404.
- **Old skill was wrong**: it said "Outlook additionally supports attachments". `co gmail send`
  has `--cc/--bcc/--attach` too. Limits differ and are checked before the send: Gmail
  25MB (`GMAIL_ATTACHMENT_LIMIT`), Outlook 3MB (`OUTLOOK_ATTACHMENT_LIMIT = 3_000_000`).
- **Old skill was missing**: `co outlook download/scheduled/cancel`, `co gmail search -n`,
  `co gdrive put --name`, and the fact that `co gmail` has *no* attachment download at all.
- `co gmail read` marks read **only** with the `gmail.modify` scope; with readonly+send it
  prints the body and says nothing was marked (`gmail_commands.py:147-153`).
- `sent` does **not** write the numbering cache — after `co gmail sent`, `read 1` still opens
  row 1 of the older inbox listing. New in the skill.
- Drive export table checked against `EXPORT_FORMATS` in `useful_tools/gdrive.py`: Docs→`.md`,
  Sheets→`.csv`, Slides→`.pdf`, **Drawings→`.pdf`** (the old skill omitted Drawings); other
  Google-native types raise "it cannot be downloaded".
- Piped forms verified from code: gmail/outlook print `N.` + `ID:` lines, gdrive prints
  `name\ttype\tsize\tid`, outlook contacts print `name\temail\tid` — so the old skill's
  `cut -f4` and `grep "ID:"` examples are correct and are kept.
- **Multi-address is not shipped**: `grep -rn from_address connectonion/` returns nothing, and
  `send_email()` still takes only `(to, subject, message, idempotency_key)` with the sender read
  from `AGENT_EMAIL`. The skill says so in a dated paragraph naming oo-api#148/#150 and #1007,
  and states that "send from another address" is not currently possible.

## Tests

`python -m pytest tests/ -q -x --timeout=300 -k "skill"` → **386 passed, 7140 deselected, 121.91s**.
Also checked that the new frontmatter parses with the repo's canonical reader
(`_parse_skill_content`) and declares no `tools:` (so it auto-approves nothing).

## Deliberately left for later

- **No CLI code changed.** The three tip-test failures are code gaps, not doc gaps, and fixing
  them here would mix a docs PR with behavior changes: (1) listing tips are inside
  `if console.is_terminal:` in all four renderers, so the piped caller — every agent — gets no
  tip; (2) the idempotency retry hint does not restate the command; (3) the stale-number message
  says "run co gmail to refresh" but not "then `co gmail read <#>`". Filing as a follow-up issue.
- **`co email` failure paths still exit 0.** Documented rather than changed — flipping exit codes
  is a behavior change for anything already scripting against it.
- No `docs/useful_skills/co-mail-and-drive.md` page: that mirror is partial today (no page for
  `co-mail-and-drive`, `commit`, `review-pr`, `find-selling-points`) and filling it is out of scope.
- `cli-skill-design` is **not** added to `DEFAULT_LIBRARY_SKILLS` — it is a builder tool, and the
  allowlist is deliberately the customer-facing set.

## Unsure / worth a second opinion

- `co email inbox` prints a Rich box table with **no** piped-plain branch, unlike the other three
  surfaces. At the default 80-column non-terminal width, long subjects will be clipped. I described
  the piped behavior of gmail/outlook/gdrive precisely and stayed quiet about `co email`'s exact
  piped rendering rather than guess at it — I could not exercise it without live credentials.
- The `co gdrive get` tip result is scored ⚠️ rather than ❌: the model named the right command but
  passed `~/Downloads` positionally, which `get` does not accept. Whether that counts as a tip
  failure or a flag-discovery failure is a judgement call; I left it visible rather than rounding it.

Closes #1008

### DIFF
--- a/connectonion/useful_skills/browser-workflow-skill-builder/SKILL.md
+++ b/connectonion/useful_skills/browser-workflow-skill-builder/SKILL.md
@@ -15,6 +15,8 @@ Create or improve browser automation skills for logged-in web apps. Use this whe
 
 The core pattern is: **save context -> analyze DOM/CSS -> write skill-local JS extraction -> verify same item by hash -> click/type with generic browser tools -> screenshot/context verify**.
 
+Designing the *command surface* of a `co <thing>` CLI and its SKILL.md, rather than a browser workflow on top of one? That is the sibling skill `cli-skill-design`.
+
 ## Architecture Rule
 
 Do not add site-specific Python to the core browser tool.
--- a/connectonion/useful_skills/cli-skill-design/SKILL.md
+++ b/connectonion/useful_skills/cli-skill-design/SKILL.md
@@ -1,0 +1,186 @@
+---
+name: cli-skill-design
+description: Design a `co <thing>` CLI surface and its SKILL.md together so an agent can drive it without guessing — every command ends by naming the next one, `--help` lists everything, and every failure says what to run instead. Use when adding a new CLI command group, writing or rewriting a SKILL.md for one, or auditing an existing one.
+---
+
+# Designing a CLI skill
+
+A CLI skill is two files that have to agree: the command surface (`co <thing> ...`)
+and the `SKILL.md` that tells an agent how to drive it. Design them together —
+the agent's whole world is *what it types* and *what comes back*.
+
+`co-browser` is the worked example. Read it before you write anything: its exit-code
+table, its "read the output, not just the exit code" rule, and its Done checklist are
+what this methodology generalizes. (Building a skill that drives a *website* through
+`co browser`? That is the sibling skill `browser-workflow-skill-builder` — DOM,
+selectors, verification scripts. This one is about the command surface itself.)
+
+Two properties. Each has a test you run and paste the result of — not a principle
+you assert in the PR.
+
+## (a) Tip-tested discoverability
+
+**Rule:** every command execution — success *and* failure — ends by naming the next
+command, spelled out, with the argument shape filled in.
+
+```
+Read one with: co gmail read <#>        ✅ names the command
+See the docs for more options            ❌ names nothing
+```
+
+### The tip test
+
+A tip is good if an agent that has *only that tip* makes the right next call. That
+is testable, so test it:
+
+```python
+from connectonion import llm_do
+
+llm_do(
+    f"You just ran a shell command. Its full output was:\n\n{out}\n\n"
+    "Your goal: read the newest email. Reply with ONE shell command and nothing else.",
+    model="co/gemini-2.5-flash",
+)
+```
+
+Use a **text-only** call (`llm_do`), not `co ai`. An agent with a shell will run the
+command it picks — measured: the first attempt at this test executed `co gmail read 1`
+and then `co auth google` against a real account. You are grading the reply, not the
+mailbox.
+
+- **Pass** — the reply is a command that exists and advances the goal (`co gmail read 1`).
+- **Fail** — it invents a name (`co gmail open 1`), asks for help, or replies with prose.
+
+Rules for the harness, or the result means nothing:
+- Give it the **output only**. No `SKILL.md`, no `--help`, no conversation history —
+  those are exactly the crutches the tip exists to replace.
+- Pin the model so a rerun compares like with like.
+- Run it per command, not once. Score the whole surface in a table and paste it
+  into the PR:
+
+  | command | tip printed | goal given to the fresh agent | it replied | pass |
+  |---|---|---|---|---|
+
+- Anything that fails: fix the tip, not the test.
+
+### What makes a tip pass
+
+- It contains the **literal command name**, not a description of it.
+- Placeholders say where the value comes from: `<#> from this listing`, not `<id>`.
+- **The tip survives piping.** Agents always pipe. A tip inside
+  `if console.is_terminal:` is invisible to every caller that needs it, and manual
+  testing never catches it because a human runs in a terminal. Check every one:
+
+  ```bash
+  co <thing> <cmd> | cat        # the tip must still be there
+  ```
+
+- **One** next step. Two tips is a fork, and the agent resolves a fork by guessing.
+- Failures get tips too, and the tip is the fix (see (b)).
+
+Measured on the mail surface (8 tips, `co/gemini-2.5-flash`, 2026-08): 5 passed. The
+three failures are the three rules above, each in its pure form —
+
+- a piped listing prints **no** tip, and the model invented `readmail 18f2a`;
+- `Retry the same command with --idempotency-key <key>` never names the command, and
+  the model replied `!! --idempotency-key k-123`;
+- `run co gmail to refresh` stops one step short of the goal, and the model replied
+  `co gmail && co gmail 3` — a command that does not exist.
+
+A tip that reads fine to a human fails this test. That is the point of running it.
+
+## (b) Self-diagnosing, self-correcting execution
+
+### Rule 1 — `--help` enumerates every capability
+
+An agent that cannot find a command from `--help` will invent one, and an invented
+command name costs a round trip every time. So: no hidden commands, no capability
+that only `SKILL.md` knows about.
+
+**Check it, both directions:**
+
+```bash
+g=<thing>
+# every subcommand the CLI has
+co $g --help | sed -n '/─ Commands/,$p' | grep -oE '^│ [a-z-]+' | awk '{print $2}' | sort -u
+# every command the skill mentions
+grep -oE "co $g [a-z-]+" SKILL.md | awk '{print $3}' | sort -u
+```
+
+Diff the two lists. Every CLI command must be either documented or deliberately
+skipped (say which, and why, in the PR). Every command the skill mentions must
+exist — a skill naming a command that `--help` does not list is a documentation bug,
+and it is the failure mode this check exists to catch.
+
+Repeat one level down for command groups (`co outlook contact --help`).
+
+### Rule 2 — every error path is a fix-it guide
+
+The exit code says *what kind* of problem; the text says *what to run*. Both, every
+time. Follow `co-browser`'s contract: a small, stable set of codes, and a table in
+`SKILL.md` whose right-hand column is a command, not an adjective.
+
+**Check it by producing each row.** For every exit code your surface can return,
+write down the command that provokes it and run it:
+
+```bash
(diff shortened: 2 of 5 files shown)


wu-changxing:
## Review (self-review)

Verified the load-bearing claims by running things rather than reading the summary.

**Independently confirmed:**
- `co email` exits 0 on failure — reproduced: `HOME=$(mktemp -d) co email inbox` prints `❌ No API key found` and exits 0, while `co gmail inbox` exits 1 in the same situation. Confirmed in source too: `handle_email_send` prints `❌ Failed:` and returns normally, so `co email send … && echo delivered` prints `delivered` when nothing was delivered.
- The old doc's "Outlook additionally supports attachments" was indeed wrong: `co gmail send --help` lists `--cc`, `--bcc`, `--attach/-a`.
- Using `PYTHONPATH` against the worktree instead of the installed `co` (1.5.10) was the right call — the installed binary would have documented a different version's surface.

**One thing this PR should have done and did not: file the exit-0 defect.** It was documented as a caveat in the rewritten skill and left there. A verified defect that silently breaks scripted and agent use is an issue, not a footnote — especially for this PR, whose whole thesis is that an error path must double as a fix-it signal, and an error that exits 0 cannot be one. Now filed as #1012. The skill's caveat should link it so the workaround has an expiry date rather than reading as permanent behaviour.

**The naming decision is right, and for the stated reason.** Writing it first and comparing after is the correct order. 617 lines of DOM heuristics, selector strategy and page-script contracts do not generalize to "how to design a command surface"; renaming would have forced either hauling that into a general methodology or deleting browser lessons to make a title honest. Separate skills with cross-links is the honest shape.

**The tip test's text-only constraint is the most valuable thing here, and it was learned the hard way** — the first attempt used `co ai`, which has a shell and actually executed `co gmail read 1` and `co auth google` against the real account. An evaluation harness that performs the actions it is grading is not a harness. That warning belonging in the skill is exactly right; it is the kind of thing the next person would otherwise rediscover the same expensive way.

**5/8 passing, reported as 5/8.** Reporting a methodology's first real run as a partial failure — and filing #1011 for the three gaps as *code* problems rather than quietly rewording the tips to score better — is the right instinct. Rewriting the tips to pass would have optimized the score and left the surface exactly as unhelpful.

**Honest gaps, appropriately left visible**: `co email inbox`'s piped rendering undocumented rather than guessed (no live credentials to exercise it); `co gdrive get` scored ⚠️ rather than ❌ with the judgement call stated. Both are better than a clean-looking table with invented entries.

**Bidirectional `--help` diffing across six command groups, with exit codes provoked against a throwaway HOME rather than asserted**, is the property (b) evidence that actually means something — it measures the surface instead of asserting the doc matches it.

No blocking issues. Merge once #1012 is linked from the skill's caveat.

PR #1021, opened 2026-08-15
Make co/gemini-3.7-flash the default model everywhere (#1002)

wu-changxing:
Makes `co/gemini-3.7-flash` the default model everywhere a model can be omitted, per #1002. The exact upstream ID `gemini-3.7-flash` (GA, verified live against Google's API) is the one oo-api registers.

**Merge ordering: do NOT merge until openonion/oo-api#146 is deployed to production.** The client default must not ship before the backend accepts the same exact model ID. `co/gemini-3.6-flash` stays priced, routed, and on the free-models list as the rollback — reverting is one constant, no user-config migration.

Explicit user configuration is untouched: only signature/option defaults changed; nothing rewrites a stored `MODEL=` or a pinned `model=` argument, and tests pin that.

## One source of truth

`DEFAULT_MODEL = "co/gemini-3.7-flash"` now lives in `connectonion/core/usage.py` next to the other model-fact tables, and every runtime default imports it. Previously "the default" was answered by independent literals that drifted.

## Default sites changed

Now importing `DEFAULT_MODEL`:
- `connectonion/core/agent.py:70` — `Agent` signature default
- `connectonion/llm_do.py:235` — `llm_do` signature default
- `connectonion/transcribe.py:86` — `transcribe` signature default
- `connectonion/core/llm.py:1056` — `OpenOnionLLM` (managed route)
- `connectonion/cli/main.py:351` — `co ai --model` option default
- `connectonion/cli/co_ai/agent.py:116` — `create_agent` default
- `connectonion/cli/github_action.py:435` — `CO_ACTION_MODEL` fallback

Literals updated (cannot import Python, or are frontmatter/templates):
- `action.yml:11` and `.github/workflows/co-ai-review.yml:13` — GitHub Action `model` input default
- `connectonion/core/llm.py:617` — `GeminiLLM` bring-your-own-Google-key default (`gemini-3.7-flash`, bare ID)
- `connectonion/core/llm.py:1030` — `gemini-3.7-flash` added to the provider routing map (3.6 kept)
- `connectonion/cli/co_ai/agents/registry.py:30,36` — explore/plan subagent models
- `connectonion/useful_plugins/subagents.py:21,166`, `subagents/loader.py:10,149` — subagent frontmatter default
- `connectonion/useful_plugins/builtin_agents/{explore,plan}/AGENT.md` — shipped frontmatter
- `connectonion/useful_plugins/eval.py:65` — `DEFAULT_SCORING_MODEL`
[...]

PR #1024, opened 2026-08-15
release: 1.6.6 — Gemini 3.7 Flash by default, and the CLI stops hiding features

wu-changxing:
Stable-train release branch (tag-only per 1.6.x precedent — this PR exists for the CI gate and will be closed unmerged like #1009).

Carries, adapted to the 1.6 line:
- default model → `co/gemini-3.7-flash` everywhere (native port of #1021; the 1.7-track file shapes made a direct cherry-pick conflict-heavy, so the equivalent change was applied natively with main's red-first default tests)
- `co transfer`, `co email addresses`, piping-safe tips (cherry-pick of #1023; two conflicts resolved surgically against what the tips tests pin)
- gemini-3.7-flash pricing: list rate $1.50/$7.50, cached at the standard 25% rule ($0.375) — 3.6's $0.15 cached is a published exception that does not extend to 3.7; the cached-rate rule test caught the initial copy-paste of the exception, which is exactly what it exists for
- 1M context row, free-list entry (3.6 kept as rollback)

Unit+CLI suites: 5721 passed / 1 failed → the cached-rate failure above, fixed properly (rule, not exception); re-run green. Backend prerequisite (oo-api#146) deployed and verified serving `co/gemini-3.7-flash` before this branch was cut.


wu-changxing:
v1.6.6 released: tagged from this branch tip, on PyPI (https://pypi.org/project/connectonion/1.6.6/), GitHub Release up with both artifacts. Closed unmerged per the 1.6.x tag-only precedent; CI ran green here first (including Windows), and the clean-venv smoke drove co --version, co transfer --help, co email addresses, the 3.7 default in co ai --help, and the exit-1 failure contract.

PR #1032, opened 2026-08-15
fix: allow append and command-output redirects

### LINKED ISSUE #1028: prefer_write_tool blocks appends and output redirects, with no allowed alternative — agents burn their full iteration budget writing nothing

mikemikimike:
## Description
Allow append redirects, `tee`, and ordinary command-output redirects in `prefer_write_tool` while continuing to block model-authored overwrite redirects and heredocs.

## Related Issue
Fixes #1028

## How this was written
- [ ] I wrote this myself
- [ ] Mostly AI-generated

The least confident area is shell syntax outside the covered patterns; this change intentionally keeps the existing heuristic rather than introducing a shell parser. I did not run the full suite because this Windows environment has unrelated dependency/configuration issues. If the rule is still too broad, the first breakage would be a legitimate command-output or append operation being blocked again.

## Type of Change
- [x] Bug fix (non-breaking change which fixes the issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Dev Blog
`docs/blog/2026-08-15-when-logging-looked-like-authoring.md`

## Changes Made
- Restrict hard blocking to heredocs and single `echo`/`printf` overwrite redirects.
- Allow `>>`, `tee`, and command stdout redirects.
- Add regression coverage for append, logging, output redirection, and stderr redirect boundaries.
- Add the required design-journal post.

## Testing

`PYTHONUTF8=1 python -m pytest -o addopts='' --basetemp=.pytest-tmp tests/unit/test_prefer_write_tool.py -q`

Result: `32 passed, 1 warning`.

The warning is the local pytest 7.4.4 environment reporting the repository's `asyncio_default_fixture_loop_scope` option as unknown. The default configured command initially could not collect because the environment lacked `textual`; after installing project dependencies, the focused suite ran successfully. The full suite was not run.

## Scope
Only the plugin, its focused unit tests, and the required dev-blog are touched.

## Checklist
- [x] This PR ships its dev-blog under `docs/blog/`
- [x] My changes generate no new test failures
- [ ] Full test suite run (not run; see Testing)

## Additional Notes
Commit: `dd57b31a`

### DIFF
--- a/connectonion/useful_plugins/prefer_write_tool.py
+++ b/connectonion/useful_plugins/prefer_write_tool.py
@@ -37,11 +37,8 @@ if TYPE_CHECKING:
 FILE_CREATION_PATTERNS = [
     re.compile(r"cat\s+<<"),              # cat <<EOF, cat <<'EOF', cat << 'EOF'
     re.compile(r">\s*\S+\.\w+\s*<<"),     # > file.py <<EOF
-    re.compile(r"echo\s+.*[^2]>\s*\S+"),  # echo "..." > file (but not 2>)
-    re.compile(r"printf\s+.*[^2]>\s*\S+"),# printf "..." > file (but not 2>)
-    re.compile(r"tee\s+\S+"),             # tee file.py
-    re.compile(r"(?<!\d)>\s*[~\.]"),      # > ./file, > ~/file (not 2>/dev/null)
-    re.compile(r"(?<!\d)>>\s*[~\.]"),     # >> ./file, >> ~/file
+    re.compile(r"echo\s+.*(?<![2>])>(?!>)\s*\S+"),   # echo "..." > file (but not 2> or >>)
+    re.compile(r"printf\s+.*(?<![2>])>(?!>)\s*\S+"), # printf "..." > file (but not 2> or >>)
 ]
 
 # Patterns that indicate bash is being used to read files (standalone, not in pipelines)
--- a/docs/blog/2026-08-15-when-logging-looked-like-authoring.md
+++ b/docs/blog/2026-08-15-when-logging-looked-like-authoring.md
@@ -1,0 +1,9 @@
+# When logging looked like authoring
+
+The failure was easy to miss because nothing crashed. An agent tried to append a JSONL record with `>>`, then tried `tee`, then tried another shell spelling. The write policy rejected every spelling, the step kept retrying, and the pipeline eventually exited successfully with an empty ledger.
+
+The rule had been written for a useful distinction: model-authored source should go through the Write or Edit tools so a reviewer can see a diff. But append-only state and command output are different operations. Replacing an entire ledger is not equivalent to appending one row, especially when two workers may write at the same time. Likewise, redirecting a subprocess's stdout is logging, not the model authoring a file.
+
+The fix narrows the block to the dangerous case: `echo` and `printf` with a single overwrite redirect, plus heredocs. Append redirects, `tee`, and ordinary command-output redirects are now allowed. The regression tests make the boundary explicit, including `>>` next to the existing `2>` stderr exemption.
+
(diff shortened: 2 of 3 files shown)


wu-changxing (approved):
Reviewed `dd57b31`. No code blocker found.

I ran `python -m pytest -q tests/unit/test_prefer_write_tool.py`: **32 passed**. The revised lookarounds preserve single-`>` blocking for `echo`/`printf`, while allowing `>>`, `tee`, and ordinary command-output redirects as #1028 requires.

This is the corrected version of the duplicate work in #1031: that branch still classifies `echo hello >> file.txt` as file creation. Keep one implementation, not both. GitHub currently reports no checks for this branch, so the full required CI still needs to run before merge.

wu-changxing:
The implementation review is approved, but the PR is currently not mergeable against `main`, and all fork workflows remain `action_required`. Please rebase onto current `main` (1.8.0a1), resolve the conflict without taking the duplicate #1031 implementation, then have a maintainer approve and run Tests, CodeQL, and blog-gate before merge.

wu-changxing:
Closing as superseded by the already-merged canonical fix in #1062 (commit 16d8d541). The release/1.7 line already allows append redirects, tee, ordinary command-output redirects, and stderr redirects while continuing to block heredocs and model-authored echo/printf overwrite creation. The accepted 1.7.0b4 artifact includes that implementation. Thank you for the focused regression coverage and clear write-up.

PR #1039, opened 2026-08-15
Honor caller timeouts for long-running shell commands

### LINKED ISSUE #1037: bash() silently clamps timeout to 600s and returns success on timeout — agents calling agents cannot exceed 10 minutes and cannot tell they were cut
## Summary

`bash()` silently clamps `timeout` to 600s and, on timeout, **returns a string
instead of raising**. An agent orchestrating other `co ai` runs therefore cannot
run any step longer than 10 minutes, is never told its parameter was ignored,
and marks the killed step as **complete**.

## The clamp is silent

`useful_tools/bash.py:49` (and `shell.py:63`):

```python
# Cap timeout at 10 minutes
timeout = min(timeout, 600)
```

Passing `timeout=7200` yields 600. Nothing in the return value says so. Measured
2026-08-15 across one pipeline run — four separate calls, each passing 7200:

```
16:56:57  bash(timeout=7200, ...)  600.0s
17:09:48  bash(timeout=7200, ...)  600.0s
17:20:40  bash(timeout=7200, ...)  600.0s
17:32:04  bash(timeout=7200, ...)  600.0s
```

Four steps, all stopping at exactly 600.0s. I had recorded "timeout=7200 is now
being honoured" as a passing test, because the parameter appeared in the log.
**The parameter was accepted and discarded.**

## A timeout returns success

```python
except subprocess.TimeoutExpired:
    return f"Error: Command timed out after {timeout} seconds"
```

It **returns** — so the tool call succeeds, the string is just tool output, and
the calling agent's own step is marked done. In our runs the orchestrator ticked
"Step 3: engage ✓" on a step that had been killed mid-work, and the whole
pipeline exited **0** with three of five steps truncated. From outside, a
truncated run and a clean run are identical.

## Why 600s is the wrong ceiling for agents calling agents

`co call` and `co ai` make bash the transport for *whole agent runs*, not shell
one-liners. A browser-driving sub-agent takes 10–12 minutes of real work — page
loads, a single-threaded browser daemon, per-item judgement. It is not a hung
process; it is the job.

Concretely: one of our sub-skills writes its output file at around minute 11.
Run alone it finishes in 63 iterations and produces 6 files. Run through the
orchestrator it is killed at minute 10 and produces **zero**, every time,
with no error surfaced anywhere. The 600s cap is the entire difference, and it
cost several full-priced runs before we found it.

## Requested changes

1. **Do not clamp silently.** Either honour the caller's value, or clamp and say
   so in the returned output: `"note: timeout reduced from 7200 to 600 (max)"`.
   A parameter that is accepted and ignored is worse than one that is rejected.
2. **Raise the cap, or make it configurable.** 600s predates agents invoking
   agents over bash. A ceiling in `.co/host.yaml` (per host, per whitelist entry)
   would fit the existing model.
3. **Make a timeout distinguishable from success.** Raising `TimeoutExpired`, or
   returning a structured result the caller can branch on, so an orchestrator
   cannot mark a killed step complete. Today the only signal is prose inside the
   output string.

Environment: connectonion 1.6.6 CLI, `useful_tools/bash.py` and `shell.py`
identical on this point.

wu-changxing:
## Summary

- honor the caller-supplied timeout in `bash()`, `Shell.run()`, and `Shell.run_in_dir()` instead of silently capping it at 600 seconds
- let `subprocess.TimeoutExpired` propagate so agent tool execution and remote EXEC report an error status instead of successful prose output
- add regression coverage for 7,200-second timeouts and timeout exceptions across all three entry points

## Why

The shell tools accepted values such as `timeout=7200` but replaced them with 600 without telling the caller. When the command reached that hidden ceiling, they caught `TimeoutExpired` and returned an error-looking string as a successful tool result. Long-running agents were therefore killed after ten minutes while orchestrators recorded the step as complete.

## Impact

Agents invoking agents can now run beyond ten minutes when the caller explicitly allows it. If a real timeout is reached, the existing tool executor turns the exception into an error trace and remote EXEC exposes it as a failed result, so callers can distinguish truncation from completion.

## Release

Planned for the ConnectOnion 1.6.7 stable release. The stable release branch should carry this fix from `v1.6.6` when 1.6.7 is cut.

## Checks

- `pytest -q tests/unit/test_shell_tool.py tests/unit/test_ws_exec.py` — 48 passed, 1 skipped (Windows-only)
- `ruff check connectonion/useful_tools/bash.py connectonion/useful_tools/shell.py`
- `git diff --check`

Fixes #1037

### DIFF
--- a/connectonion/useful_tools/bash.py
+++ b/connectonion/useful_tools/bash.py
@@ -5,8 +5,8 @@ LLM-Note:
   Data flow: receives command: str, description: str, cwd: str, timeout: int → subprocess.run() with shell=True → captures stdout+stderr → truncates if >10000 chars → returns formatted output: str
   State/Effects: executes system commands via subprocess.run(shell=True) | no persistent state | reads/writes filesystem based on command | can have any side effect depending on command (network calls, file operations, etc.)
   Integration: exposes bash(command, description="", cwd, timeout) function | used as agent tool | description is OPTIONAL (defaults "") — an LLM is prompted to fill it for the approval UI, but direct/programmatic callers (remote.call, co call, scripts) can pass command alone | not passed to shell | Unix/Mac only (raises ValueError on Windows)
-  Performance: timeout default 120s, max 600s | truncates output >10000 chars to prevent token overflow | synchronous execution (blocks until command completes)
-  Errors: raises ValueError on Windows | returns formatted error on timeout | non-zero exit codes included in output | stderr merged with stdout
+  Performance: timeout default 120s | truncates output >10000 chars to prevent token overflow | synchronous execution (blocks until command completes)
+  Errors: raises ValueError on Windows | raises subprocess.TimeoutExpired on timeout | non-zero exit codes included in output | stderr merged with stdout
 
 Bash tool for executing terminal commands (Unix/Mac only).
 
@@ -23,8 +23,8 @@ Usage:
 Note: This tool is for Unix/Mac systems. For cross-platform usage, use Shell class instead.
 """
 
-import subprocess
 import platform
+import subprocess
 
 
 def bash(command: str, description: str = "", cwd: str = ".", timeout: int = 120) -> str:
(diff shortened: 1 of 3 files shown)


wu-changxing:
Superseded by #1042, which is merged and includes the timeout error-classification regression coverage and required design-journal entry. Closing this duplicate to keep the stable release path unambiguous.

PR #1050, opened 2026-08-15
fix(bash,shell): honour the caller's timeout, and make a timeout unmistakable (#1037)

### LINKED ISSUE #1037: bash() silently clamps timeout to 600s and returns success on timeout — agents calling agents cannot exceed 10 minutes and cannot tell they were cut

wu-changxing:
Closes #1037.

## What was wrong

`bash()` and `Shell.run()` / `run_in_dir()` clamped every call with `timeout = min(timeout, 600)` and reported nothing. `co call` and `co ai` make bash the transport for *whole agent runs*, so a ten-minute ceiling is a ceiling on what an agent can ask another agent to do — and the caller could not tell its parameter had been discarded.

## Evidence

A five-step pipeline passing `timeout=7200` on every call:

```
bash(timeout=7200, ...)  [OK] 600.0s
bash(timeout=7200, ...)  [OK] 600.0s
bash(timeout=7200, ...)  [OK] 600.0s
bash(timeout=7200, ...)  [OK] 600.0s
```

One sub-skill writes its output file at around minute 11. Through the pipeline it produced **zero files, every run**; run standalone it finishes in 63 iterations and writes 6. The clamp also silently swallowed a separate `-i` fix — iteration ceilings really were raised to 200, but nothing could reach one because wall-clock killed the process first. **Two limits, and the invisible one won.**

## Changes

1. **Remove the clamp** in `bash.py` and both `shell.py` sites. The caller's value is used as given; the default stays 120s.
2. **Make a timeout unmistakable.** It previously returned `"Error: Command timed out after N seconds"` — an ordinary successful tool result as far as the caller is concerned, which is how orchestrators came to mark killed steps complete while the run exited 0. It now leads with `TIMEOUT:` and states plainly that the work did not complete.

Kept as a returned string rather than a raised exception: the existing tests pin the return contract, and other failure paths in this tool (Windows, missing bash) return strings too.

## Tests

Two regression tests assert the timeout reaches `subprocess.run` unchanged. **Both fail on the unpatched code** (`assert 600 == 7200`) — verified by stashing the fix and running them — and 25 pass with it.

Existing timeout assertions were updated to the new wording.

### DIFF
--- a/connectonion/useful_tools/bash.py
+++ b/connectonion/useful_tools/bash.py
@@ -5,7 +5,7 @@ LLM-Note:
   Data flow: receives command: str, description: str, cwd: str, timeout: int → subprocess.run() with shell=True → captures stdout+stderr → truncates if >10000 chars → returns formatted output: str
   State/Effects: executes system commands via subprocess.run(shell=True) | no persistent state | reads/writes filesystem based on command | can have any side effect depending on command (network calls, file operations, etc.)
   Integration: exposes bash(command, description="", cwd, timeout) function | used as agent tool | description is OPTIONAL (defaults "") — an LLM is prompted to fill it for the approval UI, but direct/programmatic callers (remote.call, co call, scripts) can pass command alone | not passed to shell | Unix/Mac only (raises ValueError on Windows)
-  Performance: timeout default 120s, max 600s | truncates output >10000 chars to prevent token overflow | synchronous execution (blocks until command completes)
+  Performance: timeout default 120s, caller's value honoured | truncates output >10000 chars to prevent token overflow | synchronous execution (blocks until command completes)
   Errors: raises ValueError on Windows | returns formatted error on timeout | non-zero exit codes included in output | stderr merged with stdout
 
 Bash tool for executing terminal commands (Unix/Mac only).
@@ -36,7 +36,10 @@ def bash(command: str, description: str = "", cwd: str = ".", timeout: int = 120
             — an LLM should fill it in so the approval UI can show intent, but direct
             callers (scripts, remote.call, co call) may pass just the command.
         cwd: Working directory (default: current directory)
-        timeout: Seconds before timeout (default: 120, max: 600)
+        timeout: Seconds before timeout (default: 120). Honoured as given —
+            an agent invoking another agent over bash routinely needs more than
+            ten minutes, and a silently clamped value is indistinguishable from
+            a respected one.
 
     Returns:
         Command output (stdout + stderr)
@@ -45,8 +48,6 @@ def bash(command: str, description: str = "", cwd: str = ".", timeout: int = 120
     if platform.system() == "Windows":
         return "Error: bash tool is for Unix/Mac only. Use Shell class for Windows."
 
-    # Cap timeout at 10 minutes
-    timeout = min(timeout, 600)
 
     try:
         result = subprocess.run(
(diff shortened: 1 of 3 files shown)


wu-changxing:
Review of `e4ecdde4`: not merge-ready yet.

The silent 600-second clamp is a real bug, but deleting the ceiling entirely turns an agent-controlled tool argument into an unbounded synchronous worker lease. A model can now ask for `timeout=2147483647` and monopolize the tool executor effectively forever. Please keep the caller-visible behavior honest while retaining an operator-controlled upper bound (or split a trusted long-running transport from the model-facing tool), and reject out-of-policy values explicitly instead of silently clamping them.

The second claimed fix is also still only prose inside a successful string result. `TIMEOUT:` is easier for a model to notice, but the tool protocol still reports an ordinary successful return, so a non-LLM orchestrator can continue marking the step complete. Please expose timeout as a machine-readable error/typed result, or teach the tool executor to mark this outcome as an error while preserving the display text.

Coverage also needs the third changed path, `Shell.run_in_dir()`. The branch currently conflicts with `main` and has no CI result, so those gates remain after the contract is fixed.

wu-changxing:
Update required before merge. This fix is release-relevant and is now tracked by #1154, but GitHub reports the branch as not mergeable and no workflow runs are attached to the current head. Please rebase onto current `main`, preserve the focused timeout regressions, and run the complete required CI matrix.

wu-changxing:
Closing as superseded by #1042 (commit 26255940), which is already present on release/1.7 and goes further by honoring caller timeouts and propagating subprocess.TimeoutExpired so callers can distinguish timeout from a successful tool result. The accepted 1.7.0b4 artifact includes the canonical implementation and cross-platform matrix.

Issue #1056, opened 2026-08-15
[1.6.9 production blocker] Implement, verify, and merge the stable patch train


wu-changxing:
## Objective

Ship ConnectOnion **1.6.9** as a stable production patch from the `v1.6.8` lineage.

This issue owns implementation, PR sequencing, CI evidence, merging, and the final release candidate. The audited scope and rationale live in #1051.

## Hard branch boundary

- Branch from `v1.6.8` / `release/1.6.8`, never from current `main`.
- Do not merge `main` into the stable branch.
- Backport individual fixes only.
- Do not import ACP, OIP preview, provider Workroom/session UI, permission-vocabulary migration, or other 1.7 architecture.
- Each logical fix should be independently reviewable and revertible.

### PR 1 — Contact usability and permission model

Implements #1054.

- [ ] Keep three product roles: stranger/blocked, contact, admin.
- [ ] A contact can use normal COAI execution, edit/run tools, answer approval requests for its own authenticated session, and delegate to Codex/Claude Code.
- [ ] Admin adds only control-plane capabilities: trust/admin mutation, protected host configuration, deployment management, privileged logs/session inspection, and equivalent owner operations.
- [ ] Contacts cannot promote themselves, modify another user, access another user's session, or bypass protected policy.
- [ ] Replace tests that encode “non-admin can never approve” with capability/session-ownership tests.
- [ ] Add the complete fresh-invite user journey and reconnect/restart coverage.
- [ ] No onboarding step requires editing YAML.

### PR 2 — Bind deploy account to the actual remote identity

Implements #1055 and completes #995.

- [ ] Discover the actual key retained by an existing server.
- [ ] Authenticate account metadata as that identity, including a server-only `--own-identity` key.
- [ ] Never combine remote identity A with derived account/mailbox/token B.
- [ ] If the actual identity cannot be authenticated, fail safely or omit account metadata with an actionable diagnostic.
- [ ] Cover fresh derived identity, legacy/rotated identity, and `--own-identity`.
- [ ] Verify behavior from an installed wheel.

### PR 3 — Prevent `co deploy --to` runtime-state loss

Implements #911.

- [ ] Make the server deploy path honor the project's ignore/protected-state contract.
- [ ] Preserve intentional deletion of removed source code.
- [ ] Preserve ignored/server-only runtime data and prevent stale local state from overwriting it.
- [ ] Preserve framework secrets and identity state.
[...]

Issue #1051, opened 2026-08-15
[1.6.9] Stable patch gate: audit, backport, and verify without merging the 1.7 alpha


wu-changxing:
## Goal

Cut a deliberately boring, stable **1.6.9** from `v1.6.8`. Do not merge current `main` into the stable line: `main` carries the 1.7 alpha protocol/runtime work (including ACP code that is now being removed).

## Verified baseline

Audit performed against `v1.6.8` (`34a7e1a`) and current `main` (`64b4f7b`).

Already present in the v1.6.8 artifact and not to be reimplemented:

- fresh `co ai` owner invite onboarding;
- received-inbox pagination;
- caller-supplied shell timeouts are honored and timeout exceptions are surfaced;
- normal `co ai` evaluation is opt-in;
- payment cannot be granted from a self-declared request field (#1014 is fixed in code and regression-tested, although the issue is still open);
- deployer/admin seeding and remote EXEC requester checks;
- the first-deploy portion of the agent-account fix from #995.

Important correction: #995 is **not fully solved**. Its post-merge review found that an existing remote/server-only identity can still be paired with the locally derived account metadata. This is now #1055 and blocks 1.6.9.

The release notes for v1.6.8 mention only the CI-unblock PR, so code/tag evidence—not the release body—must be the source of truth.

Focused v1.6.8 audit: 37 tests passed across onboarding, deployed identity, requester-aware EXEC, and approval authority. Those permission tests encode the current unusable-contact policy, so their passing result is not product acceptance.

### P0 — release blockers

- [ ] **#1054 — invited contacts are usable.** A contact can perform ordinary work, select normal execution modes, respond to approvals in its own session, and delegate to coding tools. Admin remains required only for control-plane settings, deployment/host management, privileged logs, trust/admin mutation, and cross-user inspection.
- [ ] **#1055 — deploy identity/account consistency.** Bind token, mailbox, address, and billing to the actual key retained by the server, including legacy, rotated, and `--own-identity` paths.
- [ ] **#911 — `co deploy --to` must not silently destroy runtime state.** Make the server deploy path honor the project's ignore contract (or an equally explicit protected-state contract), add deletion/overwrite regression tests, and document the boundary. Preserve intentional deletion of deployed source without treating arbitrary runtime data as source.
[...]

PR #1067, opened 2026-08-15
fix(skills): honor Claude allowed-tools metadata

### LINKED ISSUE #967: Claude Code skills load, but their `allowed-tools` permissions are silently dropped
We read skills from `.claude/skills/<name>/SKILL.md` and `~/.claude/skills/…` at runtime (`useful_plugins/skills.py:117-136`, tagged `claude-project` / `claude-user` at `:207-212`). That is a genuinely strong property: a skill someone already wrote in Claude Code runs here with no porting.

But permissions are read only from the `tools:` key (`skills.py:346,393`). Claude Code writes that key as `allowed-tools`. Grepping `connectonion/` for `allowed-tools` returns nothing.

So a Claude Code skill loads and its instructions run unchanged, but its permission grants are silently dropped. The skill's own auto-approvals do not apply, and every gated tool call falls back to prompting the user. The skill still works — it is just more annoying than it was in the tool it came from, and nothing says why.

Silent is the problem. Someone whose skill suddenly asks for approval on every `git` call will assume our approval system is broken rather than that a frontmatter key was ignored.

Two things worth doing, in order of value:

1. **Read `allowed-tools` as an alias for `tools`.** The formats look compatible — our `Bash(git *)` becomes `when: {'command': 'git *'}` (`skills.py:246-287`), and Claude Code uses the same `Bash(git *)` shape. Worth confirming the edge cases match before treating them as identical.
2. **If we deliberately do not honour it**, warn once on load: "skill `x` declares allowed-tools, which ConnectOnion does not read; add a `tools:` key to grant permissions." Silently ignoring a permission declaration is the worst of the three options.

Until this is resolved, the honest public wording is **"reads the same SKILL.md files"** — never "runs identically". That distinction is recorded in `docs/SELLING_POINTS.md`.

Related: #965, #966 — same audit.

wu-changxing:
## Outcome

Claude Code skills loaded from `.claude/skills` now keep their declared permissions instead of silently prompting for every gated tool.

## Compatibility contract

- native `tools:` remains authoritative when both spellings exist
- `allowed-tools:` accepts YAML lists and Claude's comma-separated scalar form
- title-cased Claude tool names such as `Edit` match the same lower-case ConnectOnion tool
- `Bash(git *)` keeps the existing command-pattern semantics
- malformed YAML never grants permissions from line-based recovery
- imported permissions remain turn-scoped and are restored after success or failure

## Verification

127 focused skill parsing, permission lifecycle, doctor, sync, and approval tests passed with seed 967. Python compilation and `git diff --check` passed. Documentation explains precedence and accepted forms.

Fixes #967
Tracks #1051 and #1056.

### DIFF
--- a/connectonion/useful_plugins/skills.py
+++ b/connectonion/useful_plugins/skills.py
@@ -215,7 +215,7 @@ def _parse_skill_content(content: str) -> tuple[Dict[str, Any], str]:
 
 # The only keys worth rescuing from a frontmatter YAML refuses to parse.
 #
-# Deliberately not `tools:`. test_one_reader_for_skill_frontmatter records why
+# Deliberately not `tools:` or `allowed-tools:`. The frontmatter tests record why
 # the strict reader won: "neither invents a reading of a file that has a syntax
 # error in it", and `tools:` is fed to _grant_skill_permissions — guessing it
 # from a file that does not parse would widen an agent's permissions on the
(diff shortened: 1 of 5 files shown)


wu-changxing:
### Stable merge blocker: Claude compatibility is not fail-closed yet

The current alias implementation should not merge into 1.6.9 as written.

1. The parser does not accept Claude Code's documented/common scalar form. Anthropic's own example is space-delimited:

`allowed-tools: Bash(git add *) Bash(git commit *) Bash(git status *)`

This patch only splits commas, so that declaration becomes one impossible pattern and permissions are still silently dropped while the new docs claim compatibility.

2. `Read`, `Edit`, and `Glob` are not merely title-cased ConnectOnion tool names. The stable runtime uses names such as `read_file` and `write`. Making all permission matching globally case-insensitive does not map those vocabularies, and it changes native `tools:` semantics outside the alias.

3. Claude Code applies project-skill grants after workspace trust. This patch imports a previously inert foreign permission field into the turn auto-approval path without an equivalent trust gate or source-specific restriction. The regression only uses already-native `read_file`, so it does not exercise the advertised Claude vocabulary or the security boundary.

For 1.6.9, please either implement a strict, source-scoped parser/mapping with malformed and unknown entries failing closed plus source-authority coverage, or take #1056's explicit-warning path: keep `allowed-tools` inert, warn once when invoked, and report it in `co doctor`.

The warning path is safer for the stable patch. I attempted to submit this as REQUEST_CHANGES, but GitHub disallows requesting changes on one's own PR; this comment records the blocker explicitly.

wu-changxing:
Superseded by merged stable PR #1070. The direct alias remains unsafe for 1.6.9 because Claude's scalar grammar, tool vocabulary, and workspace-trust boundary are not equivalent to the stable ConnectOnion permission grammar. #1070 takes #1056's explicit-warning path and keeps the foreign declaration fail-closed.

PR #1071, opened 2026-08-15
fix: make ignored Claude skill permissions explicit

### LINKED ISSUE #967: Claude Code skills load, but their `allowed-tools` permissions are silently dropped

wu-changxing:
Closes #967. Keeps the stable permission boundary unchanged: Claude Code allowed-tools is not treated as ConnectOnion tools because their tool vocabularies and authority rules are not identical. Instead, co doctor reports every affected skill and invocation emits one warning per skill/session with the exact tools: migration. This avoids both silent failure and silent permission widening. Focused result: 91 skill/frontmatter/subscription tests passed. Docs now state the compatibility boundary.

### DIFF
--- a/connectonion/useful_plugins/skills.py
+++ b/connectonion/useful_plugins/skills.py
@@ -421,8 +421,14 @@ def _why_the_skill_cannot_be_read(skill_md: Path) -> Optional[str]:
     if not match:
         return None
 
-    import yaml
     yaml_text = match.group(1)
+    top_level_keys = {
+        line.split(':', 1)[0].strip()
+        for line in yaml_text.splitlines()
+        if ':' in line and not line.startswith((' ', '\t'))
+    }
+
+    import yaml
     try:
         yaml.safe_load(yaml_text)
     except yaml.YAMLError as e:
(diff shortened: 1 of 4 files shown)


wu-changxing:
Superseded by merged PR #1070, which preserves the same fail-closed boundary, warns once per skill path at load time, reports valid and malformed allowed-tools declarations in co doctor, and gives manifest errors priority. I reviewed the merged diff and will use #1070 as the 1.6.9 source of truth.

PR #1084, opened 2026-08-16
release: 1.6.10 fixes CJK browser typing

### LINKED ISSUE #1078: keyboard_type crashes with NameError: _active_text_len when typing CJK text (1.6.8)
## Reproduction

Against a live page with a focused input, on connectonion 1.6.8:

```
co browser keyboard_type "DRAFT SYNC TEST ignore"     # ✅ works
co browser keyboard_type "【草稿同步测试，请忽略】"        # ❌ NameError: name '_active_text_len' is not defined
co browser type_text_by_selector "div[contenteditable]" "【…】"   # ❌ same NameError
```

Both typing paths crash identically on CJK input. Notably the **first character
lands before the crash** — the composer was left containing `【` followed by
nothing — so the failure is mid-stream in the typing instrumentation, not in
argument handling.

ASCII typing through the same daemon, same session, works fine (verified
seconds apart). Overnight automation typed dozens of ASCII comments without
issue; the crash appeared the first time a Chinese string went through.

## Impact

Any automation that types Chinese (or presumably any non-ASCII) text through
the daemon dies mid-word and leaves a half-typed field behind — which for a
message composer is one Enter away from sending a fragment. For a bilingual
user this makes `keyboard_type` unusable for half their content.

`_active_text_len` looks like a variable in the human-jitter/typing pacing path
that is only assigned on some code path ASCII takes and CJK does not.

Environment: connectonion 1.6.8, patchright 1.61.2, macOS, real Chrome binary.

wu-changxing:
## Outcome

Fixes #1078 and prepares stable 1.6.10.

The humanized CJK paste path called `_active_text_len()` after a clipboard refactor accidentally removed its definition. This restores the focused input/contenteditable probe, adds a regression test for the complete CJK paste path, and updates the stable version metadata.

## Evidence

- Red reproduction: the new test failed with `NameError: _active_text_len`.
- `tests/unit/test_browser_humanize.py`: 15 passed.
- Release/version tests: 53 passed after lock refresh; the only local cross-repo mismatch is handled by the coordinated docs PR.
- Stable suite: 6,215 passed; all 25 sandbox-denied socket/process failures passed when rerun with OS access. The remaining local-only check points at the intentionally not-yet-merged docs stable version.
- Built `connectonion-1.6.10` wheel and sdist; both pass `twine check`.
- Installed-wheel acceptance: 8 passed.
- Real headless Chrome from the built wheel entered `【草稿同步测试，请忽略】` exactly and captured `/tmp/frontend-test-screenshots/connectonion-1.6.10-cjk-input.png`.

This is a small stable maintenance fix, so release notes are sufficient and no design-journal entry is needed.


wu-changxing:
v1.6.10 is now publishing from the exact reviewed commit. Because main is already on the 1.7 preview train, this branch's stable version files conflict by design. The code/test fix is ported without version downgrade in #1085, which supersedes this PR for the main merge.

PR #1133, opened 2026-08-17
docs(useful_skills): add deploy-agent — production agent deployment guide

wu-changxing:
Replaces #1130.

### Why a new branch rather than a fix-up commit

The customer name appears in #1130's **commit message and diff**. A later commit removing it from the file would leave both publicly readable in that branch's history. This is a fresh branch off current `main` with a single clean commit, so nothing identifying exists anywhere in it.

**#1130 should be closed without merging, and its branch deleted.**

### What was removed

| Removed | Why |
| --- | --- |
| Customer name in the opening paragraph | named under NDA |
| "a contract-ledger agent and a candidate-mapping agent" | describes their systems |
| "non-technical Chinese client" | narrows identity |
| `NW_WORK=` | env prefix carries their initials |
| "Feishu Base" as *the* stack | now one example among Airtable/Notion/Sheets |
| `fde-client-agent` naming | "client" framing throughout — file, title, slash command |

Even with the name deleted, "contract-ledger + candidate-mapping agents, Chinese client, Feishu" identifies them to anyone in the space. All of it is gone.

### What was kept

The engineering content, which is genuinely general once de-identified — state outside the rsync root, incremental diffing with normalized values, row identity by stable token, user-owned columns never overwritten, verifying billing identity against the live endpoint rather than config, and the sandboxed-iframe constraint.

The $1.97 / 1.28M-token anecdote is kept: it is instructive and identifies nobody.

Reframed from a client-engagement case study into general deployment guidance, and renamed `deploy-agent` (`/deploy-agent`).


wu-changxing:
Update required before merge. Tests and CodeQL pass, but blog-gate fails on the current head. Please add or adjust the required design-journal entry, then rerun the gate. Because this replaces the NDA-sensitive #1130, keep the clean one-commit history and do not reintroduce the old branch through a merge/rebase that carries its commits.

wu-changxing:
Rebased and unblocked, still one docs commit plus the blog post the gate asked for.

- The branch had no common ancestor with today's `main` (its parent commit is only reachable from the `backport/1.7-*` branches now), so GitHub could not compute mergeability. The single commit is cherry-picked onto current `main` (`00388ea`, same author, same two files). Nothing from the earlier NDA branch is carried.
- Added `docs/blog/2026-09-05-every-rule-was-paid-for.md` for `blog-gate`. It tells the starved-service and stale-snapshot incidents the way the skill does, with no customer, system, or stack named.
- Skill-doc tests (`test_default_skills_are_for_customers`, `test_the_ship_skill_matches_this_repo`, and neighbours): 48 passed on the rebased tree.

One thing this PR does not do, same as before: `deploy-agent` is a docs page only. The README table calls these "skills you can copy", but there is no `connectonion/useful_skills/deploy-agent/SKILL.md` and no `co copy` registry entry, so `/deploy-agent` is not yet invocable. `cli-skill-design` is in the same state, so this is a pre-existing pattern rather than a regression; tracking it separately would be better than widening this PR.

Merging once CI is green on the new head.

---

Issue #1150, opened 2026-08-20
A run that hits max-iterations exits 0, and there is no cost ceiling


wu-changxing:
## Summary

An agent run that exhausts `--max-iterations` prints `Task incomplete` and **exits 0**, so an orchestrator cannot tell a finished task from an abandoned one. Combined with there being no cost ceiling and no in-context budget signal, a scheduled fleet can bill a large amount for work that never completed, and every layer above it reports success.

Measured on 2026-08-20 running the LinkedIn automation on `co/gemini-3.7-flash`.

## 1. `max_iterations` exits 0

```
$ co ai -m co/gemini-3.7-flash -i 1 "List every file in the current directory, then read three of them, then summarise."
...
Task incomplete: Maximum iterations (1) reached.
$ echo $?
0
```

`core/agent.py:588` returns the outcome tagged `'max_iterations'`, correctly distinct from `'natural'`. But that distinction is lost by the time it reaches a caller: `cli/co_ai/acp_server.py:425` and `:1278` group `max_iterations` together with `natural` as a normal end state, and the process exits 0.

The consequence for anything scripted: our orchestrator runs four agent steps in sequence and logs `exit 0` for each. Two of today's runs ended on the iteration cap having completed none of the requested work, and the log is indistinguishable from a clean run. **A shell orchestrator has no way to detect a truncated agent run today** short of grepping stdout for an English sentence.

Suggested: exit non-zero (or at minimum expose the outcome tag in a machine-readable way) when the run ends on `max_iterations`. Hitting the cap is a failure of the run, not a completion of it.

## 2. Cost is uncapped and invisible to the agent

Three runs today, same skill, same model, only `-i` differing:

| `-i` | tokens | cost | outcome |
|---|---|---|---|
| 25 | 1.13M | $1.73 | max_iterations |
| 60 | 5.10M | $7.77 | max_iterations |
| 100 | 8.23M | $12.51 | max_iterations |

Cost scales linearly with the iteration cap because **~95% of every request is resent context**:

```
106.5k tok (101628 cached)
106.7k tok (101621 cached)
107.0k tok (101613 cached)
```

101.6k of 106.5k cached, re-sent every single iteration. At 100 iterations that is the same ~100k tokens paid for a hundred times. (Cached tokens are currently billed at full price — that part is tracked separately in oo-api#162 — but even priced correctly, the shape of the problem is the same: the iteration cap is a proxy for cost, and a bad one, because per-iteration cost grows as the context does.)

[...]

PR #1160, opened 2026-08-21
fix: report max-iteration runs as incomplete

wu-changxing:
Closes the stable-1.7 portion of #1150. Cost ceilings and in-context cost guidance remain explicitly deferred to 1.8.

## Contract

- a natural one-shot turn exits 0
- a one-shot turn ending at `max_iterations` preserves and prints its partial result, but exits 1
- JSON output always includes `outcome`: `natural`, `max_iterations`, or `error`
- resumable state is committed before an incomplete JSON run exits
- library `Agent.input()` keeps its existing string return API; the CLI reads the already-recorded canonical `turn_result` event
- missing/malformed terminal outcome is an invariant failure, not silently converted to success

## Examples

Natural completion:

```json
{"session_id":"...","result":"done","outcome":"natural","error":null}
```

Iteration cap:

```json
{"session_id":"...","result":"Task incomplete: Maximum iterations (1) reached.","outcome":"max_iterations","error":null}
```

The second envelope exits 1.

## Verification

- focused CLI / resumable-session / agent-server tests: 67 passed
- Ruff on changed Python files: passed
- `git diff --check`: passed
- regression coverage for both plain and JSON max-iteration outcomes

Release control: #792

### DIFF
--- a/connectonion/cli/commands/ai_commands.py
+++ b/connectonion/cli/commands/ai_commands.py
@@ -52,7 +52,7 @@ def handle_ai(
     if not prompt and (json_output or resume):
         message = "--json and --resume require a one-shot prompt"
         if json_output:
-            _print_envelope(None, None, message)
+            _print_envelope(None, None, "error", message)
         else:
             console.print(f"[red]{message}[/red]")
         raise typer.Exit(2)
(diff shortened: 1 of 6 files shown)


wu-changxing:
CI follow-up at `c3cd7136`: all four Linux jobs exposed the same missed in-repo consumer, `test_json_review_uses_the_333_agent_factory_seam`. Its fake Agent completed without the canonical `turn_result`, and its exact JSON assertion still expected the pre-outcome envelope. The fake now records `reason: natural` and expects `outcome: natural`. Local rerun: 85 GitHub Action / one-shot / CLI tests passed; scoped Ruff and diff checks passed. Full CI is rerunning.

PR #1200, opened 2026-08-22
fix(browser): isolate profiles without replacing home

### LINKED ISSUE #1199: [1.7 release] Isolate co browser profiles without replacing HOME
## Problem

Release automation needs a fresh `co browser` identity without sharing or mutating a user's persistent profile. Today the profile is fixed under `~/.co/browser_profile`, so callers can isolate it only by replacing `HOME`.

On macOS, a real Chrome process launched with a synthetic temporary `HOME` starts but cannot navigate even to a healthy localhost server: `Page.goto` times out, the authoritative URL remains `about:blank`, and the server receives no browser request. `curl` against the same server succeeds immediately.

## Required behavior

- support an explicit, absolute browser profile directory independently from `HOME`;
- retain `~/.co/browser_profile` as the default;
- keep socket isolation independent through the existing `CO_BROWSER_SOCK`;
- document the safe isolated automation combination;
- cover default and override resolution in unit tests;
- prove a real isolated `co browser` can navigate localhost without replacing `HOME`.

## 1.7 release relevance

This blocks the real temporary-invite + isolated Co-browser Beta gate tracked by #792 and openonion/oo-chat#215.

wu-changxing:
## Outcome

Add `CO_BROWSER_PROFILE_DIR` so an isolated `co browser` run can use its own Chrome identity and daemon socket while retaining the real operating-system `HOME`. The default remains `~/.co/browser_profile`. Text-entry commands also accept a final `--stdin`, keeping passwords and one-run invite codes out of process arguments.

This fixes the macOS localhost failure found by the 1.7 release gate: both headless and headed Chrome started under a synthetic temporary `HOME`, but never sent a request and stayed at `about:blank`. With this change, the same exact candidate, isolated profile, and isolated socket navigated a local server successfully and returned the exact authoritative URL.

## Verification

```text
pytest -q -c /dev/null tests/unit/test_browser_automation.py tests/unit/test_browser_session_pages.py tests/e2e/cli/test_browser_daemon.py
125 passed
git diff --check

real candidate smoke:
curl localhost fixture: HTTP 200
co browser isolated profile: Navigated to http://localhost:3119/
get_current_url: http://localhost:3119/
stdin text-entry smoke: DOM reported the exact 14-character length without exposing the value in argv
```

The socket-backed tests were run outside the filesystem/network sandbox. Documentation now shows the safe profile + socket isolation combination and explicitly says not to replace `HOME`.

Fixes #1199. Advances #792 and openonion/oo-chat#215.

### DIFF
--- a/connectonion/cli/commands/browser_commands.py
+++ b/connectonion/cli/commands/browser_commands.py
@@ -26,6 +26,7 @@ USAGE = (
     "  co browser tab close <NAME>              release your tab when the task is done\n"
     "  co browser close                         close the browser and stop the daemon\n"
     "  co browser help                          list every browser function\n"
+    "  printf secret | co browser type_text_by_selector '#field' --stdin\n"
     "\n"
     "One task = one tab. Solo use needs no -t at all. Running several agents on this\n"
     "browser? Each opens its own tab once, adds -t <name> to EVERY command (including\n"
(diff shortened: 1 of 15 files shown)


wu-changxing:
Fresh exact-artifact release gate passed with this Core head (`19f5d95d`), including isolated browser profile, secure stdin invite entry, controlled-input fill, browser/C/Rust/native Codex, Workroom, modes, Stop, and reconnect.

PR #1221, opened 2026-08-23
fix(outlook): prevent scheduled sends from being delivered immediately

### LINKED ISSUE #1198: co outlook send --at delivers immediately AND schedules a duplicate
### What happened

`co outlook send ... --at <utc-iso>` reports `✓ Scheduled`, but the message is **also delivered immediately**. The recipient gets it twice: once right away, and again at the scheduled time.

### Reproduction (2026-08-22)

```
co outlook send adi@example.com "RE: subject" - --cc other@example.com --at 2026-08-22T16:00:00Z < body.txt
✓ Scheduled for 2026-08-22T16:00:00Z to adi@example.com
  From: xietianle@outlook.com
```

Immediately after, at 06:48:40Z (about 9 hours before the scheduled time):

```
co outlook sent -n 1
1. To: adi@example.com
   Subject: RE: subject
   Date: 2026-08-22T06:48:40Z      <-- already sent
```

and at the same time:

```
co outlook scheduled
2026-08-22T16:00:00Z  adi@example.com  RE: subject   <-- still queued
```

The two are different messages, with different IDs and in different folders (Sent Items vs Drafts):

```
sent:      ...AQkAAAB...lscTJdAAAA
scheduled: ...AQ8AAAB...lscSqKAAAA
```

Reading the Sent Items copy shows the full intended body, so it is a real delivery, not a placeholder.

### Impact

Silent duplicate delivery to an external recipient. Worse, the failure is invisible from the command output — `✓ Scheduled` is the only feedback, and it is wrong. The whole point of `--at` is to have a window in which nothing has gone out yet; this defeats it.

### Workaround

After any `--at` send, check both:

```
co outlook sent -n 1
co outlook scheduled
```

and if a copy already went out, cancel the queued one with `co outlook cancel <#>` (that part works correctly).

### Note

The docs previously said scheduled mail could only be cancelled from the Outlook client because the API always returns 403. `co outlook cancel` worked fine in this test, so that note appears to be out of date.

mikemikimike:
## Description
Prevent scheduled Outlook sends from being delivered immediately or leaving a duplicate deferred copy.

## Related Issue
Fixes #1198

- **Suggested labels:** bug, browser
- **Proposed target version:** future roadmap
- **Estimated release window:** unknown date
- **Why this release:** This is a narrowly scoped correctness fix for a user-visible duplicate-send behavior. It changes only the scheduled-send transport path; immediate sends and threaded replies keep their existing endpoints. The rollback path is a single commit.
- **Forward-port tracking issue (stable patches only):** N/A — this is not targeting a stable patch line.

## How this was written

- Least confident: I could not exercise a live Microsoft Graph mailbox locally, so the exact Exchange delivery lifecycle remains covered by the mocked HTTP contract and the repository CI suite rather than a real account.
- Not verified locally: the locked project test environment could not finish dependency installation on Windows because package downloads stalled; the available global pytest environment has incompatible plugins/configuration, and Ruff is not installed globally.
- If this is wrong, what breaks first? A scheduled send could fail to remain deferred, or the mailbox could show an unexpected draft state; the HTTP method, payload, scope check, and request count assertions should expose that regression before merge.

## Type of Change

## Dev Blog (required)

The required Design Journal post is included in the same diff:

> docs/blog/2026-08-28-a-scheduled-send-was-still-a-send.md

## Changes Made

- Keep immediate sends on Microsoft Graph `/me/sendMail`.
- Create scheduled messages as drafts through `/me/messages` with the deferred-send extended property attached before storage.
- Require `Mail.ReadWrite` before the scheduled draft request.
- Extend the Outlook unit tests to assert the actual endpoint, payload, request count, and missing-scope failure.
- Add the required Design Journal post for the scheduled-send bug.

## Testing

Remote GitHub Actions for commit `d77560fa`:

- `pytest tests/ -v --tb=short -m "not slow and not real_api and not network"` — passed on Python 3.10, 3.11, 3.12, and 3.13.
- `pytest tests/unit/test_transport.py tests/e2e/cli/test_browser_daemon.py -v --tb=short` — passed on Windows with Python 3.10, 3.12, and 3.13.
- Windows wheel/CLI/browser E2E — passed.
- macOS wheel/CLI/browser E2E — passed.
- Lockfile audit, CodeQL, and Python analysis — passed.
- `blog-gate` — passed, including the story-quality check.

Local checks:

- `python -m compileall -q connectonion/useful_tools/outlook.py tests/unit/test_outlook.py` — passed.
- `git diff upstream/main...HEAD --check` — passed.
- `uv run --frozen pytest -q tests/unit/test_outlook.py` — not completed: first-time environment setup stalled while downloading project dependencies.
- Host `python -X utf8 -m pytest -q tests/unit/test_outlook.py` — not completed: the global environment's `langsmith` pytest plugin is incompatible with its installed Pydantic version; disabling auto-loaded plugins then exposed missing `pytest-asyncio` configuration support.
- Ruff — not run locally because no Ruff executable is installed; the remote analysis job passed.

## Scope

- `connectonion/useful_tools/outlook.py`: correct the scheduled-send endpoint and scope guard.
- `tests/unit/test_outlook.py`: verify the regression contract.
- `docs/blog/2026-08-28-a-scheduled-send-was-still-a-send.md`: satisfy the repository's required Design Journal gate with the Issue's actual failure and fix.

## Example Usage

```python
from connectonion.useful_tools.outlook import Outlook

# Outlook.send(..., send_at="2026-08-28T18:00:00Z") creates a deferred draft.
```

## Checklist

- [x] My changes generate no new warnings in the remote CI suite
- [x] This is not a stable patch-line target, so no forward-port tracker is required
- [x] The linked issue's implementation contract is covered by the unit-test assertions and remote CI evidence

Not applicable: this is a mocked HTTP behavior fix.

## Additional Notes

The implementation leaves immediate sends and threaded replies unchanged. The scheduled path now relies on the mailbox's draft lifecycle, so the scheduled item remains available to existing listing behavior while Exchange owns deferred delivery.

Generated-by: Codex


wu-changxing (changes requested):
The current implementation is not safe to merge yet.

`POST /me/messages` creates a draft; it does not submit that draft for delivery. Microsoft Graph documents a separate `POST /me/messages/{id}/send` operation for sending an existing draft:

- https://learn.microsoft.com/en-us/graph/api/user-post-messages?view=graph-rest-1.0
- https://learn.microsoft.com/en-us/graph/api/message-send?view=graph-rest-1.0

Please change the scheduled path to:

1. create the draft with the deferred-send properties;
2. read the draft ID from the `201` response;
3. submit that exact draft through `/me/messages/{id}/send` and require `202`;
4. update the regression test to assert both requests and their lifecycle;
5. validate with a real Outlook mailbox that the message is not delivered immediately, is delivered exactly once at the requested time, and remains cancellable before then.

The mocked test currently proves only that a draft was created, so all-green CI does not establish the user-visible behavior in #1198. This should stay out of 1.7.1 until the live mailbox acceptance is recorded.

mikemikimike:
### Live Outlook validation

Implemented the requested follow-up from this review:

- scheduled sends create a draft with both `SystemTime 0x3FEF` (`PidTagDeferredSendTime`) and `SystemTime 0x000F` (`PidTagDeferredDeliveryTime`);
- the `201 Created` response is parsed for the draft ID;
- that exact draft is submitted with `POST /me/messages/{id}/send`, with no request body, and the client requires Graph's `202 Accepted` response;
- the regression test verifies the two-request lifecycle, payload, draft ID, and empty send body.

Validation performed:

- `python -m pytest tests/unit/test_outlook.py -q --basetemp=... -k send_email_scheduled`: **2 passed**
- `python -m compileall -q connectonion/useful_tools/outlook.py tests/unit/test_outlook.py`: **passed**
- `git diff --check`: **passed**
- Live mailbox: `thomasschultz1841@hotmail.com`
  - cancellation case: one scheduled draft was created, then deleted; it was absent from Drafts and Inbox;
  - delivery case: the message was absent from Inbox before the requested time, then appeared exactly once in Inbox and exactly once in Sent Items after the requested time;
  - no duplicate delivery was observed.

The implementation follows Microsoft's documented draft/send lifecycle and Exchange deferred-delivery properties:

- https://learn.microsoft.com/en-us/graph/api/user-post-messages?view=graph-rest-1.0
- https://learn.microsoft.com/en-us/graph/api/message-send?view=graph-rest-1.0
- https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxomsg/e6793635-16cf-4356-913f-199f1a286cd6

The local environment had an intermittent proxy TLS timeout for Python HTTP clients, so the live validation process used the already-authorized access token with proxy inheritance disabled. The production code was not changed for that environment issue.

wu-changxing:
Thanks for the live validation on a real mailbox; the two-request lifecycle and the cancellation case are exactly the evidence this needed. Two things stand between this and merge:

1. **Rebase onto `main`.** The base is from 2026-08-23; `main` is at 1.8.1 today and GitHub cannot compute mergeability for the branch as it stands. The Outlook files you touch have not changed on `main` since, so it should be a clean rebase.
2. **One question on what a user sees afterwards.** #1314 reports that `co outlook send --at` never says how to cancel, and that scheduling is split across three commands that do not point at each other. With this change the scheduled message becomes a draft carrying `PidTagDeferredSendTime`. Does that draft appear in `co outlook` listings before it goes out, and is deleting it (the cancellation you validated) reachable from the CLI, or only through Graph directly? A sentence in the PR description is enough; if the answer is "Graph only", say so and we track the CLI side under #1314 rather than here.

---

wu-changxing:
Merged as `61e4759`. Answering my own question from the code: the deferred draft carries `SystemTime 0x3FEF`, which is exactly what `get_scheduled()` filters Drafts on, so it shows up in `co outlook scheduled` and `co outlook cancel <#>` deletes it. No CLI gap introduced here.

Two things this PR intentionally left alone are tracked in #1416: `reply --at` still goes through the `/reply` action, and `docs/cli/outlook.md` still says scheduling needs only `Mail.Send`.

---

PR #1223, opened 2026-08-23
fix: stream Work Room permission downgrade after outer mode change

### LINKED ISSUE #1222: [1.7 blocker] Outer Auto downgrade leaves Codex Work Room on Full Access until reconnect
## Release impact

Discovered by the exact `co 1.7.0rc3` + production-build O Chat `a6e9d09f5bcf65801416a7170b5823ee8f6a2ea8` live release gate for #792.

This is a 1.7 stable authority blocker: the outer COAI permission mode is the ceiling for every provider Work Room.

## Reproduction

1. Start the exact public `co 1.7.0rc3` with bounded Full access.
2. Complete a native Codex turn and open its Work Room.
3. Confirm the four Codex profiles are present.
4. Select Read Only and wait for the Host ACK.
5. Select Full Access, separately confirm the elevated risk, and wait for the Host ACK.
6. Close the Work Room and lower the outer COAI mode from Full access to Auto.
7. Reopen the same Codex Work Room.

Observed browser state after 30 seconds:

```json
{
  "provider": "Codex",
  "activeLabel": "Full Access",
  "outerMode": "Auto"
}
```

The durable mode transaction reconciles and appends a narrowed `provider_invocation` revision, but the WebSocket mode handler sends only `mode_changed`. The connected client therefore never receives the appended provider revision until a later reconnect/replay.

## Required behavior

- [ ] Lowering the outer mode immediately emits every provider lifecycle revision appended by the same durable transaction.
- [ ] The browser commits the outer mode and the narrowed provider state without reconnect.
- [ ] Under outer Auto, Codex Full Access is no longer active and is visibly disabled with the Host-ceiling reason.
- [ ] The behavior is revision-monotonic and does not replay unrelated or duplicate trace items.
- [ ] A focused Host WebSocket regression test covers Full access → Auto, not only the existing detached-policy Full access → Read only unit case.
- [ ] Repeat the exact `co ai` + `co browser` release gate and capture allowed/denied screenshots.

## Scope

Bounded 1.7 security/authority correction. No new permission profile or product feature.

wu-changxing:
## Summary
- return provider permission lifecycle revisions appended by the atomic outer-mode transaction
- send those revisions immediately after `mode_changed`
- keep the existing session-mode helper return contract intact
- cover the Full Access → Auto regression and duplicate-free idempotent repeat

## Why
A real installed-artifact 1.7 gate found that storage correctly narrowed Codex authority, but the open Work Room retained Full Access until reconnect because the WebSocket only sent `mode_changed`.

## Verification
- 95 focused and adjacent tests pass
- 129 broader related tests pass
- full suite: 7169 passed, 18 skipped; the sole docs checkout drift failure passes when pointed at docs PR wu-changxing/connectonion-docs-site#103
- `git diff --check`

Closes #1222
Tracks #792.

### DIFF
--- a/connectonion/network/host/session/mode.py
+++ b/connectonion/network/host/session/mode.py
@@ -190,22 +190,60 @@ def commit_host_session_mode(
 ) -> Session:
     """Commit one idle mode change under the cross-worker storage lock."""
 
+    record, _ = commit_host_session_mode_with_events(
+        storage,
+        registry,
+        session_id,
+        owner,
+        mode_id,
+        policy,
+        is_admin,
+    )
+    return record
+
+
+def commit_host_session_mode_with_events(
+    storage: SessionStorage,
+    registry,
+    session_id: str,
+    owner: str,
+    mode_id: Any,
+    policy: HostPermissionPolicy,
+    is_admin: bool,
+) -> tuple[Session, tuple[dict[str, Any], ...]]:
+    """Commit a mode change and return provider revisions it appended."""
+
     active = registry.get(session_id) if registry is not None else None
     if active is not None and getattr(active, "owner", None) not in (None, owner):
         raise _not_found()
     if active is not None and getattr(active, "status", None) == "running":
         raise _busy()
 
+    appended_provider_events: tuple[dict[str, Any], ...] = ()
+
     def commit(current: Session | None) -> Session:
+        nonlocal appended_provider_events
         if current is None or session_owner(current) != owner:
             raise _not_found()
         if current.status in SessionStorage.UNFINISHED:
             raise _busy()
-        session = policy.normalized(current.session or {}, is_admin=is_admin)
+        current_session = current.session or {}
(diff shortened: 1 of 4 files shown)


wu-changxing:
Exact local-candidate E2E evidence at head `2456e36a`:

- fresh invite onboarding, real co-browser fixture, C11/C++20/Rust, native Codex, responsive layouts: passed
- regression under #1222: Full Access -> outer Auto streamed immediately to the same open Codex Work Room as Ask for approval; Full Access became disabled; acknowledgement passed
- later full-gate Claude delegation did not start: the parent model announced it would call Claude Code but emitted no `claude_code` tool event, and the run timed out
- isolated direct Claude adapter smoke immediately afterward passed on Claude Code 2.1.235 (`exit_code=0`, exact marker), so this was not a local Claude authorization failure

The #1222 acceptance path is proved, but I am retaining the full E2E gate as failed and will repeat it before merge/release.

Issue #1251, opened 2026-08-24
release: publish and validate ConnectOnion 1.7.0rc8


wu-changxing:
## Candidate source

Cut `1.7.0rc8` from `release/1.7` only after #1249, at or after merge `0c901fb511c2cea3fc004e21b8decdb3a9e2cc01`. RC7 is superseded for Stable qualification by #1248.

## Acceptance

- [ ] Version sources, lockfile, VERSIONING record, and release notes agree on `1.7.0rc8`.
- [ ] Full release CI, build, Twine checks, wheel install, and public PyPI/GitHub artifact hash comparison pass.
- [ ] Exact public RC8 repeats the complete production-style browser/C11/C++20/Rust/native Codex/native Claude Code/voice/Stop/reconnect/responsive UI gate.
- [ ] Browser state polling and per-tab cleanup remain within their finite command deadlines after the #1249 import-boundary fix.
- [ ] A successful provider tool result receives a non-empty parent handoff and restores the outer composer (#1242).
- [ ] The Claude Workroom accepts a provider-targeted follow-up and renders exactly one attributed user message and one provider reply.
- [ ] Exact Codex permissions and outer-ceiling downgrade checks remain green.
- [ ] Manual design review passes every final screenshot; screenshots stay local unless separately authorized for O Chat PR #232.
- [ ] Stable remains untagged until #792 records every final gate, #176 is resolved, and the unchanged-candidate window passes.

Whole-browser graceful shutdown latency is separately tracked for 1.7.1 in #1250 and is non-blocking because the release harness owns exact isolated daemon/Chrome PIDs and has bounded cleanup.

No private invite, account, provider transcript, raw log, or screenshot upload belongs in this issue.

wu-changxing:
RC8 publication is complete from the exact reviewed candidate.

- release PR: #1252
- reviewed head: `a02f9d7c225ea32d096e93474a261c664ccc97f3`
- merge/tag target: `e1f243150468b2822bcc7f8296109871ff43b993`
- tag: `v1.7.0rc8`
- official Trusted Publishing run: https://github.com/openonion/connectonion/actions/runs/32695658590 (success)
- GitHub prerelease: https://github.com/openonion/connectonion/releases/tag/v1.7.0rc8
- PyPI: `connectonion==1.7.0rc8`
- public clean install: `python_version=1.7.0rc8`, `co 1.7.0rc8`, `pip check` reports no broken requirements
[...]

Issue #1242, opened 2026-08-24
[1.7 RC blocker] Parent Agent can remain active after successful Claude Code return


wu-changxing:
## Release impact

Blocks the unchanged-artifact 1.7 RC6 end-to-end gate.

In an exact public `connectonion==1.7.0rc6` run, the native `claude_code` tool completed successfully in 118.1 seconds and produced a valid, independently compilable C11 project. The parent Agent then started its next model step, but never emitted a final assistant message or terminal outcome. O Chat therefore never restored the outer composer, and the bounded five-minute release gate timed out.

This is not the deferred server Anthropic API-key issue: Claude Code authenticated, ran, wrote the scoped files, and returned success.

## Sanitized evidence

- Host: `claude_code(...) ✓ 118.1s`
- Project contains `stack.h`, `stack.c`, `test_stack.c`, `README.md`, and a passing test binary.
- After the tool result the parent logged its next model step, then remained active without a terminal event.
- Browser at timeout: `running=false`, `sendReady=false`, `composerPresent=false`.
- No approval was pending.
- The exact same gate already passed this Claude start/follow-up path on earlier RC2/RC3 candidates, so this may be a race or lost terminal settlement rather than a deterministic adapter failure.

## Required behavior

Every successful provider tool return must lead to exactly one bounded parent outcome: continue with another action, request a surfaced approval/input, complete with a final assistant message, or fail/cancel with a safe error. The parent session must not remain active indefinitely after a provider tool has returned.

## Acceptance

- [ ] Add a regression where `claude_code` returns successfully and the following model step produces no usable content/event; the Agent settles instead of hanging.
- [ ] Preserve exactly-once provider terminal and parent terminal semantics.
- [ ] Restore the outer composer after settlement.
- [ ] Do not convert a successful Claude result into `stopped` or fabricate a success message.
- [ ] Keep approval and interrupt behavior fail-closed.
- [ ] Repeat the exact public-artifact Claude parent + Workroom continuation gate.

Related: #792, #1109, release-candidate issue #1240.

wu-changxing:
RC7 follow-up: PR #1243 is merged and the official `1.7.0rc7` package is public. The exact live runs did not reproduce the original “successful Claude result followed by a blank terminal model response” shape: Claude completed and the parent moved into a subsequent verification/model step. However, the full RC7 regression could not reach an accepted terminal manifest because browser state polling repeatedly became too slow; Core #1248 is now the active release blocker.

[...]

PR #1253, opened 2026-08-24
fix(agent): bound post-provider settlement

wu-changxing:
## Summary

- bound every parent-model call after a native Codex or Claude Code tool returns
- abandon a timed-out model response without letting its late result enter history
- make one constrained recovery request grounded only in committed tool results
- emit an explicit terminal `TimeoutError` if that retry also fails to settle
- trace each timed-out LLM call as an error so hosted clients can restore their composer honestly

## Why

The exact public `connectonion==1.7.0rc8` + O Chat PR #232 gate reproduced #1242 in a form the RC7 fix did not cover. Native Claude Code completed successfully in 209.8s and produced the valid C11 project. The parent then ran a successful `glob` verification, but its following model call never returned. Because no model response existed, the empty-response recovery in #1243 could not run; after the 420s parent window O Chat still had no send-ready composer.

This change gives post-provider settlement only four outcomes: continue, ask, complete, or fail. A permanently in-flight parent model call is no longer an unbounded fifth state.

## Verification

- `tests/unit/test_interrupt.py tests/unit/test_agent.py`: 63 passed
- `tests/unit/test_asgi.py tests/unit/test_runtime_input.py tests/unit/test_turn_outcome.py tests/unit/test_co_ai_one_shot_sessions.py`: 92 passed
- Ruff on changed Core and interrupt test files: passed
- `git diff --check`: passed
- full `tests/unit tests/e2e/cli` is running against coordinated RC8 docs and will be posted separately

**Proposed target version:** 1.7.0rc9

**Release impact:** blocks Stable 1.7 until merged, published in a new RC, and the exact public-artifact Workroom gate passes.

Refs #1242, #792, #1251.

**Estimated release window:** current 1.7 release-candidate window

### DIFF
--- a/connectonion/core/agent.py
+++ b/connectonion/core/agent.py
@@ -20,7 +20,7 @@ from uuid import uuid4
 from ..logger import Logger
 from ..prompts import load_system_prompt
 from .events import EventHandler
-from .interrupt import run_interruptible
+from .interrupt import InterruptibleStepTimeout, run_interruptible
 from .llm import LLM, TokenUsage, create_llm
 from .mode import FULL_ACCESS, full_access_turns_left, mode_of, set_mode
 from .provider_messages import messages_for_provider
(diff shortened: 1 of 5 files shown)


wu-changxing:
Full local release-branch regression is green on exact head `fefe4f0fc7f047e123ab550a3c4995c064b56413`:

- command: `CONNECTONION_DOCS_SITE=/private/tmp/connectonion-docs-rc8 python -m pytest tests/unit tests/e2e/cli -q`
- selected: 7,077
- result: 7,061 passed, 16 platform/optional skips, 28 deselected, 0 failed
- duration: 838.07s

Focused post-provider/interrupt tests also passed 63/63, and ASGI/runtime-input/turn-outcome/one-shot coverage passed 92/92. Ruff on changed Core files and `git diff --check` are clean.

PR #1257, opened 2026-08-24
fix: arm native-provider settlement before hosted execution

wu-changxing:
Closes #1256.

## Summary

- arm the native-provider settlement deadline from the model decision before hosted tool execution
- re-derive that deadline from the durable provider `tool_result` trace as a hosted fail-safe
- record the applied timeout on `llm_call` trace entries so release evidence proves the boundary was active
- add a Design Journal entry for the RC9 public-gate finding

## Why

The exact public RC9 gate completed a real Claude Code C11 project in 132.8 seconds, but the following parent-model call did not emit the expected 90-second timeout/retry and left the Work Room without a composer after the harness 420-second bound. Minimal Agent tests retained the original ToolCall identity; the hosted path needs the deadline armed before observers can normalize that call, with the durable trace as the authoritative fallback.

## Verification

- 370/370 agent, interrupt, turn-outcome, one-shot, ASGI, hosted mode, provider Work Room, provider permission, runtime-input, and session tests pass
- `git diff --check`

## Release gate

After merge, a local hosted Claude smoke must prove the timeout trace and usable terminal state before RC10 is cut. The exact public RC10 artifact must then repeat the complete gate; RC9 cannot promote unchanged.

**Proposed target version:** 1.7.0rc10

**Estimated release window:** immediately after green CI and hosted smoke

PR #1258, opened 2026-08-24
fix: make native-provider settlement terminal-only

### LINKED ISSUE #1256: fix: bound native-provider settlement in the real co-ai Work Room path
## Problem

The exact public 1.7.0rc9 gate proves a native Claude Code invocation can return successfully while the following parent-model settlement remains unbounded in the real hosted co-ai path.

## Evidence

- RC9 exact commit: `aaf1b1d2cad1a35141ef8e1c43303b0afca0690b`
- Claude tool returned successfully in 132.8s and produced a valid independently tested C11 stack project.
- The expected 90-second settlement timeout/retry trace never appeared.
- After the 420-second harness bound: `running=false`, `stopped=true`, `sendReady=false`, `composerPresent=false`, no approval pending.
- Sanitized local evidence: `/private/tmp/oo-rc9-live2.nSMpDn/evidence`

## Acceptance criteria

- [ ] native-provider completion marks the next parent settlement before any post-tool path can lose that fact
- [ ] the real co-ai tool registry (`codex`, `claude_code`) is covered, not only a minimal Agent fake
- [ ] first settlement timeout abandons late results and performs one grounded concise retry
- [ ] second timeout reaches an explicit terminal failed outcome that restores a usable composer
- [ ] success, Stop, and non-provider tools retain existing behavior
- [ ] exact public next RC repeats the full Work Room gate

## Release target

**Proposed target version:** 1.7.0rc10

**Estimated release window:** immediately after review, full CI, and exact public gate

wu-changxing:
Closes #1256.

## Summary

- stop the parent agent from opening a new outer tool chain after native Codex/Claude work completes
- discard one attempted post-provider tool batch and request a bounded final-only answer
- fail explicitly on a repeated tool attempt so Host can publish a terminal outcome and restore the composer
- extend the Design Journal with the installed-wheel smoke evidence

## Evidence

The #1257 installed-wheel smoke refined the failure: Claude returned in 102.4s; the first parent call returned `glob` in 3.3s and executed; the next parent call returned another one-tool response in 2.8s, then wedged before that tool reached visible execution or approval. The LLM timeout was not the blocking boundary.

## Verification

- 95/95 agent, interrupt, turn outcome, and runtime-input tests pass
- regression proves the attempted `glob` is not executed and a final answer follows
- `git diff --check`

## Release gate

An installed-wheel hosted smoke must pass before RC10 is cut, then exact public RC10 must repeat the complete release journey.

**Proposed target version:** 1.7.0rc10

**Estimated release window:** immediately after green CI and installed-wheel hosted smoke

### DIFF
--- a/connectonion/core/agent.py
+++ b/connectonion/core/agent.py
@@ -653,6 +653,7 @@ class Agent:
         empty_terminal_retries = 0
         provider_settlement_required = False
         provider_timeout_retries = 0
+        provider_tool_retries = 0
         while self.current_session['iteration'] < max_iterations:
             self.current_session['iteration'] += 1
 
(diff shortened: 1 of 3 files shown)


wu-changxing:
Installed-wheel hosted acceptance on exact head `ffec0d208be965739039da8a6cd9ed8055ed3447` confirms this PR fixes #1256's settlement boundary:

- real browser, C11, C++20, Rust, native Codex, permissions, voice-denial draft, and responsive checks reached
- real Claude Code created, compiled, and tested the C11 project
- after Claude returned, the parent Workroom settled terminally in 185.5s
- `composerPresent=true`, `composerEnabled=true`, `sendReady=true`; no approval was stranded

The remaining gate failure is a separate second-continuation lifecycle defect: a subsequent Claude direct turn ran, but its reused invocation ID restarted at an older state revision, so the browser correctly rejected the stale running frame and did not expose Stop. That will be tracked independently; RC10 remains withheld until it passes.

Sanitized evidence: `/private/tmp/oo-1258-smoke.kJNRP6/evidence` (local only; no private logs or screenshots uploaded).

PR #1264, opened 2026-08-24
release: promote accepted RC12 code to 1.7.0 Stable

### LINKED ISSUE #1263: release: promote unchanged RC10 code to ConnectOnion 1.7.0 Stable
## Scope

Promote the code already published and accepted as `1.7.0rc10` to `1.7.0` Stable without source changes beyond release metadata. Source of truth: #792.

## Preconditions

- [x] Exact public RC10 full acceptance passed.
- [x] `@connectonion/react@0.4.3` published to npm `latest` and public install verified.
- [ ] O Chat #232 final Stable-dependency head has authorized visual evidence, passes all checks, merges, and deploys.
- [ ] Final unchanged-candidate disposition recorded in #792.

## Release steps

- [ ] Change only version metadata from `1.7.0rc10` to `1.7.0` and add the Stable history entry.
- [ ] Full release CI passes on `release/1.7`.
- [ ] Tag exact reviewed release commit `v1.7.0`; approve protected PyPI environment.
- [ ] Trusted workflow publishes and verifies exact PyPI/GitHub artifacts.
- [ ] Normal clean install resolves `connectonion==1.7.0` and `co --version` reports it.
- [ ] Merge the completed release branch back to `main` without moving the release tag.
- [ ] Update production docs and close #792 after coordinated verification.

No feature or bug-fix source code belongs in this promotion PR.

### LINKED ISSUE #1269: image_result_formatter: a truncated data-URL in any tool output crashes the whole agent loop with binascii 'Incorrect padding'
## What happens

If any successful tool result happens to contain a **truncated** `data:image/...;base64,...` substring, the agent loop dies:

```
File "connectonion/useful_plugins/image_result_formatter.py", line 118, in ...
    image_url = _upload_to_oo_api(base64_data, mime_type)
File "connectonion/useful_plugins/image_result_formatter.py", line 162, in _upload_to_oo_api
    files={"file": ("screenshot", base64.b64decode(base64_data), mime_type)},
File "python3.10/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
Error: Incorrect padding
```

The exception escapes the `after_tools` event (`agent._invoke_events('after_tools')` in `tool_executor.py:144`) and takes down the entire `co ai` run — not just the one screenshot.

## Root cause

`_is_base64_image()` detects images with this regex:

```python
data_url_pattern = r'data:image/(png|jpeg|jpg|gif|webp);base64,([A-Za-z0-9+/=]+)'
```

Two problems:

1. **It matches any substring, with no length/padding validation.** A DOM-scanning page script that returns JSON like `{"src": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAABkAAAAOECAYAAAD5Tf"}` (an `` attribute truncated by the script for readability) matches the regex, and the truncated payload goes straight to `b64decode` → crash.
2. **The failure is fatal by design** — the docstring says *"Fails loudly on upload error"*, which is right for a failed HTTP upload, but a malformed candidate should mean *"this was never an image, skip it"*, not *"kill the run"*.

## Real-world impact (all on 1.6.8, macOS)

- **2026-08-24 04:00** — nightly LinkedIn pipeline, engage step: crashed mid-run on page content it scanned (`~/.co/cron/li-run-20260824-0400/3-engage.log`).
- **2026-08-25** — `/linkedin-article-orchestrator` scheduling run: crashed 3 times in a row at the same point, right after `verify-cover-upload.js` returned JSON containing truncated `` data-URLs from LinkedIn's cover-confirmation page. Reproduced across a fresh browser daemon, so it is content-triggered, not state-triggered. We had to bypass `co ai` entirely and drive `co browser` by hand to get the article scheduled.

## Still present on every version

`useful_plugins/image_result_formatter.py` is **byte-identical** from 1.6.8 through 1.7.0rc11 (verified by file diff of installed packages), so upgrading does not help.

## Suggested fix

In `_is_base64_image()` / before upload:

- validate the candidate: `len(base64_data) % 4 == 0` and a minimum length (a real screenshot is tens of KB of base64; a 40-char match is never an image)
- or `base64.b64decode(base64_data, validate=True)` inside a `try` — on failure return `(False, "", "")` and leave the tool result untouched
- keep failing loudly on actual **upload** errors (network/5xx), which is the case the docstring is about. A malformed candidate is not a broken image pipeline; it is a false positive of the detector.

Happy to PR this if useful.

### LINKED ISSUE #1270: 1.7.0rc11 denies even whitelisted commands for unattended agents: 'outside the focused verification allowlist; no approval channel is available'
## What happens

On **1.7.0rc11**, an unattended `co ai` run (no TTY, launched the way our launchd cron does) can no longer execute even the commands the 1.6.x whitelist explicitly allows. First bash call of the run:

```
Tool 'bash' denied by connectonion.auto: command is outside the focused
verification allowlist; no approval channel is available
```

Tested three ways, same result each time:

1. bare prompt, arbitrary cwd → denied
2. `cwd=$HOME` (the approval workspace, exactly what our cron script sets) → denied
3. **the real production path** — `cd ~; co ai -m co/gemini-3.7-flash -i 5 "/linkedin-notifications ..." < /dev/null` asking for exactly one command, `co browser status` → denied

`co browser status` is the first command of every step of our nightly LinkedIn pipeline and is covered by the standing `Bash(co *)` whitelist on 1.6.x. On rc11 the deny fires on iteration 1, so **every unattended round would die at its first action** — a total outage, not a degradation.

Rolled back to 1.6.8 and re-ran the identical invocation: works.

## Why this matters beyond our setup

This is the same failure mode as #1126 ("1.6.11 approval policy breaks unattended agents"), which #1169/#1170 fixed with the fail-closed Auto policy — and to be clear, **failing closed is the right default**. The gap is that there is still no documented, working way for a *deliberately* unattended agent to grant itself the standing permissions it needs:

- the deny message says "no approval channel is available" but does not say what an operator should do about it for a cron job
- `~/.co/whitelist.txt` (or wherever `Bash(co *)` grants live on 1.6.x) appears not to be honored by `connectonion.auto` on rc11
- #1126 itself named the missing piece: *"no way to grant out-of-workspace reads or select a profile headless"* — the headless-profile-selection half still seems absent, or exists but is undocumented

## What would resolve it

Any one of:

1. `connectonion.auto` honors the existing standing whitelist (`Bash(co *)`), so 1.6.x cron setups keep working on upgrade, or
2. a documented flag/profile for unattended runs (e.g. `co ai --profile <name>` or an explicit `--unattended-allowlist <file>`), with the deny message pointing at it, or
3. release notes for 1.7.0 stating the migration step cron/CI operators must take before upgrading.

Without one of these, 1.7.0 stable is a silent breaking change for every unattended deployment.

## Environment

macOS, Python 3.10.13 (pyenv), `pip install connectonion==1.7.0rc11` into the same environment that runs 1.6.8 cleanly. Model `co/gemini-3.7-flash`. Invocations run with `< /dev/null` (no TTY), matching launchd.

wu-changxing:
## Description

Promote the product source accepted as public `1.7.0rc12` to Stable `1.7.0`.

RC11 was not promoted unchanged: late release evidence exposed #1269 and #1270. Both blockers are fixed on `release/1.7`, carried to `main`/1.8, and published in RC12. This branch now contains RC12 product behavior unchanged, plus only Stable release metadata and the Stable release journal.

## Related issue

Closes #1263 after coordinated promotion completes. Source of truth: #792. Late blockers: #1269 and #1270.

## Stable-only diff

Relative to `release/1.7`, only these five release files change:

- `pyproject.toml`
- `connectonion/_version.py`
- `uv.lock`
- `VERSIONING.md`
- `docs/blog/2026-08-24-the-last-change-is-the-name.md`

No product behavior is introduced after RC12.

## Verification

- 64 release, artifact, versioning, and forward-port contract tests passed.
- The separate docs-channel assertion remains intentionally pending because production must continue advertising Stable 1.6.12 until PyPI serves 1.7.0.
- `uv build` produced the 1.7.0 wheel and sdist.
- `twine check dist/*` passed.
- `git diff --check` passed.
- Exact public RC12 clean-install acceptance passed on CPython 3.13: truncated image-looking output remains ordinary text, configured headless `co browser status` is allowed, and `co deploy` remains denied.

## Remaining coordinated gates

- [x] Publish RC12 and verify exact public artifacts.
- [x] Fix #1269/#1274 on 1.7 and #1276 on main/1.8.
- [x] Fix #1270/#1275 on 1.7 and #1277 on main/1.8.
- [x] O Chat #232 final visual declaration, merge, and production deploy passed at `6e24af36a3f70c1675251ae3b01588a38c6c219e`.
- [ ] Merge/tag/publish Core and verify public Stable artifacts.
- [ ] Switch production docs to Stable only after PyPI serves `1.7.0`.
- [ ] Backmerge the final release to `main`.

This PR remains Draft until the coordinated gates above are complete.

**Proposed target version:** 1.7.0 Stable

**Estimated release window:** after the authorization-gated O Chat #232 visual and deploy gate

**Forward-port tracking issue (stable patches only):** N/A — this is the 1.7.0 minor Stable promotion


wu-changxing:
### Exact public RC11 installed-artifact gate passed

The complete release harness now passes against this exact tuple:

- Core: public `co 1.7.0rc11` (executable SHA-256 `ce72cae739e89a6f86c8fa72918e5c68429090d5bfe633e697fe5a2419cd9ec6`)
- React: `@connectonion/react@0.4.3`
- O Chat: `ff92ea88fe00e1dd28f43d1f3c946e550460b5a7`
- Codex: `codex-cli 0.147.0`
- Claude Code: `2.1.241`
- frontend: exact local production build
- onboarding: invocation-scoped temporary invite and isolated browser identity

Authoritative checks passed:

- real `co browser` deterministic search and attachment download, independently checked from fixture/report evidence;
- strict C11, C++20, Rust, a second native-Codex C11 project, and a native-Claude C11 project, all independently rebuilt/tested;
- Codex and Claude Workroom conversation, composer, follow-up, voice recovery, and provider-scoped Stop;
- Codex four-profile catalog, acknowledged Read Only → separately confirmed Full Access, outer ceiling downgrade to Auto, stale Full Access disabled, and Ask for approval acknowledgement;
- Full access, Read only, Auto, outer Stop, and terminal composer recovery;
- Host restart: disconnected showed Reconnect and not Retry; one click/automatic handoff restored connected state; prompt occurrence stayed exactly one and no duplicate Host `INPUT` was emitted;
- 1440×900, 768×1024, and 390×844 layouts had no horizontal overflow.

The technical manifest initially set `uiReviewPassed: false` as designed. I then inspected all 28 local screenshots, recorded a concrete six-part review, and finalized the same bundle. Every technical check and `uiReviewPassed` is now true.

Local-only evidence hashes:

- finalized manifest: `909eed7517959b897f3a8ef92cc6b38cc231b9ea8c1660672c2ef4f27bc9bde8`
- UI review: `1db5245cb5f4e1f6f403f28ede1acbf7a0139a6dfaebbfd0fe590b2eaae7a970`

One non-blocking medium visual-polish observation remains: the desktop three-column outer workspace feels sparse in some settled states. It does not remove conversation, composer, process state, permissions, Stop/reconnect, or responsive access and can continue after 1.7.

No screenshot or evidence bundle was uploaded to O Chat #232. Its explicit authorization gate remains unchanged.

Issue #499, opened 2026-07-31
Replace the serial browser daemon loop with concurrent async sessions and safe tab claims


wu-changxing:
Parent epic: #208

## Implementation status — audited against `origin/main` on 2026-08-10

**Partially implemented foundations, but the concurrency objective is not implemented.** Already on `main`:

- per-session tabs and binding, ownership metadata, TTL/LRU reclamation (#186);
- named tab claims and multi-agent CLI behavior (#195);
- authenticated native Windows transport and bounded transport deadlines (#209).

The daemon still describes and implements a single-threaded serial serve loop, and the browser still owns a one-worker executor. One long browser operation therefore still blocks unrelated work. There is no asyncio server, interleaving operation scheduler, async cancellation model or slow-reader/backpressure load coverage.

## Scope remaining

- Run protocol connections and browser operations on one asyncio architecture
- Redesign tab claims/leases for genuinely interleaving operations
- Bound per-client work and backpressure
- Preserve the authentication, liveness and ownership rules already shipped

## Definition of done

- [x] Per-session tabs and ownership metadata exist
- [x] Native Windows/POSIX transport seam and authentication tests exist
- [ ] Two independent tabs make progress concurrently during long operations
- [ ] Conflicting claims are deterministic and auditable under interleaving
- [ ] Cancel/disconnect cannot orphan ownership or browser resources
- [ ] Load tests cover slow readers, long tasks and stalled clients
- [ ] No single one-worker executor remains as the global browser bottleneck

## Dependencies

Depends on #498. Parent epic: #208.

## Non-goals

Do not weaken #209's fail-closed authkey, corrupt-key recovery or liveness behavior.

wu-changxing:
1.8 implementation checkpoint (2026-08-26): the foundations listed here remain present, but the daemon is still a serial `accept → receive → dispatch → send` loop and the browser still has one global worker. #1282 has now merged the focused-element/destructive-shortcut safety primitive that concurrent Remote Browser flows can rely on.

Planned review boundary after #498:

- one asyncio-owned browser runtime; a lifecycle lock for context launch/close and a lock per tab, so conflicting operations on one tab serialize while independent tabs can interleave;
- concurrent per-connection tasks with bounded request size, receive/send deadlines, a global in-flight cap, and explicit slow-reader backpressure;
- preserve current claim identity, expiry, audit metadata, authkey recovery, stale-endpoint rules, and exit-code behavior;
[...]

Issue #498, opened 2026-07-31
Port the browser driver core to patchright.async_api


wu-changxing:
Parent epic: #208

## Implementation status — audited against `origin/main` on 2026-08-10

**Not implemented yet.** The current driver imports `patchright.sync_api`, calls `sync_playwright().start()`, and routes public methods through a one-worker `ThreadPoolExecutor`. No `patchright.async_api`/`async_playwright` driver core exists on `main`.

Async tool-function support from #125 is already complete and remains a prerequisite, not the browser port itself.

## Scope

- Port lifecycle, pages, selectors, input, downloads, uploads, screenshots and stealth checks
- Preserve error and timeout semantics
- Keep humanized input behavior
- Measure regressions against current browser tests

## Definition of done

- [ ] No sync Patchright call remains in the async core
- [ ] Behavioral and stealth checks stay green
- [ ] Resource cleanup is deterministic under cancellation
- [ ] Existing capture/injection, per-session tab and persistent-profile behavior is preserved

## Dependencies

Parent RFC #208; async tool execution #125 is complete.

## Non-goals

Do not break the public synchronous API; #500 owns the compatibility facade.

wu-changxing:
1.8 implementation checkpoint (2026-08-26), after #1282 merged:

The audit still matches the issue: `browser.py` is 2,240 lines, imports `patchright.sync_api`, starts `sync_playwright()`, and routes public methods through a one-worker executor. The helper modules (`element_finder.py` and `humanize.py`) also call the sync Page API, so changing only the lifecycle import would create a mixed sync/async core and would not satisfy this issue.

Planned review boundary for #498:

1. introduce an internal async driver/runtime with explicit lifecycle and per-session page ownership;
2. port element extraction and humanized input alongside the methods that consume them;
3. preserve current result strings, timeout behavior, persistent-profile rules, capture/injection hooks, and the newly merged focused-element safety contract from #1151;
4. add cancellation tests that prove page/context/driver cleanup, plus contract tests that run equivalent sync-facade and async-core calls against the same fixtures;
5. keep the public synchronous facade until #500, but require that no `patchright.sync_api` call remains beneath the new async core before #498 closes.

This should land before #499 changes daemon scheduling. Mixing the driver port and multi-client transport rewrite in one PR would make regressions in browser semantics indistinguishable from scheduling bugs. The detailed release dependency remains in #1281.

wu-changxing:
Foundation PR opened: #1283

[...]

PR #1290, opened 2026-08-26
feat(browser): preserve sync API over async core

### LINKED ISSUE #500: Preserve the synchronous BrowserAutomation API and migrate cross-platform IPC/tests
Parent epic: #208

## Implementation status — audited against `origin/main` on 2026-08-10

**Partially implemented.** Native cross-platform IPC and its tests are already delivered by #209:

- POSIX keeps AF_UNIX behavior;
- Windows uses authenticated named pipes;
- bounded handshake/I/O, self-healing authkey and liveness regression tests exist;
- `windows-latest` CI coverage exists for the transport path.

The public synchronous `BrowserAutomation` API also still works, but only because the core remains synchronous. The actual compatibility facade over the future async core does not exist and cannot be completed before #498/#499.

## Scope remaining

- Implement the sync facade boundary over the async driver/runtime
- Migrate all in-repo synchronous callers without changing their public behavior
- Reuse or deliberately replace #209's transport framing while preserving its security invariants
- Publish compatibility/deprecation rules

## Definition of done

- [x] Native Windows and POSIX transport tests pass
- [x] Current synchronous examples/public methods work on the current sync core
- [ ] The same API works over the async core from #498/#499
- [ ] Nested/running event-loop callers have defined behavior
- [ ] Repeated creation, cancellation and shutdown leak no event loop, thread, pipe or process
- [ ] Compatibility/deprecation behavior is documented

## Dependencies

Depends on #498 and #499. Parent epic: #208.

## Non-goals

Do not introduce an unplanned public breaking change.

wu-changxing:
**Estimated release window:** after #1288 and #1289, before the 1.8 alpha entry gate
**Forward-port tracking issue (stable patches only):** not applicable

Closes #500

## Problem

The 1.8 browser driver and daemon are async, but existing Python callers and Agent tools depend on the synchronous BrowserAutomation contract. Nested asyncio.run calls would fail for server callers that already own an event loop, while waiting on the browser runtime thread would deadlock.

## Approach

- expose the existing public BrowserAutomation names, signatures, docstrings, context-manager behavior, and ordinary return values through a sync facade;
- own one AsyncBrowserCore, asyncio loop, and private runtime thread per facade instance;
- propagate the caller session binding across the thread boundary for every operation;
- support calls from threads that already run an asyncio loop;
- reject calls made from the owned runtime thread with a clear RuntimeError;
- make bound close release one tab and unbound close deterministically stop and join the runtime;
- migrate the stealth detector harness and `human_jitter` plugin from the removed private executor to a loop-owned internal adapter;
- retain the 1.7 implementation only as an internal comparison oracle while contract tests migrate;
- document the compatibility and deprecation boundary, update the 1.8 plan, and add the required engineering blog.

## Verification

- 126 focused async-core, facade, public-contract, headless, session, browser-tool, and `human_jitter` tests passed locally.
- 95 native daemon socket tests pass on macOS after the stacked #1289 lifecycle fix.
- The real stealth E2E module collects successfully against the facade; real Chrome execution remains a hosted authority because the managed local sandbox blocks Crashpad.
- Ruff passes for all new/modified facade boundary files.
- `git diff --check` passes.

## Stack

This PR is intentionally based on #1289 because issue #500 depends on #498 and #499. Merge order: #1288, #1289, then this PR.

### DIFF
--- a/connectonion/useful_plugins/human_jitter.py
+++ b/connectonion/useful_plugins/human_jitter.py
@@ -9,8 +9,8 @@ Use it on sites whose terms permit automation.
 
 Data flow: before_each_tool event -> read agent.current_session['pending_tool'] ->
 if its name starts with 'click', grab the BrowserAutomation instance off the tool
-registry (agent.tools.browserautomation) and dispatch jitter() onto the browser's
-single worker thread (Playwright is not thread-safe).
+registry (agent.tools.browserautomation) and dispatch the async jitter routine onto
+the browser core's owning event loop (Playwright objects are loop-bound).
 
 Why before_each_tool (not after_tools): this is the only hook that fires
 synchronously between "LLM picked a click tool" and "Playwright performs the click",
@@ -25,6 +25,7 @@ if the page is mid-navigation when jitter reads the viewport, the click must sti
 proceed, so that transient failure is caught and logged, never raised.
 """
 
+import asyncio
 import random
 import time
 from typing import TYPE_CHECKING
@@ -38,7 +39,7 @@ if TYPE_CHECKING:
 def jitter(page):
     """Wander the mouse through a few random points on the page.
 
-    Must run on the browser's worker thread (Playwright is single-threaded).
+    Legacy synchronous-driver helper; run it on that driver's owning thread.
     """
     w, h = page.evaluate("() => [window.innerWidth, window.innerHeight]")
     x = random.uniform(w * 0.2, w * 0.8)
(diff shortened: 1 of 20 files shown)


wu-changxing:
Closing as superseded by consolidated PR #1317. GitHub had deleted the stacked base after its dependency merged; #1317 explicitly preserved this PR's exact head `96711dd6c218a634ba0b75214133a5244b015567` as an ancestor and merged the reviewed synchronous-facade implementation into `main`.

Issue #500 is being closed against that actual merge. Keeping this old PR open would incorrectly imply the facade is still absent from `main`.

PR #1303, opened 2026-08-26
feat(remote-browser): pin egress sockets

wu-changxing:
## Description

Add the authenticated, bounded loopback gateway that resolves, classifies, and pins every future Remote Browser connection to an approved numeric socket. The component remains internal: no Chromium or OIP navigation command consumes it yet.

## Related Issue

Fixes #1302. Child of #1297 and #991. Stacked on classifier PR #1301 and design PR #1299; it must not merge ahead of either boundary.

## Labels and target release (required)

- **Suggested labels:** feature, browser, tests
- **Proposed target version:** 1.8.0
- **Estimated release window:** next alpha, before BrowserDaemon/Chromium integration
- **Milestone:** maintainer assigns the confirmed release milestone
- **Why this release:** URL policy is not a socket boundary while Chromium can resolve or dial independently. This gateway must be reviewed before navigation exists. Rollback deletes an unused internal component and leaves navigation unavailable; there is no Direct fallback.
- **Forward-port tracking issue (stable patches only):** N/A — alpha feature work

## How this was written

- [ ] I wrote this myself
- [ ] AI-assisted — I reviewed every line and can explain why each change is there
- [x] Mostly AI-generated

**Which model?** GPT-5.6 via Codex desktop.

The part I am least confident about is compatibility with Chromium's eventual effective proxy traffic: this PR deliberately tests HTTP/CONNECT protocol behavior without launching Chromium, so later preflight may reveal another conservative header form that must be supported without weakening framing. I did not run Python 3.10-3.13 or Windows locally; the new 12-runner OS/runtime matrix is intended to supply that evidence. I also did not test QUIC/WebRTC blocking, browser resolver flags, downloads, Service Workers, or Remote Browser `open` because those belong to the next native integration boundary. If this is wrong, the first break should be a focused socket/protocol test or a fail-closed proxy refusal before any public navigation API exists.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## Dev Blog (required)

Blog file in this PR:

> docs/blog/2026-08-26-the-proxy-replied-but-the-429-disappeared.md

## Changes Made

- bind one ephemeral IPv4 loopback listener with per-instance random Basic proxy credentials and constant-time authorization;
[...]

PR #1305, opened 2026-08-26
feat: isolate the Remote Browser private runtime

### LINKED ISSUE #1304: [1.8] Isolate Remote Browser daemon namespace, profile, and launch policy
## Parent and dependencies

- Child of #1297 and #991.
- Implements delivery step 3 from DD-056.
- Depends on destination classifier #1300 / PR #1301 and egress gateway #1302 / PR #1303.

## Problem

RemoteBrowserService currently delegates lifecycle to the ordinary local co browser daemon. That shares its socket or named pipe, singleton sidecars, persistent profile, environment-derived proxy configuration, and Chrome context with local interactive browsing. Chromium proxy policy is context-scoped, so this cannot become a safe Remote Browser navigation boundary.

The egress gateway also exists only as an unused component. Before any open command is exposed, the Host-private daemon must own the gateway and hold an explicit launch policy that cannot inherit BROWSER_PROXY, a local profile, extensions, Direct fallback, or an unreviewed alternate egress path.

## Decisions

1. Add an immutable browser runtime target/configuration passed explicitly through client spawn, BrowserDaemon, and AsyncBrowserCore. Environment variables remain compatibility inputs only for the ordinary local target.
2. Derive one deterministic Host-private namespace from the Remote Browser state boundary. It has a distinct Unix socket or Windows named pipe, PID/lock/auth sidecars, log, and persistent profile.
3. RemoteBrowserService defaults to the Host-private requester. Tests may still inject a requester. Local request_as and co browser keep their existing endpoint, profile, proxy environment behavior, and lifecycle.
4. The private BrowserDaemon constructs and starts EgressGateway before accepting Remote Browser lifecycle commands, owns it for its whole lifetime, and stops browser traffic before stopping the gateway. Failure to bind or keep the gateway available fails closed and never starts Chrome or falls back to Direct.
5. The daemon passes credentials in memory to an immutable remote launch policy. Secrets never appear in argv, logs, status, repr, registry JSON, exceptions, or OIP responses.
6. The launch policy has a fixed authenticated HTTP proxy, no bypass except subtracting Chromium implicit loopback bypass, no Direct alternative, target-host resolver blocking with only the numeric gateway endpoint exempted, QUIC disabled, non-proxied WebRTC UDP disabled, extensions disabled, Service Workers blocked, and an explicit private profile directory.
7. This slice records and tests requested launch options but does not claim effective Chromium enforcement. Native Chromium preflight and system-engine negative vectors remain delivery step 4.

## Acceptance criteria

- Local and Host-private daemons can bind and serve concurrently.
- Their endpoint, PID, lock, Windows authkey path, log, profile, and proxy settings are pairwise distinct.
- Starting/stopping/restarting either daemon does not mutate or stop the other.
- Ordinary local requests retain environment-derived BROWSER_PROXY behavior; the private target ignores that environment and receives only its owned gateway endpoint.
- BrowserDaemon starts its gateway before admission and shuts down browser/runtime work before gateway teardown; partial start and cancellation leave no listener or sidecar behind.
- Requested launch options are structurally asserted: authenticated fixed loopback proxy, loopback bypass subtraction, no Direct fallback, resolver blocking, QUIC/WebRTC alternate-path controls, disabled extensions, blocked Service Workers, and private profile.
- System and Onion engine selection can consume the same immutable policy interface; this slice does not yet launch or preflight either engine.
- Tab/session lifecycle still survives Host service reconstruction through the private native transport.
- Logs, process arguments, repr, state files, errors, and public envelopes contain no gateway password.
- Tests cover POSIX and Windows namespace derivation, concurrent daemon isolation, lifecycle ordering, gateway failure, environment contamination, and secret non-disclosure.
- A focused Linux/macOS/Windows x Python 3.10-3.13 CI job runs the private-runtime contracts in parallel.
- Documentation describes ownership, paths, lifecycle ordering, launch-policy fields, rollback, and the remaining preflight boundary.

## Non-goals

- No Remote Browser open, page action, screenshot, download, diagnose-ledger, redirect accounting, or browser navigation.
- No claim that requested Chromium flags are effective until native preflight in delivery step 4.
- No shared proxy mode, relay encryption, generic proxy product, TLS interception, firewall provisioning, or merge/release action.

## Implementation plan

1. Freeze typed runtime-target and launch-policy values, including secret-safe repr and validation.
2. Parameterize transport/client/daemon seams without changing local defaults.
3. Add Host-private target derivation and make RemoteBrowserService use it by default.
4. Give BrowserDaemon fail-closed gateway ownership and ordered cleanup.
5. Pass the immutable launch policy and explicit profile into AsyncBrowserCore; keep local environment behavior isolated.
6. Add unit/native-transport tests, concurrency/race tests, documentation, dev blog, and the 12-cell focused matrix.
7. Review the diff and hosted evidence. Keep the PR draft behind #1299, #1301, and #1303; do not merge or deploy.

## Rollback

Revert the private-runtime integration. Remote Browser lifecycle/navigation remains unavailable rather than reusing the local daemon or Direct egress.

wu-changxing:
**Estimated release window:** 1.8 alpha after stacked security parents and DD-056 native step 4

Closes #1304

Stacked on #1303. This is DD-056 delivery step 3 and intentionally keeps Remote Browser navigation unavailable.

## What changes

- derive a deterministic Host-private daemon namespace, PID/lock sidecars, profile, log, and Windows IPC key from the durable session-state path
- start the authenticated loopback egress gateway before browser construction and bind IPC only after both succeed
- pass an immutable, environment-independent launch policy to the existing AsyncBrowserCore
- request one canonical loopback proxy, subtract Chromium loopback bypass, fail browser-side DNS, disable QUIC/non-proxied WebRTC UDP/extensions, and block Service Workers
- close the browser before its gateway and reject every command before tab mutation if gateway authority disappears
- preserve ordinary local co browser environment/profile behavior and allow local/private native daemons to run and stop independently
- keep proxy credentials out of argv, registry state, paths, logs, reprs, and errors

## Verification

- 218 focused and adjacent tests passed locally, including ordinary daemon regressions
- focused Remote Browser suite: 64 passed
- Ruff on every changed Python file: passed
- compileall: passed
- full local run reached 457 passed and 4 skipped before interruption; its one completed failure was the test subprocess resolving a global co executable outside the worktree virtualenv
- CI extends the existing focused job across Linux/macOS/Windows and Python 3.10-3.13

## Security boundary

This PR proves requested launch configuration and private runtime lifecycle. It does not claim that pinned Chromium honored every switch or that every request class used the gateway. Navigation remains disabled until DD-056 step 4 passes native zero-socket redirect, subresource, worker, WebSocket, download, DNS-rebinding, and gateway-crash vectors.

## Rollback

Remove private target selection and the private daemon launch mode together. Remote Browser returns to lifecycle-only operation. Never point Remote Browser at the ordinary local daemon or remove the gateway while page commands remain enabled.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -102,9 +102,9 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e ".[dev]"
 
-      - name: Run destination policy and native loopback gateway contracts
+      - name: Run destination, gateway, and private runtime contracts
         run: |
-          python -m pytest tests/unit/test_remote_browser_destination_policy.py tests/unit/test_remote_browser_egress_gateway.py -v --tb=short
+          python -m pytest tests/unit/test_remote_browser_destination_policy.py tests/unit/test_remote_browser_egress_gateway.py tests/unit/test_remote_browser_private_runtime.py tests/unit/test_remote_browser_service.py tests/e2e/cli/test_remote_browser_daemon.py -v --tb=short
 
   # Validate the native Windows browser IPC (named-pipe daemon transport) — the
   # `co browser` daemon uses Unix sockets on POSIX and Windows named pipes here.
--- a/connectonion/cli/browser_agent/client.py
+++ b/connectonion/cli/browser_agent/client.py
@@ -2,8 +2,8 @@
 Purpose: Browser daemon client — sends short browser commands over the platform transport and runs the natural-language `do` agent locally so model waits never occupy the daemon.
 LLM-Note:
   Dependencies: imports from [socket, os, json, sys, time, shlex, inspect, functools, pathlib, browser_agent.transport | lazy: BrowserAutomation and browser_agent.agent for `do`] | imported by [cli/commands/browser_commands.py] | tested by [tests/e2e/cli/test_browser_daemon.py]
-  Data flow: direct verb → _request() → wire-v1 JSON {v,caller,account,tab,line,raw}; `do` → local Agent + DaemonBrowserProxy → each tool invocation becomes one short _request(raw=True) to the shared daemon
-  State/Effects: may spawn the daemon via `python -m connectonion.cli.browser_agent.daemon <sock> [--headless]` detached (transport.spawn_detached: start_new_session POSIX / DETACHED_PROCESS Windows), logging to ~/.co/browser.log | writes to stdout/stderr
+  Data flow: direct verb → _request() → wire-v1 JSON {v,caller,account,tab,line,raw}; Host request_target_as() selects an explicit private endpoint/profile/authkey/log without putting proxy credentials on argv; `do` → local Agent + DaemonBrowserProxy → each tool invocation becomes one short _request(raw=True) to the shared daemon
+  State/Effects: may spawn the daemon via `python -m connectonion.cli.browser_agent.daemon <sock> [--headless]` detached (transport.spawn_detached: start_new_session POSIX / DETACHED_PROCESS Windows), logging to ~/.co/browser.log or the explicit private target log | writes to stdout/stderr
   Integration: exposes _caller() -> str, send(line, headless=False, tab=None) -> int | FIRST-RUN AUTO-INSTALL: on the cold-start path (no daemon yet) AND when a warm daemon answers "No browser is installed for this user" (send retries once), a page-driving verb with no system Chrome triggers `python -m patchright install chromium` right in the user's terminal (chromium: per-user dir, never needs admin — the branded chrome channel runs a system installer) (_ensure_browser_ready) — `co browser` just works with zero setup commands; PAGELESS_VERBS (status/tab/close/...) never provision
   Performance: direct verbs import only the lightweight transport client, then make one connect + request/response; Agent/Playwright and browser tool schemas load only for `do` or inside the daemon | daemon spawn adds browser launch latency on first call | model thinking happens in the caller process and holds no daemon lane
   Errors: _connect() retries transient refusal/backpressure and never unlinks an endpoint whose recorded owner is alive; only a truly stale POSIX socket is removed | missing endpoint → spawn daemon and wait until ready or timeout | setup RuntimeErrors become one clean stderr line, never a traceback containing HMAC-path locals | daemon death, overload shedding, or disconnect mid-request becomes a clean exit 1
@@ -54,7 +54,7 @@ def _wait_for_pid_exit(pid: int, timeout: float = 15.0) -> bool:
     return True
 
 
-def _connect(sock_path: str):
+def _connect(sock_path: str, authkey_path: str | Path | None = None):
     """Connect to the daemon. Returns a live connection, or None if the daemon is
     genuinely gone. A daemon at its bounded connection cap can momentarily refuse —
     that is NOT a dead daemon, so retry
@@ -63,7 +63,7 @@ def _connect(sock_path: str):
     authenticated named-pipe client from `transport`. Raises RuntimeError only for
     an authkey mismatch against a live daemon (send() turns it into a clean line)."""
     if transport.IS_WINDOWS:
-        return _connect_windows(sock_path)
+        return _connect_windows(sock_path, authkey_path=authkey_path)
     return _connect_posix(sock_path)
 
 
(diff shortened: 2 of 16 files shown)


wu-changxing:
Final validation for head b9ff8c9b is green. The compatibility regression found by the exhaustive suite was fixed by preserving the historical two-argument local _spawn_daemon call when no private target is requested. Hosted evidence: full test suites pass on Python 3.10, 3.11, 3.12, and 3.13; all 12 remote-browser-egress cells pass across Linux, macOS, and Windows; macOS E2E, Windows E2E, Windows browser, lockfile, and blog gate pass. The PR remains draft and stacked on #1303. Navigation remains intentionally unavailable pending DD-056 native step 4. No merge, publish, or deployment was performed.

wu-changxing:
## Adversarial review: one dead switch (real egress gap), fixed at `082e3215`

Third review in this stack, same regime. Rebased onto main first (resolving the `remote-browser.md` rewrite — the private-runtime paragraph now sits in that page's Security section — and a duplicate `repr=False` on `ProxyEndpoint` that this PR and #1303 both added).

### The finding that mattered

`--force-webrtc-ip-handling-policy=disable_non_proxied_udp` **is not a switch Chrome defines.** Chrome 151 has `--webrtc-ip-handling-policy`. Verified two ways:

```
strings 'Google Chrome Framework' | grep -x force-webrtc-ip-handling-policy   -> 0 hits
strings 'Google Chrome Framework' | grep -x webrtc-ip-handling-policy         -> 1 hit
```

and behaviourally — same page, real Chrome, real proxy:

```
PR flag (--force-webrtc-ip-handling-policy): 1 candidate  [('udp','host')]
correct flag (--webrtc-ip-handling-policy) : 0 candidates []
no webrtc flag at all                      : 1 candidate  [('udp','host')]
```

The flag was byte-for-byte equivalent to passing nothing: a page could still gather a UDP host candidate and open direct UDP sockets to an IP literal, outside the gateway and outside `destination_policy`. Chromium ignores an unknown switch **in silence**, so a misspelling and a working control are indistinguishable from the outside.

The unit test asserted the misspelled string was in `args` — it pinned the defect.

**Root cause, and the fix that generalizes:** the invariants lived in one module's tuple while `BrowserLaunchPolicy` validated only arg uniqueness, the proxy shape, and `service_workers`. A policy with **zero** egress arguments constructed fine. `__post_init__` now requires all four egress switches and refuses a bare `--proxy-server`, with a parametrized test dropping each one in turn. A future misspelling fails at construction rather than in production.

### Two more

- **`service_workers="block"` is requested but not effective** on the shipped driver — measured, it registers/activates/controls a page exactly as `"allow"`. No egress widens (worker traffic still goes through the gateway) but the doc claimed a guarantee, and now states what was measured. Decision tracked in **#1325**; option 3 there (fold it into #1307's preflight, which already probes worker paths against a loopback sentinel) costs nothing extra.
- **`--profile-dir` was accepted and silently ignored without `--remote-egress`** — the profile only reaches the browser through the launch policy. A caller got its own socket, lock and log while pooling cookies with the ordinary local `co browser`. Now refused rather than ignored; verified the guard actually raises.

### Also: a vacuous test, replaced

`test_private_spawn_argv_and_log_never_contain_gateway_secret` asserted the literal `"gateway-secret"` was absent from an argv built by a code path that never constructs a gateway. It would pass on an implementation that put a real password on argv. It now starts a real `EgressGateway` and asserts **that gateway's actual minted credential** is absent from argv and from the log file.

### Verified clean (the review tried hard here)

- **Credentials**: not in argv (only `--profile-dir`/`--authkey-file`/`--remote-egress`; the password is minted inside the daemon process and never crosses a process boundary), not in three real Chrome profiles on disk (`Login Data`, `Preferences`, `Network Persistent State` all clean), not in a forced-launch-failure traceback, not in any repr. The strongest theory — patchright copying proxy credentials into unscoped `httpCredentials` — was chased empirically: the origin sees its 401 and no `Authorization` ever follows.
- **Ordering**: gateway up before browser, browser closed before gateway; after `gateway.stop()` both `go_to` and read-only `status` fail closed with `EGRESS_GATEWAY_UNAVAILABLE` before any tab mutation. The lost-bind race leaves no orphaned listener, and `_remove_endpoint` is pid-guarded.
- **The flags that do work**: with real Chrome, `127.0.0.1`, `localhost`, `[::1]` and `example.com` all reached the proxy with **zero** direct hits on a local origin server — including with `PLAYWRIGHT_DISABLE_FORCED_CHROMIUM_PROXIED_LOOPBACK=1` set. Resolver rules effective. Runtime dir `0700`, all paths distinct from the ordinary daemon's, no env reads in the private modules.
- **No local regression**: `_async_browser.py`'s refactor is keyword-for-keyword equivalent; 106 local browser-daemon tests pass.

7,111 unit tests pass; ruff clean.

PR #1280, opened 2026-08-25
feat: wire the 1.8 paid browser engine

wu-changxing:
## Problem

ConnectOnion main still has the obsolete Free/Pro attestation stub from the superseded #797 design. The real `co browser` launch seam cannot select or supervise the 1.8 PAYG Onion Browser described by #511. The first implementation also targeted the old synchronous browser driver, while the final 1.8 runtime in #1288–#1290 is async-owned.

## Approach

- stack this draft on #1290, after #1288 → #1289 → #1290
- add immutable `auto`, `system`, and strict `onion` engine resolution
- use Onionwright public `PaidSessionClient.prepare()` before billing and `launch_paid_async()` at the one charge boundary
- keep blocking preflight/start/release work off ConnectOnion's browser event loop
- preserve $0 system fallback only before a paid session starts; never hot-swap after billing
- make cancellation during charge or Chromium startup finish the irreversible operation and release any created session before propagating
- keep system and paid browser profiles separate
- pin Chromium `150.0.7871.187` and expose safe engine/artifact/session status without tokens, licence bytes, or paths
- propagate the engine choice through Typer, the client wire envelope, daemon startup, direct commands, and the local `do` proxy
- reject mode changes on a warm daemon and probe old daemons before an explicit Onion action can reach them
- document mode, cost, fallback, lifecycle, and the sync/async integration boundary

## Verification

- `7340 passed, 21 skipped, 200 deselected` in the fresh CI-equivalent environment; the two initially reported failures were isolated to local PATH selection and the workflow job-count assertion, then both passed after correction
- `108 passed` across browser status, engine CLI, and the native Unix-socket daemon suite
- `75 passed` focused engine, async core/facade, paid seam, and cross-repository tests
- real ConnectOnion sync facade → async core → Onionwright `launch_paid_async` contract proves exact executable/licence arguments and one close/release/driver-stop lifecycle
- local cross-repository test loads the matching Onionwright #43 source and runs that real boundary; ConnectOnion's hosted token cannot read the private sibling repository, so hosted cross-repo execution requires a separately scoped credential
- `git diff --check`
- `compileall` for changed packages and tests

## Dependency / release disposition

Draft until openonion/onionwright#43 is merged and its 0.0.11 wheel is released through the licence-gated artifact endpoint, and at least one exact browser artifact is certified and published under openonion/browser#96/#97. The paid catalogue remains empty until then. This PR must not by itself close #511; exact downloaded-artifact PAYG E2E remains required.

[...]

Issue #105, opened 2026-03-06
Add on_session_start and on_session_restore events


wu-changxing:
**Priority:** Medium

**Summary:** Add separate events for new session creation vs session restoration to enable different initialization logic.

**Motivation:** No way to distinguish between creating new conversation vs restoring existing one. These require different handling.

**API:**
```python
@on_session_start
def init_new_session(agent):
    agent.current_session['user_id'] = load_user_id()
    agent.current_session['preferences'] = load_preferences()

@on_session_restore
def restore_session(agent):
    if not validate_session(agent.current_session):
        raise ValueError('Invalid session')
    restore_plugin_state(agent.current_session)
```

**Implementation:** 
- on_session_start: agent.py line 243 (after new session created)
- on_session_restore: agent.py line 232 (after session restored)

**Use Cases:** Session initialization, state validation, plugin state restore

Issue #104, opened 2026-03-06
Add before_user_input event


wu-changxing:
**Priority:** High

**Summary:** Add `before_user_input` event that fires before user input is added to session, enabling input validation and sanitization.

**Motivation:** Currently `after_user_input` fires after message is already in session. No way to validate/reject input before processing.

**API:**
```python
@before_user_input
def validate_input(agent):
    prompt = agent.current_session.get('user_prompt')
    if len(prompt) > 10000:
        raise ValueError('Prompt too long')
```

**Implementation:** Add to agent.py line 251, fire before messages.append()

**Use Cases:** Input validation, content moderation, rate limiting, sanitization

PR #1307, opened 2026-08-26
feat: fail closed on unproven browser egress

wu-changxing:
## Problem

The private Remote Browser launch policy requests a loopback proxy, but Chromium launch success alone does not prove that the effective browser process used that gateway. Chromium can silently bypass loopback destinations, and an unauthenticated or system-browser launch must never make Remote Browser page commands available.

## Approach

- add a versioned native egress preflight to the private launch policy
- atomically write browser #105's canonical credential document inside the Host-private runtime root
- pass Playwright only the loopback proxy server and pass the paid browser only a path-only native credential switch
- remove the credential file on browser construction failure and after browser shutdown
- require an owned `.invalid` request to return the gateway DNS witness (502)
- require an owned loopback main-frame request to return the authenticated gateway denial (403)
- exercise fetch, image, WebSocket, and worker paths against a loopback sentinel
- require the sentinel to observe zero accepted sockets and zero bytes
- run the preflight after the real persistent context launches but before seed cookies or user pages
- collapse all uncertainty to one secret-free `EGRESS_PREFLIGHT_FAILED` error and tear down partial browser/driver state
- redact driver-launch failures before they can echo the private credential-file path

## Dependencies and intentional limits

This is a fail-closed checkpoint, not the final installed-artifact acceptance claim.

- Base: #1305 (Host-private runtime)
- Browser revision: openonion/browser#104
- Native scoped proxy credentials: openonion/browser#105
- Host-side credential-file production is included here and matches browser #105; native consumption still depends on that browser patch
- Final integration must stack with the paid Onionwright engine from #1280 and force `engine=onion`; system Chrome is not Remote Browser eligible
- The installed Mac/Linux paid-artifact E2E matrix is intentionally deferred until those stacks meet. No system-Chrome test is presented as equivalent evidence.

Remote Browser page commands remain unavailable until the credential-enabled paid artifact passes the real installed-artifact suite.

## Verification

- `/private/tmp/connectonion310-venv/bin/python -m pytest tests/unit/test_remote_browser_native_egress.py tests/unit/test_remote_browser_private_runtime.py -q` — 33 passed
- Ruff on every changed Python file — passed
- `python -m compileall` on every changed Python file — passed
- `git diff --check` — passed

## Follow-up acceptance matrix

The stacked paid-artifact PR must run in parallel on macOS arm64 and Linux x86_64 and prove positive public navigation, main-frame and redirect denial, iframe/image/script/style/font/media/fetch/XHR/EventSource/worker/WebSocket/popup/download coverage, concurrent allowed/denied tabs, gateway crash/restart behavior, and zero loopback-sentinel sockets/bytes throughout.

**Estimated release window:** after browser #104/#105 and the stacked paid-engine installed-artifact gate

### DIFF
--- a/connectonion/cli/browser_agent/daemon.py
+++ b/connectonion/cli/browser_agent/daemon.py
@@ -847,7 +847,9 @@ class BrowserDaemon:
                 proxy_auth_path,
             )
             browser = self._browser_factory(
-                headless=self._headless, launch_policy=policy
+                headless=self._headless,
+                launch_policy=policy,
+                egress_gateway=gateway,
             )
         except BaseException:
             remove_proxy_auth_file(proxy_auth_path)
--- a/connectonion/cli/browser_agent/daemon.py
+++ b/connectonion/cli/browser_agent/daemon.py
@@ -244,6 +244,7 @@ class BrowserDaemon:
         self._authkey_path = Path(authkey_path) if authkey_path is not None else None
         self._gateway_factory = gateway_factory
         self._browser_factory = browser_factory
+        self._proxy_auth_path = None
         if self._remote_egress and self._profile_dir is None:
             raise ValueError("remote browser daemon requires an explicit profile")
         if not self._remote_egress and self._profile_dir is not None:
@@ -825,7 +826,10 @@ class BrowserDaemon:
             raise RuntimeError("private browser runtime is only prepared once")
         from connectonion.network.host.egress_gateway import EgressGateway
         from connectonion.network.host.private_browser_runtime import (
+            proxy_auth_path_for_profile,
             remote_browser_launch_policy,
+            remove_proxy_auth_file,
+            write_proxy_auth_file,
         )
 
         gateway = (
@@ -834,15 +838,23 @@ class BrowserDaemon:
             else EgressGateway()
         )
         endpoint = await gateway.start()
+        proxy_auth_path = proxy_auth_path_for_profile(self._profile_dir)
         try:
-            policy = remote_browser_launch_policy(self._profile_dir, endpoint)
+            write_proxy_auth_file(proxy_auth_path, endpoint)
+            policy = remote_browser_launch_policy(
+                self._profile_dir,
+                endpoint,
+                proxy_auth_path,
+            )
             browser = self._browser_factory(
                 headless=self._headless, launch_policy=policy
             )
         except BaseException:
+            remove_proxy_auth_file(proxy_auth_path)
             await gateway.stop()
             raise
         self._gateway = gateway
+        self._proxy_auth_path = proxy_auth_path
         self.browser = browser
 
     def _accept_posix_client(self, reader, writer) -> None:
@@ -1081,6 +1093,12 @@ class BrowserDaemon:
                     f"browser cleanup failed: {type(exc).__name__}: {exc}",
                     file=sys.stderr,
                 )
+        from connectonion.network.host.private_browser_runtime import (
+            remove_proxy_auth_file,
+        )
+
+        remove_proxy_auth_file(self._proxy_auth_path)
+        self._proxy_auth_path = None
         if self._gateway is not None:
             try:
                 await self._gateway.stop()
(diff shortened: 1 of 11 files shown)


wu-changxing:
Green checkpoint for head `8ec08fac`:

- focused local private-runtime/native-preflight suite: 33 passed;
- Ruff, compileall, and diff checks: passed;
- blog, release metadata, and lockfile gates: passed;
- full Python suites: 3.10, 3.11, 3.12, and 3.13 passed;
- Remote Browser matrix: Ubuntu, macOS, and Windows on Python 3.10-3.13 all passed (12 jobs);
- macOS wheel E2E, Windows wheel E2E, and Windows browser jobs all passed.

Self-review also added launch-error redaction so a driver failure cannot echo the private credential-file path.

The PR remains draft and unmerged. This green result proves the Host producer/lifecycle/preflight contracts, not native browser consumption. Browser #105 compilation and the combined paid Onionwright + exact-artifact Mac/Linux E2E remain hard release gates.

wu-changxing:
## Adversarial review: the preflight could never pass. Every test was green. Fixed at `6f5b592b`

Rebased onto main first (picking up the merged #1303/#1305 fixes; this branch was stacked on their pre-fix commits).

### It had never worked once

**1. The `.invalid` witness required status 502. The real gateway answers 403.** The destination policy denies `.invalid` by name before DNS is consulted, so the DNS-failure path the witness expected is unreachable for that host:

```
goto http://remote-browser-preflight.invalid/ -> 403  X-ConnectOnion-Error: DESTINATION_HOST_DENIED
```

**2. Chromium will not commit a bodyless 4xx/5xx main-frame navigation.** The gateway sends every refusal with `Content-Length: 0`. Measured on Chrome 151:

```
403 empty body -> ERR_HTTP_RESPONSE_CODE_FAILURE
403 with body  -> status=403
502 empty body -> ERR_HTTP_RESPONSE_CODE_FAILURE
502 with body  -> status=502
```

So `page.goto` **raised** for both probes and collapsed to `EGRESS_PREFLIGHT_FAILED`. Fail-closed — the right direction to be broken in — but the private Remote Browser could never start. Refusals now carry a one-line body holding the stable code.

**3. The subresource probes could not fail.** They run inside a bounded race that swallows their own errors by design, then the sentinel's zero is read as proof. But zero is also what "the probe never ran" produces — a wrong URL, a blocked API, an `await` that resolves on the timer. Replacing the entire subresource function with a no-op left the preflight **passing**.

`EgressGateway` now exposes `handled_requests`, and the preflight requires that count to move. An absence is only evidence once the thing that would have produced a presence is known to have run. This also removes the dependence on the ~1.6 s window in which a late-landing leak was invisible.

**4. The witness is now the `X-ConnectOnion-Error` header**, not a bare status — a 403 from anything the browser could reach satisfied the old check. The status is still required to be 403, because a 407 carries the header too and means the credential never arrived.

### Why nothing caught it

`FakePage.goto` returns a response object where real Chromium raises, and its constructor hard-codes `witness_status=502` — the status the real gateway never sends. The fake modelled two behaviours the real stack does not have, and the suite verified the fake. **No test anywhere ran this preflight against a real browser.**

`tests/e2e/test_native_egress_preflight_real_driver.py` now does, in both directions.

### Every guarantee verified by breaking it

| mutation | result |
|---|---|
| remove the positive control | neutered-probe test goes **red** |
| remove the gateway-witness check | foreign-403 test goes **red** |
| restore the empty error body | real-driver test goes **red** |

And end to end against the real stack:

```
correct launch                      -> PREFLIGHT PASSED, gateway handled 7 requests
--proxy-bypass-list=127.0.0.1 (leak) -> PREFLIGHT FAILED, gateway handled 2
```

Passing alone would only prove the check is lenient; the leak case is what makes it mean something.

### Also confirmed by the review, unchanged

No path was found where the preflight reports proven egress while egress is open — the design is genuinely fail-closed, the sentinel detector works when it runs (a real bypass produced 6 sockets / 2847 bytes and was caught), the credential file is 0600 with an atomic write and uid+mode checks on its parent, ordering is correct (before seed cookies and before any user page, with no way to reach a page command that skipped it), and browser/driver teardown leaves no surviving Chromium.

Remaining review findings not addressed here, filed instead: the two-daemon credential-file race, and gateway/credential teardown on preflight failure.

7,133 unit tests pass; ruff clean on every touched file. Design Journal entry added.

Issue #1380, opened 2026-09-02
Forward-port 1.7.2 environment loading fixes to main / 1.8


wu-changxing:
The 1.7.2 stable patch aligns dotenv loading with the canonical bounded project root in package startup and Gmail, Outlook, GDrive, and Synology reloads. Main / 1.8 is the active higher line and requires the same code, regression tests, and runtime documentation. The previously merged canonical oo skill and pypdf lock fix are already on both lines (PR #1379 and main PRs #1364 / #1370). Version-only 1.7.2 metadata must not be copied to main. This ledger stays open until the focused main forward-port has merged with green CI. Stable and main PR links will be recorded here.

wu-changxing:
Stable fix: PR #1381, commit b36e294a. Main forward-port: PR #1382, commit a99a994b (only import-order conflict; main version untouched). Focused stable regression suite passed 217 tests; main selection passed 65; daemon socket regressions passed 89 outside the local sandbox. Cross-platform CI and installed-wheel verification are in progress.

wu-changxing:
Main forward-port #1382 merged after all checks passed at 3c8870f6cd6c30f2fb891faf3eb52372d41e6f8b. Stable fix #1381 and release metadata #1383 remain pending CI. Do not close the ledger until the stable release preparation is merged and final verification is recorded.

wu-changxing:
Both product paths have merged. Stable fix #1381 is included in release PR #1383 at 463a721fe194bfd3e3cfaefd7f64b32f38f59e6b; main #1382 merged at 3c8870f6cd6c30f2fb891faf3eb52372d41e6f8b. Final 1.7.2 candidate CI passed every required check (PR run 33604605948 and push run 33604599001), including Python 3.11 twice. The obsolete fix-only run 33603931721 had one stalled Python 3.11 worker and was cancelled after the independently tested final candidate merged. Publication verification will follow.

wu-changxing:
The original environment-root fix is merged to both main (#1382) and release/1.7 (#1383). The v1.7.2 tag is immutable at 463a721fe194bfd3e3cfaefd7f64b32f38f59e6b. Release run 33605199281 was cancelled before publish/finalize (both cancelled), following a user clarification: no-path co init must initialize global ~/.co/keys.env; project initialization requires an explicit path such as co init ./. A follow-up fix is being prepared for both lines. Do not resume the cancelled 1.7.2 publication or move the tag; a new release version awaits maintainer direction.

[...]

PR #1381, opened 2026-09-02
fix: keep project environment consistent from subdirectories

wu-changxing:
## Summary

Align dotenv loading with the canonical project root already used by identity, auth, and status. Previously, launching inside `project/src/` missed `project/.env` or selected `src/.env`; global fallback and third-party credentials could silently change with cwd. Remove the redundant CLI startup load and fix the four integration reload paths too. Preserve explicit process-variable precedence and the existing repository/home boundaries.

- **Suggested labels:** bug, tests, documentation
- **Proposed target version:** 1.7.2
- **Estimated release window:** next stable patch, September 2026
- **Forward-port tracking issue (stable patches only):** #1380

## Verification

- New regression suite reproduced 8 failures before the fix (2 boundary cases already passed); now 13 cases pass.
- 217 focused startup, import-cost, project-boundary, integration CLI, and auth tests passed.
- Credential safety/diagnostics selection: 49 passed before adding the last three boundary cases.
- Added an installed-wheel SDK and real co entry-point regression.
- Full suite and hosted cross-platform matrices must pass before release.
- git diff --check passed.

## Documentation and review

Updated docs/cli/README.md and docs/blog/2026-09-02-one-project-one-environment.md. The docs-site checkout is absent locally; the live site has not been changed.

AI-assisted using Codex. The principal compatibility risk is a project intentionally relying on an unconfigured subdirectory .env; it should use the canonical project file, an explicit process variable, or a separately configured nested project. No live provider credentials or mailbox calls were used for these tests. Account-mismatch checks, token persistence, and browser behavior are unchanged.

### DIFF
--- a/connectonion/__init__.py
+++ b/connectonion/__init__.py
@@ -2,10 +2,10 @@
 Purpose: Main package entry point exposing public API for ConnectOnion framework
 LLM-Note:
   Dependencies: imports from [core/, logger.py, llm_do.py, transcribe.py, prompts.py, debug/, useful_tools/, network/, address.py] | imported by [user code, tests/, examples/] | no direct tests (integration tests import from here)
-  Data flow: loads cwd .env then ~/.co/keys.env via load_dotenv() (first file to define a key wins) → exports all public API symbols → user imports `from connectonion import Agent, llm_do, ...`
-  State/Effects: auto-loads both the cwd .env (NOT module directory) and the global ~/.co/keys.env at import time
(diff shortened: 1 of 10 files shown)


wu-changxing:
Final local product verification: 6,722 passed, 15 skipped, 17 deselected in the full unit suite; 89 daemon tests pass outside the sandbox. The built 1.7.2 candidate was installed with its declared dependencies into an isolated venv and passed all 9 original wheel acceptance assertions, including the new real co + SDK nested-directory regression. Local tests use fake credentials only.

Issue #1391, opened 2026-09-02
OIP: rendezvous for two agents both behind NAT (parked for 2.5)


wu-changxing:
## Parked design note (target: 2.5, with the decentralisation work)

Today a direct OIP socket to a public address must be https/wss (`endpoint_is_safe`), because a signed CONNECT captured on a plaintext link can be replayed within its freshness window (#649). The cost lands on every self-hosted host: a domain, a certificate, Caddy, ports 80 and 443. The Melbourne rental host hit exactly this on 2026-09-02 — it announced `http://34.129.161.131:8001`, the 1.8.0b1 laptop refused it, and the fix was to provision it like a `co deploy` host.

The alternative is to make the direct channel itself safe instead of borrowing TLS:

1. both sides exchange one-time ephemeral public keys through the relay, each signed by its long-term agent key (the relay forwards, cannot read or alter);
2. the direct socket opens with a Noise-style handshake (long-term identity + ephemerals); every later frame is end-to-end encrypted and sequenced;
3. a captured frame can neither be read nor replayed — the replay problem is solved in the protocol, and the "public endpoints must be https" rule can go. A host then announces `ws://IP:port` and needs one open port, nothing else.

What it does not change: the host still needs an inbound port. Both-ends-behind-NAT still means relaying bytes, which is a separate cost decision.

## Decision (Aaron, 2026-09-02)

Not now. Until 2.5 the relay stays a control-frame forwarder only, and direct connections keep requiring https. Revisit when the decentralised forwarding layer is designed, since this handshake is the same primitive that layer needs.

wu-changxing:
Update: the direct-channel half shipped in 1.8.0 (#1393) — `SEAL`/`SEALED_OK` with one-time X25519 keys signed by both Ed25519 identities, no relay involvement, https requirement dropped. Aaron reversed the 'park it' decision the same afternoon once it was clear the address already is the key and oo-api needs no change. What stays parked for 2.5 is only the both-behind-NAT case (a rendezvous that splices TCP). Retitling the scope of this issue to that.

Issue #1387, opened 2026-09-02
co proxy diagnose should name the endpoint it could not reach


wu-changxing:
## What happened

`co proxy share to <host>` on a laptop reported only:

```
reconnecting (the host is not reachable directly; retrying in 60s)
```

and `co proxy diagnose <host>` repeated the same sentence plus "the host has to be reachable directly, not through the relay". Nothing said *which* address the share had tried.

The cause took a manual query of the relay to find: `GET /api/agents/<host>` showed the host announcing `http://34.129.161.131:8001` (no `AGENT_PUBLIC_DOMAIN`, so it publishes IP:port), and GCP's firewall for that instance allows only 22/80/443. `curl` to `:8001` from the laptop timed out. Seen 2026-09-02 on the Melbourne rental host during the 1.8.0b1 two-machine run (#1036).

## What diagnose should print

- the endpoints the relay currently lists for the host, verbatim;
- one probe result per endpoint (`connect timeout`, `refused`, `answered /info, address mismatch`, `ok`);
- when every public endpoint times out, say so in words: "the host announces `34.129.161.131:8001` but nothing answers there from this network — usually a cloud firewall; a `co deploy --to` host answers on 443".

`connectonion/network/connect.py::resolve_endpoint` already walks this list and probes each one; diagnose only needs to surface what it saw instead of discarding it.

**Proposed target version:** 1.8.0b2
**Estimated release window:** this week

Issue #1036, opened 2026-08-15
[1.8 Epic] Laptop Proxy for Remote WTF Browser


wu-changxing:
# 1.8 Laptop Proxy for Remote WTF Browser

## First-preview decision

The first Proxy product is a deliberate two-machine workflow:

```text
Laptop: co proxy share to <remote-agent>
          │ local DNS + final public socket
          ▼
Remote server: loopback gateway → WTF Browser
```

WTF Browser owns profile/cookies/page state on the remote server. The Laptop owns target DNS resolution and final public egress, so sites see the Laptop IP rather than the datacenter IP. BrowserDaemon is only a controller and cannot reinterpret or silently change the proxy choice.

Earlier three-party/unattended proxy designs are future work, not a requirement for this preview.

## CLI and ownership

```bash
co proxy share to <remote-agent> [--bind HOST] [--ttl SEC]
co proxy status
co proxy diagnose <remote-agent>
co proxy stop <remote-agent>

co remote-browser <remote-agent> start --proxy shared
co remote-browser <remote-agent> start --proxy direct
```

- Proxy mode is resolved once at runtime/session creation and written to private mode-0600 state.
- WTF Browser receives one authenticated loopback gateway with no `DIRECT` fallback or runtime switching.
- All browser-controlled target DNS belongs to the Laptop. Both ends reject private, loopback, link-local, metadata, reserved and rebinding targets.
- Proxy loss stops/fails network access; it never changes the public IP to the remote server.
- HTTPS remains end-to-end between WTF Browser and the destination; the Laptop observes destination metadata/timing/byte counts because it owns the exit.
- The current authenticated listener needs an existing protected VPN/SSH tunnel or equivalent across untrusted networks. Automatic encrypted reverse transport/NAT traversal is future work; no plaintext Relay downgrade is allowed.

## Implementation PR

Draft PR #1362 implements the first local/runtime path.

Current state on 2026-09-02:

- the prior head `a62914c62ed03113cbf40ea5c09e523f00f77ddc` completed 25 GitHub checks with no failures;
- a newer head, `eca1cda9cd4e7c421301a8c1f64d91bb3e8937ab`, changed the real-driver preflight to headless and triggered a fresh matrix;
- the PR is restored to Draft, is mergeable, but is currently `UNSTABLE` because the required lockfile job failed in run `33568073057`; the remaining platform jobs are passing or still completing;
- 351 focused Proxy/Remote Browser security/lifecycle tests;
[...]

PR #1393, opened 2026-09-02
network: seal direct sockets end to end — a host needs a port, not a certificate

wu-changxing:
## Problem

`endpoint_is_safe` refuses any plaintext public endpoint, because a signed CONNECT captured on such a link can be replayed within its five-minute window (#649). So a direct connection needed TLS, i.e. every self-hosted agent needed a domain, a certificate and Caddy. Today (2026-09-02) the Melbourne rental host announced `http://34.129.161.131:8001`; with the port open and a bare WebSocket connecting fine, the 1.8.0b1 laptop still reported `not reachable directly` and the Laptop Proxy could not attach (#1036, #1387).

## Change

- `connectonion/network/sealed.py`: `SEAL` / `SEALED_OK` handshake — one-time X25519 keys signed by each side's long-term Ed25519 identity (the address *is* the key, so no relay involvement) — then every frame both ways as `SEALED {n, c}` in a NaCl Box with a per-direction counter. Captured frames are unreadable; replayed or reordered ones fail to open and end the session.
- Client (`connect.py::_open_best_connection`): every direct socket is offered a SEAL. A host that answers is used, plaintext or not. A host that does not is used bare only if the link is already private (TLS/loopback) or the client has no keys (nothing signed to lose); otherwise the socket is closed and the client moves on to the relay. `resolve_endpoint` no longer filters plaintext candidates.
- Host (`asgi/websocket.py::_maybe_seal`): if the first frame is SEAL, answer and wrap send/recv; a bad SEAL gets `ERROR seal refused` and close 4003 — no plaintext second try. Any other first frame is passed through unchanged, so 1.7 clients are unaffected. `route_handlers["identity"]` carries the host's keys (only the `host()` path; the bare `create_app` path has no signing key and says so).
- Relay path untouched. `PROXY_ATTACH` still requires a direct socket; a sealed plaintext one qualifies.
- Docs: protocol reference section, SELLING_POINTS row corrected (direct sockets are E2E now; relay path is not). Design Journal: `docs/blog/2026-09-02-the-port-that-was-open-and-still-refused.md`.

## Tests

`tests/unit/test_a_direct_socket_is_sealed.py` (13): handshake both ways, replay of a sealed frame refused, a stranger finishing a captured SEAL gets nothing, stale / misaddressed / wrong-signer SEAL refused, SEALED_OK from the wrong host or for a different SEAL refused; client uses a public plaintext endpoint once sealed and **never sends a signed CONNECT in the clear to a host that cannot seal**; host answers SEAL through the real ASGI adapter and CONNECT→CONNECTED travel sealed; bad SEAL closes 4003; unsealed 1.7 client still gets through.

The two client-integration tests were run against the unpatched `connect.py`/`websocket.py` first: 2 failed, then green. Two endpoint-preference tests in `test_a_captured_connect_opens_nothing_twice.py` were rewritten to the new contract (a LAN plaintext endpoint is a candidate; closeness wins now that plaintext is sealed).

Full unit suite locally: 7294 passed, 13 skipped.

## Decision record

Aaron, 2026-09-02: relay stays a control-frame forwarder; do this in the protocol now rather than provisioning https on every host. Supersedes the "parked for 2.5" note in #1391 for the direct-channel part; the both-behind-NAT rendezvous stays 2.5.

**Proposed target version:** 1.8.0b2
**Estimated release window:** today, after the Parrot ↔ Melbourne run passes on this build

https://claude.ai/code/session_01YE2E7YwTvLnPp2qjfcofRq


wu-changxing:
Live two-machine run on this branch passed: Parrot laptop → Melbourne host over plaintext `ws://34.129.161.131:8001` sealed end to end, no TLS/domain/Caddy. Both IP-echo sites saw the laptop's address; `co proxy stop` → `ERR_TUNNEL_CONNECTION_FAILED`. Details and the two findings (one fixed here: `SealedSocket.__aiter__`) in #1036.

Issue #804, opened 2026-08-10
Make one-use signatures atomic across ASGI workers


wu-changxing:
## Security problem

ConnectOnion documents and implements one-use signatures for WebSocket CONNECT, protocol-v2 commands, signed admin calls, and audience-scoped HTTP routes. The current replay ledger is the module-level `_seen_signatures` dictionary in `network/host/auth.py`.

That dictionary is isolated per OS process. `create_app()` explicitly supports `uvicorn ... --workers 4`, so the same captured signature can be accepted once by each worker during its five-minute freshness window. The signature is valid and route/recipient-bound; the missing property is one-use atomicity across workers.

## Existing design decisions

- [DD-007](https://github.com/openonion/connectonion/blob/main/docs/design-decisions/007-authentication-backend-design.md) keeps authentication to one client-signed request and uses recency rather than a server-issued challenge. This fix preserves that public protocol; it only makes the already-promised one-use property real across processes.
- [DD-020](https://github.com/openonion/connectonion/blob/main/docs/design-decisions/020-trust-system-and-network-architecture.md) says security-critical limits belong in deterministic code. Replay exclusion cannot depend on which worker receives a request.
- [DD-022](https://github.com/openonion/connectonion/blob/main/docs/design-decisions/022-raw-asgi-implementation.md) keeps raw ASGI and external deployment flexibility. The guard must therefore work for the documented multi-worker ASGI deployment, not only `host()`'s in-process server.

## Options considered

1. **Keep the process-local dictionary and document `workers=1`.** Rejected: external `create_app()` multi-worker deployment is documented and already coordinates other state across workers. Silently weakening a security claim is not a safe compatibility policy.
2. **Require Redis or another network service.** Rejected for the default: adds setup, a dependency, and a new availability boundary to a local-first framework. A future distributed/multi-machine host may need an injected external store, but one machine with multiple workers does not.
3. **One file per signature with `O_EXCL`.** Atomic, but creates and scans many files, needs stale-file race handling, and turns request volume into directory growth.
4. **Project-local SQLite ledger with a unique signature digest.** Selected: Python stdlib only, atomic across threads/processes, bounded by the same five-minute TTL, and stored beside the hosted agent's existing `.co` state.

## Proposed contract

- Add a small replay store backed by `.co/replay.sqlite3` (or beside an explicitly supplied `SessionStorage` path for `create_app`).
- Store only `SHA-256(signature)` plus `seen_at`; never persist the raw signed payload or signature.
[...]

PR #1404, opened 2026-09-03
fix(auth): rebuild a replay ledger deleted under a running host

### LINKED ISSUE #1401: Recover from an unusable replay store instead of failing closed forever
## Problem

A **zero-byte** `.co/replay.sqlite3` takes a hosted agent down indefinitely, and nothing in the framework reports it or recovers from it.

On 2026-09-03 a production agent (`1.8.0b1`) refused **every signed call for 2h45m** — 213 consecutive `misconfigured: replay protection unavailable` errors — because a deploy had deleted and recreated the store as an empty file. The fail-closed behaviour is correct and exactly what #804 asked for. What is missing is everything after it: the agent never recovered on its own, never surfaced the fault where an operator would look, and gave every outward sign of being healthy.

The empty file was not corruption in the SQLite sense. `birth == mtime == ctime`, so it was delete-then-create — the shape any `rsync --delete` without a protect rule produces. `SignatureReplayStore.__init__` runs `CREATE TABLE IF NOT EXISTS`, which succeeds against a 0-byte file and leaves it empty; the failure appears later, per request, forever.

## Why it is hard to diagnose

The agent looked fine from every angle an operator normally checks:

| Signal | State during the outage |
|---|---|
| `/health` | **HTTP 200** |
| `/info` | reachable |
| relay heartbeat | connected |
| `systemctl is-active` | `active` |
| Postgres + scheduled crawl | fully healthy, jobs completing on time |
| **every signed call** | **rejected** |

Only paths crossing signature verification were dead. Cron jobs writing straight to the database kept working, so dashboards, data freshness and uptime checks all stayed green. This is a half-hung agent: live enough to pass every liveness probe, dead for every remote command. The 213 errors existed only in `journalctl`, which is the one place nobody looks when the health endpoint is 200.

Recovery, once diagnosed, is three commands — stop, `rm` the file, start — and the framework rebuilds the schema. Nothing about the error message points there.

## Proposed fix

**1. Self-heal an unusable store, once, loudly.**
On init, validate the store (`PRAGMA integrity_check`, confirm `used_signatures` exists and the file is non-empty). If it is unusable, move it aside to `replay.sqlite3.corrupt-<timestamp>` and recreate. This is safe in a way most data recovery is not: the store holds only signature digests and timestamps with a 5-minute TTL, so a rebuild costs at most one expiry window of replay history — never business data. Log it at WARN with the path of the preserved file.

Bounded so it cannot mask a real attack or flap: at most one recovery per process start, and if the store is unusable *again* after recovery, fail closed permanently and stay down.

**2. Make the fault visible where operators actually look.**
`/health` returning 200 while every signed call is rejected is the core reporting bug. Health should reflect replay-store status — either a non-200, or a documented `degraded` field naming the subsystem. Also worth having `co status` surface it, since that is the first thing anyone runs.

**3. Make the error actionable.**
`misconfigured: replay protection unavailable` says nothing about which file, why, or what to do. Include the path and the concrete cause (missing / zero-byte / integrity failure / permission / locked), and — given the store is disposable — the recovery command itself.

**4. Protect the file from the outside.**
`co deploy` already does the right thing via `--filter "P .co/**"` (its comment explicitly covers "anything a future version writes there"), and that is *why* `co deploy` did not cause this — a hand-rolled rsync fallback did, having enumerated excludes instead. Worth documenting `replay.sqlite3` as server-owned runtime state that must never ship from a laptop or be deleted by a sync, so anyone writing deployment tooling knows before they learn it the hard way.

## Acceptance criteria

- [ ] A zero-byte `replay.sqlite3` is detected at init, moved aside, and recreated; the agent serves signed calls normally afterwards.
- [ ] Same for a truncated/garbage file that fails `PRAGMA integrity_check`, and for a valid-but-schemaless database.
- [ ] Recovery happens at most once per process; a second failure fails closed and stays closed.
- [ ] A permission error (unreadable/unwritable path) still fails closed — it is not a corruption case and must not be silently "recovered" by recreating the file elsewhere.
- [ ] The preserved `.corrupt-<timestamp>` file is retained for forensics and its path is logged.
- [ ] Health reporting no longer returns a plain 200 while replay protection is down.
- [ ] The error text names the file, the specific cause, and the recovery step.
- [ ] Replay semantics are unchanged in the healthy path: cross-worker atomicity from #804 and the 5-minute TTL still hold, and no raw signatures are persisted.

## Context

- Introduced by #804 / #805 (the fail-closed contract itself is right — this is about what happens next).
- #804's acceptance list included "Ledger corruption/unavailability fails closed and is covered by a test" but did not consider recovery, so an unusable store is a permanent outage by design.
- Observed on `1.8.0b1`; `replay.py` and `auth.py` are byte-identical in the `1.8.0` release, so upgrading does not change this. Suggesting **1.8.1**.

wu-changxing:
## Description

A hosted agent refused **every signed call for 2h45m** — 213 consecutive `misconfigured: replay protection unavailable` errors — after a deploy replaced `.co/replay.sqlite3` while the process was running. This rebuilds the ledger when it is found missing at claim time, instead of failing closed until a human restarts the service.

## Related Issue

Fixes #1401

## Proposed target version

**1.8.1** (stable patch)

## Estimated release window

Next 1.8.x patch — whenever maintainers cut one. No urgency from my side beyond the obvious: any host on 1.8.0b1/1.8.0 whose deploy tooling touches `.co/` can hit this, and the failure mode is a silent full outage.

## How this was written

- [ ] I wrote this myself
- [ ] Mostly AI-generated

**Which model?** Claude Opus 4.5, via Claude Code.

`_is_missing_ledger` matching on the string `"no such table"`. That is SQLite's message text, not a stable API, and it is locale-independent today but nothing guarantees it stays that way. I chose it over `sqlite3.OperationalError` alone because that class also covers `database is locked`, which must *not* trigger a rebuild — but a narrower, less stringly-typed discriminator would be better if one exists. I did not find one.

Second: the security tradeoff is real and deserves a maintainer's judgement rather than mine. A rebuild forgets prior signatures, so a captured signature can be replayed until it expires (≤300s). I believe that is strictly better than a total outage, and it only occurs in a situation where the ledger was already destroyed — but it *is* a weakening of the property #804 established, and reasonable people could want it to fail closed instead.

**What did you not verify — which paths did you never actually run?**

- Windows. The `os.name != "nt"` branch is untouched but I ran nothing on Windows.
- The genuinely-concurrent case: two workers both hitting the missing table and both rebuilding at once. `_prepare_schema` uses `BEGIN IMMEDIATE` and `CREATE TABLE IF NOT EXISTS`, so I expect one to serialize behind the other, but I did not write a multi-process test for that specific race. The existing `test_workers_serialize_old_schema_migration` covers the analogous migration path.
- Real relay traffic against a patched host. I reproduced the failure and the recovery in-process, not end to end over the network.

If `_is_missing_ledger` matched too broadly, a ledger with live entries would be silently discarded and replay protection would degrade without anyone noticing — the same class of silent-weakening this code is supposed to prevent. That is why the predicate is deliberately narrow and why `test_a_locked_ledger_is_never_rebuilt` asserts a prior claim still catches a replay afterwards.

## Type of Change

## Design Journal Impact

- [x] Release notes only — maintenance change with no reusable design lesson

> The lesson here is really about deployment tooling, not framework design: #804's fail-closed contract is correct and unchanged. This only adds recovery for one specific way the file disappears. Happy to write it up if a maintainer disagrees.

## Changes Made

- Extract the schema setup from `__init__` into `_prepare_schema()`, verbatim, so it can be reused.
- Extract the claim transaction into `_claim()`.
- `already_used()` now catches a missing-table error, rebuilds once, and retries. A second failure raises `unavailable after rebuild` and still fails closed.
- `_is_missing_ledger()` restricts recovery to `no such table` — locks, I/O errors and corruption keep failing closed.
- Log at WARNING on rebuild, naming the path, the likely cause, and the replay window.
- No new dependencies.

## Testing

```
$ .venv/bin/python -m pytest tests/unit/test_cross_worker_replay.py tests/unit/test_host_auth.py tests/unit/test_an_admin_signature_is_used_once.py tests/unit/test_a_captured_connect_opens_nothing_twice.py -q --timeout=120

89 passed in 14.84s
```

Replay suite alone, three consecutive runs (the suite is randomized):

```
$ for i in 1 2 3; do .venv/bin/python -m pytest tests/unit/test_cross_worker_replay.py -q; done

25 passed in 19.34s
25 passed in 20.51s
25 passed in 20.51s
```

The production scenario, reproduced end to end:

```
$ python - <<'PY'
store = SignatureReplayStore(p); store.already_used({"signature":"a"})
os.remove(p); p.write_bytes(b"")          # what the deploy did
print(store.already_used({"signature":"b"}), store.already_used({"signature":"b"}))
PY

[WARNING] Replay ledger at /tmp/.../replay.sqlite3 was missing its schema
(no such table: used_signatures) and has been rebuilt. It was most likely
deleted or replaced by a deploy; exclude it from any sync that owns /tmp/... .
Signatures recorded before the rebuild are forgotten, so a captured signature
can be replayed until it expires (300s).
False True
```

Recovers, and still rejects the replay.

Full `tests/unit` is green. Counts vary between runs because the suite is randomized and has some pre-existing order-dependent flakes; none are in replay or auth, and my five tests passed in every run.

## Scope

Two files, one concern:

- `connectonion/network/host/replay.py` — the fix. The two extractions are pure moves; the only new behaviour is in `already_used`.
- `tests/unit/test_cross_worker_replay.py` — five tests: the production scenario, that a rebuild restores the *full* schema (a ledger missing its expiry index stops pruning and grows unbounded), that a locked ledger is never rebuilt, that corruption still fails closed, and that rebuild is not retried forever.

## Additional Notes

**The empty file alone was not the bug.** `CREATE TABLE IF NOT EXISTS` succeeds against a zero-byte file, so construction repairs it — I verified this on `main`:

```python
p.write_bytes(b"")
store = SignatureReplayStore(p)          # succeeds, file becomes 12288 bytes
store.already_used({"signature": "x"})   # False — works fine
```

The outage required the file to be swapped *after* a healthy start, so init-time validation would not have prevented it. I initially diagnosed this wrong and corrected it in [a comment on #1401](https://github.com/openonion/connectonion/issues/1401#issuecomment-5521506711) — worth reading before reviewing, since the first half of that issue describes the wrong mechanism.

**Not in this PR.** #1401 also raises that `/health` returns 200 while every signed call is rejected — that is what made this take hours to find — and that the error text names neither the file nor a recovery step. Both are worth fixing but touch other modules.

**The deployment side is ours, not the framework's.** `co deploy` already carries `--filter "P .co/**"` and would not have caused this; a hand-rolled rsync fallback did, having enumerated excludes instead. Already fixed on our side.

### DIFF
--- a/connectonion/network/host/replay.py
+++ b/connectonion/network/host/replay.py
@@ -6,12 +6,16 @@ Errors: raises ReplayProtectionError so callers can fail closed with a safe mess
 """
 
 import hashlib
+import logging
 import math
 import os
 import sqlite3
 import time
 from contextlib import closing
 from pathlib import Path
+
+
+logger = logging.getLogger(__name__)
 
 SIGNATURE_EXPIRY_SECONDS = 300
 # A healthy transaction is tiny, but another OS worker can be descheduled while
@@ -43,45 +47,49 @@ class SignatureReplayStore:
         self.expiry_seconds = expiry_seconds
         self.path.parent.mkdir(parents=True, exist_ok=True)
         try:
-            descriptor = os.open(self.path, os.O_CREAT | os.O_WRONLY, 0o600)
-            os.close(descriptor)
-            if os.name != "nt":
-                os.chmod(self.path, 0o600)
-            with closing(self._connect()) as database:
-                with database:
-                    # Serialize schema inspection and migration across workers.
-                    database.execute("BEGIN IMMEDIATE")
-                    database.execute(
-                        "CREATE TABLE IF NOT EXISTS used_signatures ("
-                        "digest BLOB PRIMARY KEY, seen_at REAL NOT NULL"
-                        ", expires_at REAL"
-                        ") WITHOUT ROWID"
-                    )
-                    columns = {
-                        row[1] for row in database.execute(
-                            "PRAGMA table_info(used_signatures)"
-                        )
-                    }
-                    if "expires_at" not in columns:
-                        database.execute(
-                            "ALTER TABLE used_signatures "
-                            "ADD COLUMN expires_at REAL"
-                        )
-                    # A pre-fix row may have represented a future-dated
-                    # signature. Retain it for the maximum validity window.
-                    database.execute(
-                        "UPDATE used_signatures SET expires_at = seen_at + ? "
-                        "WHERE expires_at IS NULL",
-                        (2 * self.expiry_seconds,),
-                    )
-                    database.execute(
-                        "CREATE INDEX IF NOT EXISTS used_signatures_expiry "
-                        "ON used_signatures(expires_at)"
-                    )
+            self._prepare_schema()
         except (OSError, sqlite3.Error) as exc:
             raise ReplayProtectionError(
                 f"replay protection storage is unavailable: {self.path}"
             ) from exc
+
+    def _prepare_schema(self) -> None:
+        """Create the ledger, or migrate an older one, at the current path."""
+        descriptor = os.open(self.path, os.O_CREAT | os.O_WRONLY, 0o600)
+        os.close(descriptor)
+        if os.name != "nt":
+            os.chmod(self.path, 0o600)
+        with closing(self._connect()) as database:
+            with database:
+                # Serialize schema inspection and migration across workers.
+                database.execute("BEGIN IMMEDIATE")
+                database.execute(
+                    "CREATE TABLE IF NOT EXISTS used_signatures ("
+                    "digest BLOB PRIMARY KEY, seen_at REAL NOT NULL"
+                    ", expires_at REAL"
+                    ") WITHOUT ROWID"
+                )
+                columns = {
+                    row[1] for row in database.execute(
+                        "PRAGMA table_info(used_signatures)"
+                    )
+                }
+                if "expires_at" not in columns:
+                    database.execute(
+                        "ALTER TABLE used_signatures "
+                        "ADD COLUMN expires_at REAL"
+                    )
+                # A pre-fix row may have represented a future-dated
+                # signature. Retain it for the maximum validity window.
+                database.execute(
+                    "UPDATE used_signatures SET expires_at = seen_at + ? "
+                    "WHERE expires_at IS NULL",
+                    (2 * self.expiry_seconds,),
+                )
+                database.execute(
+                    "CREATE INDEX IF NOT EXISTS used_signatures_expiry "
+                    "ON used_signatures(expires_at)"
+                )
 
     def _connect(self):
         database = sqlite3.connect(self.path, timeout=SQLITE_BUSY_TIMEOUT_SECONDS)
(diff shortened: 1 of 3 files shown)


wu-changxing:
Closing as superseded by #1406, which merged as the 1.8.1 release commit and carries the same fix for `create_app()` hosts: `SignatureReplayStore` recreates its schema on `no such table` and retries the claim once, other storage errors still fail closed, and the SQLite error is printed beside the client-facing string. For `co host` / `co deploy` hosts the ledger moved to memory, so the file cannot be deleted under them at all.

Two things from here are still worth carrying over in a follow-up rather than lost:

- the five regression tests in `tests/unit/test_cross_worker_replay.py` (full-schema rebuild including the expiry index, locked ledger never rebuilt, corruption still fails closed, rebuild not retried forever);
- the two acceptance items #1401 raised that neither PR addresses: `/health` returning 200 while every signed call is refused, and the error text naming the file and the recovery step. Those are 1.8.2 candidates.

---

PR #1413, opened 2026-09-04
VERSIONING.md records 1.7.3, the release 1.7 users actually received

### LINKED ISSUE #1405: test_every_release_has_an_entry fails on main: v1.7.3 was released without a VERSIONING.md entry
`tests/unit/test_every_release_has_an_entry.py` fails on current `main` (2bfd6038, the 1.8.0 release commit):

```
AssertionError: released with no entry in VERSIONING.md: ['1.7.3']
```

GitHub has a `v1.7.3` release dated 2026-09-02 10:20 UTC (between 1.8.0b1 and 1.8.0), but `VERSIONING.md` on `main` has no `- 1.7.3 (...)` line — the 1.7.x entries stop at 1.7.2. Either the release notes for 1.7.3 need forward-porting from `release/1.7` into `main`'s VERSIONING.md, or the release was cut without one.

Found while running the full unit suite for #1402; unrelated to that change, so parked here rather than fixed in that PR.

wu-changxing:
Closes #1405.

`v1.7.3` is tagged and published on PyPI (2026-09-02), but `VERSIONING.md` has no entry for it, so `test_every_released_tag_is_documented` fails on `main`:

```
AssertionError: released with no entry in VERSIONING.md: ['1.7.3']
```

The release checklist runs the suite, so this blocks any release cut, 1.8.1 included.

The entry describes what actually shipped (#1384): plain `co init` creates the global identity and `~/.co/keys.env` without writing project files, a project is set up only when a path is given explicitly, project-only options without a path fail before writing, `--key` saves a global credential explicitly instead of promoting one a project or the environment supplied, and offline setup keeps the identity and restores a missing `keys.env` without rotating keys.

It also records why 1.7.3 rather than 1.7.2 is the version 1.7 users have: `v1.7.2` was tagged but its publication run was cancelled, and 1.7.2 is not on PyPI. Verified against the PyPI release index, not inferred from the tag list.

Verification: the test fails on the unpatched tree with exactly the assertion above, and passes with this entry (4 passed).

https://claude.ai/code/session_01KmCtowqB3DqGVPpTb1VU3S

**Proposed target version:** 1.8.1a1 / next preview
**Estimated release window:** the 1.8.1a1 preview once #1405 is clear, then 1.8.1 stable after it is exercised


wu-changxing:
Superseded by #1412, which landed the 1.7.3 entry on main while this was open. Verified on current main: test_every_release_has_an_entry passes (4 passed), so #1405 is resolved and this PR would only conflict. Closing rather than rebasing.

PR #1417, opened 2026-09-05
fix(outlook): scheduled replies become a deferred reply draft, not a reply action

### LINKED ISSUE #1416: co outlook reply --at still goes through the /reply action; docs still say scheduling needs only Mail.Send
Follow-up to #1198 / #1221.

#1221 fixed `Outlook.send(send_at=...)`: a scheduled send now creates a draft (`POST /me/messages`) carrying `PidTagDeferredSendTime` (0x3FEF) and `PidTagDeferredDeliveryTime` (0x000F) and then submits that draft with `POST /me/messages/{id}/send`. It deliberately left two things alone.

### 1. `reply(send_at=...)` is still on the same class of endpoint

`connectonion/useful_tools/outlook.py::reply` still posts to `/me/messages/{id}/reply` with `singleValueExtendedProperties: [0x3FEF]` inside `message`. The `/reply` action is create-and-send in one step, exactly like `sendMail` was for `send()`. Graph's extended-property docs list `POST /me/messages`, `POST /me/mailFolders/{id}/messages` and `PATCH /me/messages/{id}` as the ways to attach an extended property to a message; neither `sendMail` nor the `reply` action is among them. So `co outlook reply <#> ... --at ...` most likely has the same "delivered now and a deferred copy left in Drafts" behaviour that #1198 reported for `send`.

Not reproduced live yet. Please verify on a real mailbox before changing anything, the same way #1221 was validated (message absent from the recipient's inbox before the time, present exactly once after; cancel deletes the draft).

Expected fix shape, mirroring #1221:

```
POST /me/messages/{id}/createReply            -> draft reply (keeps threading)
PATCH /me/messages/{draft_id}                 -> body + singleValueExtendedProperties 0x3FEF, 0x000F (+ attachments)
POST /me/messages/{draft_id}/send
```

Requires `Mail.ReadWrite`, which `co auth microsoft` already requests (see `docs/integrations/microsoft.md`). Add a `_require_scope("Mail.ReadWrite")` guard on the scheduled branch like `send()` has.

### 2. `docs/cli/outlook.md` scope statement is now wrong

`docs/cli/outlook.md` line 162-163 says:

> **Scheduling** — Exchange holds delivery until the time you give (deferred send), so it works with just the `Mail.Send` scope and no extra setup

After #1221 a scheduled send needs `Mail.ReadWrite` as well (draft creation), and `send()` raises before any network call if the token lacks it. Update that paragraph: default `co auth microsoft` tokens have it; tokens issued before `Mail.ReadWrite` was added need a re-auth.

### Labels and target release

- Labels: bug
- Proposed target version: 1.8.2
- Why: same bug class as #1198, which is now fixed on `main` for `send` only; the docs line actively misleads users with older tokens.

Discovered while reviewing #1221; not in that PR's scope (the contributor kept replies unchanged on purpose and said so).

wu-changxing:
## Description

`co outlook reply --at` still went through `POST /me/messages/{id}/reply` with the deferred-send property tucked inside `message`. The reply action is create-and-send in one step, exactly like `sendMail` was for `send()` before #1221, and Microsoft Graph lists neither of them among the requests that can carry an extended property. So a scheduled reply had the same shape as the bug #1221 just fixed for `send`.

Scheduled replies now follow the documented draft lifecycle:

```
POST  /me/messages/{id}/createReply      reply draft, threading kept (Mail.ReadWrite)
POST  /me/messages/{draft}/attachments   one per file — PATCH does not accept attachments
PATCH /me/messages/{draft}               SystemTime 0x3FEF + 0x000F (PidTagDeferredSendTime / DeliveryTime)
POST  /me/messages/{draft}/send          submit that exact draft (Mail.Send)
```

Immediate replies are unchanged. The scheduled branch checks `Mail.ReadWrite` before any request, as `send()` does. The scheduled reply draft carries `0x3FEF`, so it appears in `co outlook scheduled` and `co outlook cancel` deletes it.

Sources behind the route:
- Create extended property, supported request syntaxes (no `sendMail`, no `/reply`): https://learn.microsoft.com/en-us/graph/api/singlevaluelegacyextendedproperty-post-singlevalueextendedproperties?view=graph-rest-1.0
- `message: update` — `singleValueExtendedProperties` "Updatable only if isDraft = true": https://learn.microsoft.com/en-us/graph/api/message-update?view=graph-rest-1.0
- `message: createReply`: https://learn.microsoft.com/en-us/graph/api/message-createreply?view=graph-rest-1.0
- `message: send` accepts a reply draft: https://learn.microsoft.com/en-us/graph/api/message-send?view=graph-rest-1.0
- Add attachment to a draft: https://learn.microsoft.com/en-us/graph/api/message-post-attachments?view=graph-rest-1.0

## Related Issue

Fixes #1416 (follow-up to #1198 / #1221)

- **Suggested labels:** bug
- **Proposed target version:** 1.8.2
- **Estimated release window:** next patch from `main`
- **Why this release:** Same bug class as #1198, which shipped its `send` half on `main` in #1221. Only the scheduled-reply transport path changes; immediate replies and sends keep their endpoints. Rollback is a single commit.
- **Forward-port tracking issue (stable patches only):** #1418

## How this was written

- [x] AI-assisted — reviewed every line and can explain why each change is there

- Least confident: no live Microsoft mailbox was available in this environment, so Exchange actually holding a *reply* draft the way it holds a new draft is covered by the mocked HTTP contract only. This PR stays a draft until someone with a mailbox runs the check below.
- If this is wrong, what breaks first? Either `createReply` ignores `comment` on v1.0 (the reply body would be empty), or Exchange sends the reply draft immediately despite the PATCHed properties. Both show up in the live check before anything ships.

## Type of Change

## Dev Blog (required)

> docs/blog/2026-09-05-a-reply-is-a-send-too.md

## Changes Made

- `connectonion/useful_tools/outlook.py`: `reply(send_at=...)` calls a new `_scheduled_reply()` (createReply → attachments → PATCH → send); immediate replies unchanged; attachments are still encoded before any request so a rejected file leaves no stray draft.
- `tests/unit/test_outlook.py`: scheduled-reply tests pin the four requests, order, payloads and empty send body; new tests for the missing-scope failure, a createReply response without an id, and a rejected attachment leaving no draft.
- `docs/cli/outlook.md`: scheduling needs `Mail.ReadWrite` (it no longer claims `Mail.Send` alone suffices); scheduled replies appear in `co outlook scheduled`.
- `docs/blog/2026-09-05-a-reply-is-a-send-too.md`: the story.

## Testing

- `pytest tests/unit/test_outlook.py` — 73 passed (70 on `main` + 3 new)
- `pytest tests/unit/test_outlook_commands.py tests/e2e/cli/test_cli_outlook_reply.py` — 43 passed
- `ruff check` on both touched Python files — same 10 pre-existing findings as `main`, none new

**Live check requested before this leaves draft** (same protocol as #1221):

1. `co outlook reply <#> "..." --at +10m` on a real mailbox.
2. Before the time: the reply is *not* in the recipient's inbox and *not* in Sent Items; it *is* listed by `co outlook scheduled`.
3. After the time: exactly one copy in the recipient's inbox, threaded under the original, exactly one in Sent Items, none left in Drafts.
4. Repeat with `--attach` and confirm the file arrives.
5. Once more, then `co outlook cancel <#>` before the time: nothing is delivered.

## Scope

- `connectonion/useful_tools/outlook.py`
- `tests/unit/test_outlook.py`
- `docs/cli/outlook.md`
- `docs/blog/2026-09-05-a-reply-is-a-send-too.md`

## Example Usage

```bash
co outlook reply 3 "See you then." --at +2h
co outlook scheduled        # the reply draft is listed here
co outlook cancel 1         # deletes it before Exchange sends it
```

## Checklist

- [x] Any dependent changes have been merged and published (#1221)
- [x] Forward-port tracker linked (#1418)
- [ ] Live mailbox validation (see Testing) — pending

Not applicable.

## Additional Notes

Draft on purpose: the unit tests prove the request contract, not Exchange's behaviour on a reply draft. Ready for review once the live check above has been run.

https://claude.ai/code/session_01WaAFWfbdhFGiYo64Hvn1gn

### DIFF
--- a/connectonion/useful_tools/outlook.py
+++ b/connectonion/useful_tools/outlook.py
@@ -2,7 +2,7 @@
 Purpose: Outlook integration tool for email and contact management via Microsoft Graph API
 LLM-Note:
   Dependencies: imports from [os, html, httpx] | imported by [useful_tools/__init__.py] | requires OAuth tokens from 'co auth microsoft' | tested by [tests/unit/test_outlook.py]
-  Data flow: Agent calls Outlook methods → _get_access_token() uses a cached Microsoft access token while its expiry is valid or unknown, preemptively refreshing near expiry via oo-api and refreshing after a Graph 401 → HTTP calls to Graph API (https://graph.microsoft.com/v1.0) → returns email/contact data or confirmations | download_attachments() decodes Graph fileAttachment bytes into a caller-selected project directory without overwriting existing paths | send()/reply() with attachments share _encoded_attachments() (validate, size-check, base64) and place fileAttachments on the sent message — reply() puts them on the reply action's message so Graph still threads it | send()/reply() with send_at attach deferred-send extended property (SystemTime 0x3FEF) so Exchange holds delivery | reply() escapes bodies (html.escape) and converts to HTML <p> paragraphs (blank-line splits, \n → <br>) since Graph renders the comment as HTML | get_scheduled() and contacts page through Graph collections
+  Data flow: Agent calls Outlook methods → _get_access_token() uses a cached Microsoft access token while its expiry is valid or unknown, preemptively refreshing near expiry via oo-api and refreshing after a Graph 401 → HTTP calls to Graph API (https://graph.microsoft.com/v1.0) → returns email/contact data or confirmations | download_attachments() decodes Graph fileAttachment bytes into a caller-selected project directory without overwriting existing paths | send()/reply() with attachments share _encoded_attachments() (validate, size-check, base64) and place fileAttachments on the sent message — reply() puts them on the reply action's message so Graph still threads it | send()/reply() with send_at never use the one-shot sendMail / reply actions (they delivered at once, #1198): send() creates a draft carrying PidTagDeferredSendTime (SystemTime 0x3FEF) and PidTagDeferredDeliveryTime (0x000F) and submits it; reply() does createReply → POST attachments → PATCH the same two properties → send, so Exchange holds the exact draft until then (needs Mail.ReadWrite) | reply() escapes bodies (html.escape) and converts to HTML <p> paragraphs (blank-line splits, \n → <br>) since Graph renders the comment as HTML | get_scheduled() and contacts page through Graph collections
   State/Effects: reads MICROSOFT_* env vars for OAuth tokens/scopes | makes HTTP calls to Microsoft Graph API | can modify mailbox state (mark read, archive, send emails), create contacts, and write downloaded attachments inside the project boundary | token refresh rewrites ~/.co/keys.env
   Integration: exposes Outlook class with email methods plus add_contact(), list_contacts(), search_contacts() | structured list methods feed cli/commands/outlook_commands.py | used as agent tool via Agent(tools=[Outlook()]) | reply(email_id, body, send_at, *, attachments) keeps send_at third positional for pre-attachment callers, so attachments is keyword-only
   Performance: network I/O per API call | batch fetching for list operations | email body fetched separately
@@ -789,7 +789,8 @@ class Outlook:
             email_id: Outlook message ID to reply to
             body: Reply message body (plain text)
             send_at: Optional UTC ISO time (e.g. "2026-07-06T15:30:00Z") to
-                schedule delivery — Exchange holds the reply until then
+                schedule delivery — the reply becomes a deferred draft that
+                Exchange holds until then (needs the Mail.ReadWrite scope)
             attachments: Optional list of local file paths to attach — same
                 files and ~3MB total Graph limit as send()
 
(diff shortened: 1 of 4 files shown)


wu-changxing:
@mikemikimike this is the reply-side half of what you fixed in #1221, done the same way: `createReply` → attachments → `PATCH` the two deferred-send properties → `send`. Would you review it, and if you still have that test mailbox, run the live check in the Testing section? It is the same protocol you used for #1221, plus a threaded reply and a `--attach` case. That is the one piece of evidence this PR cannot produce on its own, and it is what takes it out of draft.

Two things worth a second pair of eyes:

- `createReply` on v1.0 is documented as needing no body, but the same page refers to the `comment` parameter and the beta page tabulates it. If the reply body comes through empty on a real mailbox, the fallback is `createReply` without a body followed by a `PATCH` of `body`, at the cost of the quoted original.
- Attachments go to `/me/messages/{draft}/attachments` one per file because the `PATCH` property table does not list `attachments`. If you know a reason to prefer passing them inside `createReply`'s `message`, say so.

If anything is off, push to the branch directly (maintainer edits are on) or leave comments and I will take them.

---

mikemikimike:
@wu-changxing I reviewed the branch and ran the live mailbox check.

The Microsoft Graph v1.0 contract supports this lifecycle: `createReply` returns a draft, attachments are added through the draft's `/attachments` collection, and the draft is submitted with an empty-body `POST /send`. I found no reason to replace the current attachment flow with `createReply`'s optional `message` payload.

Validation:
- Focused PR cases: 13 passed.
- Live scheduled delivery: no Inbox/Sent copy before due; exactly one in each after due; both stayed in the source conversation; the attachment was present on both delivered copies.
- Live cancellation: the scheduled draft was listed, deleted before due, and produced no Inbox/Sent copy after due.
- `git diff --check`: passed.

I don't see a blocking issue in this diff from my side. The broader Outlook/CLI subset had 113 passed, 1 skipped, and 2 unrelated Windows baseline failures (symlink privilege and an existing attachment-race assertion).

mikemikimike:
Follow-up on the two points:

1. The v1.0 page is worded confusingly: the request body is optional, but its JSON rules explicitly allow either `comment` or `message.body` (and reject specifying both). I tested the current `comment` path against a real mailbox: the marker/body was present in both the delivered Inbox copy and the Sent copy, so no empty-body fallback or `PATCH body` was needed. See the [v1.0 createReply documentation](https://learn.microsoft.com/en-us/graph/api/message-createreply?view=graph-rest-1.0).

2. The documented operation for adding an attachment to an existing message is a POST of one Attachment object to `/me/messages/{draft}/attachments`; the current one-request-per-file loop follows that contract. I also verified that the attachment survived into both delivered copies. Passing attachments in `createReply`'s `message` object could be considered separately, but I found no correctness reason to switch: the draft still needs to exist before the deferred-send properties are patched, and the collection endpoint is the documented path for mutating that draft. See the [v1.0 attachment documentation](https://learn.microsoft.com/en-us/graph/api/message-post-attachments?view=graph-rest-1.0).

mikemikimike:
@wu-changxing I reviewed the branch and verified the complete deferred-reply flow against a real mailbox. The documentation is slightly inconsistent, so here is the precise conclusion.

For the reply body, the v1.0 `createReply` page says the request body is optional, but its JSON guidance allows either `comment` or `message.body`. The beta page provides the parameter table and explicitly says that `comment` may be an empty string. The current implementation uses `comment`; in the live test, the body arrived intact in both the Inbox and Sent copies, so no fallback or `PATCH body` was needed.

For attachments, v1.0 explicitly documents adding an attachment to an existing message by POSTing one Attachment object at `/me/messages/{draft}/attachments`. The `createReply` page does not show a concrete `message.attachments` example, so I am not claiming that the docs prohibit that form. The current flow creates the threaded reply draft, adds each attachment through the documented collection, patches the deferred-send properties, and sends that draft. The live test confirmed the attachment was present on both delivered copies.

The live checks also showed that the scheduled reply produced no Inbox/Sent copy before its due time, then exactly one copy in each folder in the original conversation. A second scheduled reply was listed and canceled before its due time, and produced no Inbox/Sent copy afterward. Focused PR tests passed (13 passed), and `git diff --check` passed. The broader Outlook/CLI subset had 113 passed, 1 skipped, and two unrelated Windows baseline failures.

I do not see a blocking issue in this diff from my side.

wu-changxing:
Thank you, that is exactly the evidence this needed: body via `comment` intact, attachment on both copies through the attachments collection, nothing before the due time, one copy each after it in the original conversation, and a cancelled one that never went out. Out of draft and merging on that. The two Windows failures you saw are outside this diff; the Windows jobs here are green.

---

Issue #1407, opened 2026-09-03
Forward-port tracker for 1.8.1 (#1406): backport the replay-ledger self-heal to the 1.7 stable line


wu-changxing:
Tracker required by the stable-patch rule for #1406 (target 1.8.1, released from `main`).

What has to move between lines:

- **To the 1.7.x stable line:** the `SignatureReplayStore` self-heal from #1403 (recreate the schema on "no such table" and retry the claim once; print the SQLite error next to the safe client string). 1.7 hosts keep the SQLite ledger from #805 for every socket, so a deleted `.co/replay.sqlite3` takes a 1.7 host down exactly as it took the Melbourne rental host down on 2026-09-03. The in-memory `host()` ledger could go too; the sealed-channel parts cannot (1.7 has no `SEAL`).
- **To `main`:** nothing — #1406 lands on `main` and 1.8.1 is cut from it.

Close when the 1.7 backport PR is merged or a maintainer decides the 1.7 line will not receive it.

wu-changxing:
Closed by the 1.7 backport, which is merged and released.

- Backport PR: #1420, merged into `release/1.7`. It carries the `SignatureReplayStore` self-heal from #1403: only `no such table` is treated as a vanished ledger, the schema is rebuilt once, the claim is retried once, and a warning names the file, the likely cause and the consequence. Corruption, `database is locked` and I/O errors still fail closed.
- Released as **v1.7.4**: release run 33937265056 succeeded through `publish` and `finalize`, PyPI has both files for 1.7.4, and the GitHub release `v1.7.4` is published (not a draft, not a prerelease).
- Fail-first evidence recorded on #1420: on the unpatched 1.7 tree the new test raises `sqlite3.OperationalError: no such table: used_signatures` and then `ReplayProtectionError`, which is the shape of the 2h44m Melbourne outage on 2026-09-03.

The `main` side of this ledger was always empty: 1.8.1 is cut from `main`, and its own fix (the in-memory ledger for a single-process host, #1406) is already there. `main` now also documents the 1.7.4 release line, forward-integrated in #1422.

With this closed, `check_forward_port_gate.py` no longer blocks a preview, which is what 1.8.1a1 needs.

PR #1421, opened 2026-09-05
Release v1.7.4

wu-changxing:
Cuts 1.7.4 from `release/1.7`, containing exactly one change: the replay-ledger self-heal backported in #1420.

## What ships

A 1.7 host keeps one-use signature digests in `.co/replay.sqlite3` and created that schema once, at startup. Any deploy that replaced or removed the file left an empty database, every later signed call raised `no such table: used_signatures`, and the host refused all authenticated work while still reporting itself healthy. That is what took the Melbourne rental host offline for 2h44m on 2026-09-03.

The store now distinguishes a vanished ledger from a real fault: only a missing table is rebuilt, once, with the claim retried once and a warning naming the file, the likely cause and the consequence. Corruption, `database is locked` and I/O errors still fail closed.

## Version files

All four literals move together: `connectonion/_version.py`, `pyproject.toml`, `## Current Version:` in `VERSIONING.md`, and the root package in `uv.lock`. The `1.7.4` release line is added to `VERSIONING.md`.

## Verification

- Full unit suite on this branch: **6759 passed, 15 skipped**.
- `test_the_version_agrees_with_itself.py`: 15 passed.
- One test is deselected locally and only locally: `test_every_released_tag_is_documented` compares every `v*` tag in the checkout against this branch's `VERSIONING.md`, and this working copy also has `main`'s tags, so it reports `1.8.0` as undocumented on the 1.7 line. Confirmed pre-existing by running it on unmodified `release/1.7`. CI does not fetch tags, so it skips there.

Worth recording for whoever bumps a version next: after editing `_version.py`, Python may keep serving the previous value from `__pycache__`. `1.7.3` and `1.7.4` are the same length, and a `git stash pop` restored the original mtime, so the cached bytecode looked valid and the version tests failed against a file that was already correct. Clearing `__pycache__` resolved it.

**Proposed target version:** 1.7.4 / next patch
**Estimated release window:** on merge, tagged immediately
**Forward-port tracking issue (stable patches only):** #1407

https://claude.ai/code/session_01KmCtowqB3DqGVPpTb1VU3S


wu-changxing:
Applying `no-blog` to this PR rather than adding a post: the Design Journal piece that explains this exact fix, `docs/blog/2026-09-03-the-agent-that-passed-every-health-check.md`, came onto `release/1.7` with the backport in #1420 and is already on this branch. This PR only moves the four version literals and adds the VERSIONING.md line, so a second post would restate it.

PR #1433, opened 2026-09-05
[1.9.1] feat(listen): add Discord mailbox adapter

wu-changxing:
## Current release decision — 2026-09-15

Target approximately **1.8.8**, tentative, under #1463. Removed from 1.8.6; the replacement 1.8.6 plan is #1555. WhatsApp existing-group access #1543 is a separate 1.8.6 workstream. This supersedes earlier release dates below.

---

**Proposed target version:** After 1.8.5 (exact version unassigned)
**Estimated release window:** After 1.8.5, once dependencies and live provider acceptance are complete
**Forward-port tracking issue (stable patches only):** #1411

## Problem

Discord inbound support was tracked in #341, but the mailbox train had no Gateway consumer.

Closes #341. Part of #1411. This is stacked on #1400 so all providers share the mailbox contract from #1398.

## Approach

- implement Gateway v10 discovery, Identify/Resume, heartbeat ACK monitoring, sequence tracking, and bounded reconnect backoff
- request only the required guild, DM, message, and Message Content intents
- normalize MESSAGE_CREATE into the seven-field local mailbox envelope
- filter bot, webhook, self-authored, and malformed events before persistence
- add direct REST send/reply with the 2,000-character boundary and one explicit 429 retry
- wire CLI, check/status/doctor, docs, and the required design journal

The bot token remains local in keys.env and is redacted from errors.

## Verification

- provider/mailbox/CLI focused suite: 183 passed
- the combined provider/keys/status/doctor suite later passed 189 tests with the stacked WhatsApp branch
- python -m compileall -q connectonion/listen connectonion/cli/commands/listen_commands.py
- git diff --check

## External release gates

- enable the privileged Message Content intent for a real bot
- complete a live server run covering receive, disconnect/resume, reply, and rate limiting
- no Discord app registration or real messages were performed by this PR

## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

Issue #1411, opened 2026-09-04
Track listen feature delivery on main


wu-changxing:
## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

## Release classification — 2026-09-05

This is the open release-line tracker for the stacked messaging adapter train targeting a release after 1.8.5:

- #1399 DD-063 (merged design decision)
- #1398 shared mailbox plus Feishu/Lark
- #1400 Telegram inbound
- #1433 Discord inbound/outbound adapter
- #1434 WhatsApp adapter (also depends on openonion/oo-api#227)

## Forward-port plan

- **To main:** all adapter PRs land through the stack rooted on #1398/#1400; adapter commits must be excluded from releases through 1.8.5.
- **To 1.7.x:** nothing. These are new product surfaces and 1.7 is maintenance-only.
- **To a future higher active line:** if one is cut before this stack merges, forward-port the merged stack and record the resulting PR here. If it is cut from main afterward, it inherits the changes.

Keep this issue open and labelled `forward-port-required` until the complete stack has landed on every active higher line, or a maintainer records that no separate higher line exists. Live provider acceptance remains tracked in the provider issues and is not implied by merge.

wu-changxing:
Release-line audit for the owner-authorized 1.8.4a1 preview, 8 September 2026: a fresh wildcard fetch of all origin branches confirms there is **no separate higher active release line** beyond continuing main (1.8.x) and the 1.7 maintenance line. Higher-version roadmap milestones are not separate code release lines. This records the explicit exception in this issue’s forward-port plan.

Removing `forward-port-required`, whose definition is stable patch fixes missing from active higher lines: this issue tracks still-unmerged new adapters, explicitly excluded through 1.8.5, rather than a missing stable patch. The tracker stays **open** and all adapter/provider acceptance requirements and the after-1.8.5 deferral stay in force. If a separate higher line is cut before this stack lands, record and label the resulting forward-port obligation here. No adapter code is included in 1.8.4a1 by this metadata correction; the release gate itself is unchanged.

wu-changxing:
## Release scheduling update — 2026-09-08

[...]

Issue #1389, opened 2026-09-02
[1.8.5] `co <provider> listen | receive | send`: one mailbox directory, three verbs, any agent can consume


wu-changxing:
## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

## Problem

After 1.8.0b1 a deployed Agent still cannot be woken by anything a person actually uses. Nobody messages an Agent from where they already are. The outbound half exists in fragments (`co telegram send`, the 1.8 encrypted SMS inbox, `co email`). **The inbound half exists nowhere.** Audited on `main` at 8241003:

- No `co listen` or `receive` command anywhere under `connectonion/cli/`.
- No Feishu or Lark module in the tree.
- The only "whatsapp" string in the repository is the unrelated Ed25519 remark at `docs/design-decisions/006-agent-address-format.md:32`.

Seven inbound-channel issues are open and none has code: #336 (email), #341 (Discord), #352 (Telegram), #357 (Slack), #349 (WhatsApp), #358 (SMS), #1310 (Feishu/Lark).

## Decision

This issue went through three designs on 2026-09-02 (Host lifespan with trust integration; standalone OIP client with its own allowlist; a pure pipe). The third is the one. It is recorded in DD-063 on branch `claude/v1-9-adapter-work-ug07b4`.

**The tool turns messages into files. Consumers come by themselves.** It knows nothing about Agents, Hosts, OIP, trust, or permissions. Those are other layers.

### One directory per provider

```text
~/.co/feishu/
├── inbox.jsonl      every message received, one JSON line, appended, never deleted   (the log)
├── outbox.jsonl     every message sent, and every send that failed
├── new/             messages nobody has taken yet, one file each                      (the queue)
│   └── 1756808267-om_9f8e
├── cur/             taken but not yet replied; older than 1 h goes back to new/
├── log              the tool's own log: connected, reconnecting, send failed
└── listen.lock      pid of the running listener
```

[...]

PR #1398, opened 2026-09-02
feat(listen): co feishu listen / receive / send / reply — a Feishu bot as a directory of files (1.8.5)

wu-changxing:
## Description

The first inbound chat channel, built as DD-063 specifies: a Feishu (or Lark) bot becomes a directory of files under `~/.co/feishu/`, and any program that can read a file can answer messages.

```text
~/.co/feishu/
├── inbox.jsonl      every message received, one JSON line, appended, never deleted
├── outbox.jsonl     every message sent, and every send that failed
├── new/             messages nobody has taken yet, one file each
├── cur/             taken but not yet replied; back to new/ after an hour
├── log              the tool's own log
└── listen.lock      pid of the running listener
```

```bash
co feishu listen                    # hold the SDK long connection; every message → the directory
m=$(co feishu receive)              # block for the next one, take it, print one JSON line
echo "done" | co feishu reply om_9f8e   # back to the chat and thread it came from
co feishu done om_9f8e              # took it, chose silence; do not bring it back
co feishu serve -- claude -p        # one command per message, stdout is the reply
co feishu check | ls | log -f | send CHAT
```

`co lark` is the same class with the Lark domain and `LARK_*` credentials. Nothing under `connectonion/listen/` imports `core/` or `network/host/`: the tool knows nothing about Agents, Hosts, OIP, trust or permissions.

Stacked on #1399 (base branch `claude/v1-9-adapter-work-ug07b4`); the diff here is the code and its docs only.

## Related Issue

Part of #1389. Implements the tool half of #1310 (the live acceptance run there is still to do).

## Labels and target release (required)

- **Suggested labels:** feature, platform
- **Proposed target version:** 1.8.5 (first cut as `1.8.5a1`), decided 2026-09-08 in #1462
- **Estimated release window:** right after 1.8.4 publishes; the branch is merged up to `main` at 1.8.4a2 (#1461)
- **Milestone:** 1.8.5 — Feishu/Lark listen on the shared mailbox
- **Why this release:** we run Feishu/Lark daily and today watch the One group with a 30-minute polling script; this replaces it with a live long connection. Self-contained module plus a CLI group; no Host, OIP or core change, so rollback is deleting the group. `lark-oapi` is imported lazily and reported with the pip command when missing, so the dependency set is unchanged.
- **Forward-port tracking issue (stable patches only):** #1411

## How this was written

- [ ] I wrote this myself
- [ ] AI-assisted — I reviewed every line and can explain why each change is there
- [x] Mostly AI-generated

[...]

PR #1434, opened 2026-09-05
[1.9.1] feat(listen): add WhatsApp Cloud API mailbox adapter

### LINKED ISSUE #349: Make WhatsApp a real inbound channel — `co whatsapp listen` pulling an oo-api webhook inbox
## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

> ## Revision 2026-09-02
>
> Re-specified onto the contract in #1389 (one mailbox directory, three verbs). The original analysis below is still right about transport, identity and business verification; its shape predates `co <provider> listen`.
>
> - **Verbs:** `co whatsapp listen`, `receive`, `send`, `reply ID`, `check`, identical to Feishu (#1310, PR #1398) and Telegram (#352, PR #1400). Directory `~/.co/whatsapp/`. Message JSON is the same seven fields; `chat` is the sender's E.164 number, `thread` is always null.
> - **The one difference from every other provider:** Meta only pushes to a public HTTPS webhook, so a laptop cannot receive directly. `co whatsapp listen` therefore pulls from **oo-api**, which receives the Meta webhook, verifies `X-Hub-Signature-256`, and stores messages for the owning agent. Same shape as `co sms` today. `--webhook-port N` is the escape hatch for a machine with its own public address.
> - **Backend work, oo-api:** `GET`/`POST` webhook endpoint per registered WhatsApp Business Account (verify-token handshake, signature check, store by `wamid`), an inbox read/ack endpoint the CLI polls, and a send proxy is *not* needed: `send` calls the Cloud API directly with the user's token.
> - **Business model, decided for now:** users bring their own WhatsApp Business Account, Meta app and number, and point the webhook at oo-api. This matches the `co sms` precedent (user's own phone). OpenOnion provisioning numbers the way `co email` provisions addresses is a later product decision, not part of this issue. Business verification is still the long pole and should start as soon as someone owns this.
> - **`reply` outside the 24-hour customer service window** exits 3 with the reason, because Meta only accepts pre-approved templates there. Templates are out of scope.
> - **Order:** after Feishu (#1310) and Telegram (#352), because it is the only one needing backend work. The provider slots into `connectonion/listen/` beside `feishu.py` and `telegram.py` from #1398/#1400.
>
> Target 1.8.3 if the oo-api side lands in time, otherwise 1.9.0.

---

*Original text follows.*

## Motivation

Same gap as email (#336) and Discord (#341): a messaging platform that could trigger `agent.input()` but doesn't. Confirmed there's nothing here yet — the only "WhatsApp" hit anywhere in the repo is an unrelated mention of Ed25519 being the signature scheme WhatsApp also uses (`docs/design-decisions/006-agent-address-format.md:32`), not integration code.

This is the third inbound channel (email, Discord, now this), which is the point #341 flagged as "worth considering a shared abstraction." Decided explicitly: **write this standalone, matching Discord's shape, don't build the shared abstraction yet** — priority here is shipping WhatsApp soon, not spending time up front on a generalized adapter three data points can't fully inform yet.

## What's different from Discord (don't copy blindly)

- **No simple bot library.** WhatsApp requires either Meta's official WhatsApp Business Cloud API or Twilio's WhatsApp API. Both need a registered WhatsApp Business number and, for production (non-sandbox) use, Meta business verification — this has a real approval-latency cost that Discord's bot-token setup didn't have. Flag this up front so "ship soon" expectations account for external approval time, not just code time.
- **Transport shape is webhook-based**, closer to the email trigger (#336) than to Discord's persistent bot connection: Meta/Twilio calls a webhook URL on new messages, rather than the agent process holding an open connection.
- **Identity**: WhatsApp senders are phone numbers, authenticated by the platform (real SIM-backed numbers, unlike a forgeable email `From:` header) — same category as Discord's trust question (who's allowed to trigger it), not email's (is the claimed identity even real). Needs an explicit allow-list of phone numbers/contacts, not an open door to anyone who messages the business number.
- **No threading question** — WhatsApp conversations are already per-contact threads, nothing extra to preserve.
- Media (images, voice notes) exists on the platform but is **out of scope for v1** — text messages only.

## Scope

```
WhatsApp message to the business number
        │
        ▼
Meta/Twilio webhook fires
        │
        ▼
agent.input(message.text)
        │
        ▼
reply sent back via the Cloud/Twilio API to the same conversation
```

## Relationship to other issues

- #336 (email) and #341 (Discord) — same category of gap, each shipped standalone on purpose
- If a fourth channel comes up after this, that's a reasonable point to revisit the shared-abstraction question — not before

## Definition of done

- Sending a WhatsApp message to the configured business number causes `agent.input()` to run and a reply to come back in the same conversation
- Only allow-listed phone numbers/contacts can trigger it — not anyone who messages the number
- Setup docs cover the Meta/Twilio account and business-verification steps, not just the code
- A test that doesn't require a live WhatsApp/Twilio connection (mocked webhook payload → `agent.input()` call)

wu-changxing:
## Current release decision — 2026-09-15

Target approximately **1.8.8**, tentative, under #1463. Removed from 1.8.6; the replacement 1.8.6 plan is #1555. WhatsApp existing-group access #1543 is a separate 1.8.6 workstream. This supersedes earlier release dates below.

---

**Proposed target version:** After 1.8.5 (exact version unassigned)
**Estimated release window:** After 1.8.5, once dependencies and live provider acceptance are complete
**Forward-port tracking issue (stable patches only):** #1411

## Problem

WhatsApp/O API support in #349 needs a client adapter that preserves local credential ownership while making webhook delivery durable.

Closes #349. Part of #1411. Depends on openonion/oo-api#227 and is stacked on #1433.

## Approach

- add co whatsapp bind without putting app secret or verify token on argv
- poll the authenticated O API inbox using claim leases
- ACK only after the ordinary seven-field message is stored locally; NACK failed local writes
- send and reply directly through an explicitly pinned Meta Graph API version
- turn Meta error 131047 into a clear template-required policy error
- keep the access token local and redact access/app/verify secrets from keys, status, doctor, and provider errors
- add CLI docs, preview caveats, design journal, and fixture tests

## Verification

- python -m pytest tests/unit/test_listen_whatsapp.py tests/unit/test_listen_discord.py tests/unit/test_listen_telegram.py tests/unit/test_listen_feishu.py tests/unit/test_listen_mailbox.py tests/unit/test_listen_commands.py tests/e2e/cli/test_cli_help.py tests/e2e/cli/test_cli_keys.py tests/e2e/cli/test_cli_status.py tests/e2e/cli/test_cli_doctor.py tests/unit/test_doctor_credentials_are_redacted.py -q
  - 189 passed
- python -m compileall -q connectonion/listen connectonion/cli/commands/listen_commands.py
- git diff --check

## External release gates

- merge and deploy openonion/oo-api#227 with MESSAGING_CREDENTIAL_KEY and its migration
- register/configure a real Meta app and WABA webhook subscription
- exercise challenge, signed inbound delivery, duplicate delivery, local ACK/NACK, direct reply, and the 24-hour window
- no external app registration, credentials, real messages, or deployment were performed by this PR

## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -178,11 +178,12 @@ connectonion/
 │   │       ├── tools.py            # Verification tools
 │   │       ├── trust_agent.py      # TrustAgent class
 │   │       └── policies/           # Trust level policy markdown
-│   ├── listen/                     # Chat platforms as mailbox directories (co feishu / co telegram listen, receive, send, reply)
+│   ├── listen/                     # Chat platforms as mailbox directories (co feishu / co telegram / co discord / co whatsapp)
 │   │   ├── mailbox.py              # ~/.co/<provider>/: inbox.jsonl log, new/ queue, cur/, outbox.jsonl
 │   │   ├── feishu.py               # Feishu/Lark long connection → files; REST reply
 │   │   ├── telegram.py             # Telegram getUpdates long poll → files; sendMessage reply
-│   │   └── discord.py              # Discord Gateway → files; REST reply
+│   │   ├── discord.py              # Discord Gateway → files; REST reply
+│   │   └── whatsapp.py             # O API webhook inbox → files; Meta REST reply
 │   ├── tui/                        # Terminal UI components
 │   ├── logger.py                   # Unified logging facade (terminal + file + YAML sessions)
 │   ├── console.py                  # Low-level terminal output with Rich
--- a/connectonion/cli/commands/doctor_commands.py
+++ b/connectonion/cli/commands/doctor_commands.py
@@ -403,6 +403,9 @@ def handle_doctor(*, fix: bool = False, yes: bool = False, json_output: bool = F
         "MISTRAL_API_KEY": "set MISTRAL_API_KEY in <project>/.env",
         "TELEGRAM_BOT_TOKEN": "set TELEGRAM_BOT_TOKEN in ~/.co/keys.env",
         "DISCORD_BOT_TOKEN": "set DISCORD_BOT_TOKEN in ~/.co/keys.env",
+        "WHATSAPP_ACCESS_TOKEN": "set WHATSAPP_ACCESS_TOKEN in ~/.co/keys.env",
+        "WHATSAPP_APP_SECRET": "set WHATSAPP_APP_SECRET in ~/.co/keys.env",
+        "WHATSAPP_VERIFY_TOKEN": "set WHATSAPP_VERIFY_TOKEN in ~/.co/keys.env",
     }
     project_dir = project_co_dir().parent
     selected_api_key = _selected_credential_values(
--- a/connectonion/cli/commands/keys_commands.py
+++ b/connectonion/cli/commands/keys_commands.py
@@ -73,6 +73,9 @@ def _load_env_vars(
         "MICROSOFT_REFRESH_TOKEN",
         "TELEGRAM_BOT_TOKEN",
         "DISCORD_BOT_TOKEN",
+        "WHATSAPP_ACCESS_TOKEN",
+        "WHATSAPP_APP_SECRET",
+        "WHATSAPP_VERIFY_TOKEN",
     )
     return _selected_credential_values(
         names,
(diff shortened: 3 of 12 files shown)


wu-changxing:
Scheduled for **1.8.6** (plan #1463) as of 2026-09-08; the "after 1.8.5" deferral is lifted. Stays stacked on #1398, which ships in 1.8.5. Do not change the three-verb contract from #1389. Tracker: #1411.

Issue #1462, opened 2026-09-08
[1.8.5] Release plan: shared mailbox core and Feishu/Lark listen


wu-changxing:
# 1.8.5 release plan: shared mailbox core and Feishu/Lark listen

Decided 8 September 2026. This reverses the 5 September note that deferred all
new messaging adapters until after 1.8.5. We run Feishu/Lark every day, so the
first inbound channel ships as soon as 1.8.4 is published, and the remaining
adapters follow one version later (see the 1.8.6 plan).

- **Target version:** 1.8.5, published first as `1.8.5a1` per VERSIONING.md (features enter through a patch preview, `X.Y.0` stays stable)
- **Base ready:** stable [1.8.4](https://github.com/openonion/connectonion/releases/tag/v1.8.4) is published; main commit `2acd6f0e2539bd4d469a68e3b267583b234c5aab`.
- **Integration:** draft #1472 consolidates #1398, #1466, #1464 and #1465 on that stable base. Live Lark testing found an unresolved gap-recovery blocker (see progress below).
- **Branch from:** `main` at 1.8.4, not from `feat/1.8-paid-browser-release` (still at 1.8.0a3)

## Scope

| Work | PR / issue | Notes |
| --- | --- | --- |
| DD-063 one directory, three verbs | #1399 (merged) | Design is already on main |
| Shared mailbox core `connectonion/listen/mailbox.py` | #1398 | Maildir-style `~/.co/<provider>/` with `inbox.jsonl`, `outbox.jsonl`, `new/`, `cur/`; `mv new/X cur/X` is the atomic take |
| `co feishu listen / receive / send / reply` and `co lark …` | #1398, #1310 | Official SDK WebSocket long connection, outbound dial, no port opened; replies over REST with tenant-token caching and rate-limit handling |
| Three-verb contract for every later provider | #1389 | Locked by this release; 1.8.6 adapters must not change it |
| `docs/cli/feishu.md` | #1398 | Ships with the PR |

Out of scope for 1.8.5: Telegram (#1400), Discord (#1433), WhatsApp (#1434). They stay stacked on #1398 and move to 1.8.6.

## Acceptance gates

1. **Real channel, not a mock.** Stop the 30-minute lark-cli polling script that watches the One group today, run `co lark listen` against the same group, and confirm a real @-mention appears under `new/` within seconds. Record the timestamp of the message and of the file.
2. **Atomic take.** Two consumers on the same mailbox never both process one message.
[...]

Issue #1310, opened 2026-08-27
[1.8.5] Add co feishu listen / co lark listen — a mailbox directory any agent can consume


wu-changxing:
## Release scheduling update — 2026-09-05

All new messaging adapters are deferred until **after 1.8.5**. They are excluded from 1.8.2, 1.8.3, 1.8.4, and 1.8.5. The exact later version is not assigned yet. This supersedes earlier release estimates below; existing functionality is unaffected. Provider acceptance and dependency gates still apply. Coordination: #1389 and #1411.

> ## Revision 2026-09-02
>
> The design settled in #1389 and DD-063 supersedes parts of the original text below. What changed:
>
> - **Spelling:** `co feishu listen` / `co lark listen`, noun first, matching `co telegram send` and `co auth login`. Same adapter, two domains, unchanged.
> - **The listener is not in the Host and does not talk to the Agent.** It writes files: `~/.co/feishu/inbox.jsonl` (the log), `new/` (the queue, one file per message), `cur/`, `outbox.jsonl`, `log`. The three verbs are `listen`, `receive`, `send`, plus `reply ID`. The "Identity and authority", "Delivery and session rules" and `co deploy` service sections below are superseded: the tool decides nothing about who may command an Agent, which session a message belongs to, or approvals. Those are the consumer's (`co ai --listen feishu`, Claude Code via `co feishu mcp`, a shell loop). The reasoning is in #1389.
> - **What survives verbatim:** the Feishu application requirements (self-built app, `im.message.receive_v1`, `im:message.group_at_msg:readonly`, `im:message:send_as_bot`, SDK long connection, no public endpoint), the secrets rules, `--check` and `--once` (now `receive --timeout`), foreground process with clean Ctrl-C, `message_id` as the idempotency key, durable stage before ACK, reply in thread when the input came from a thread, exactly one reply after restart or duplicate delivery, redaction in logs, and the whole live-acceptance definition of done.
> - **Target:** 1.8.1 after 1.8.0 stable, fallback 1.9.0. No milestone decision about widening 1.9 is needed any more.
>
> The message format and directory layout are in #1389. `docs/cli/feishu.md` is the page to write.

---

*Original text follows.*

## Problem

A deployed Agent is passive until an OIP client or another caller sends it input. Adding a Feishu application bot to a group does not currently let a group member wake that Agent by writing:

```text
@Agent investigate the failed deployment
```

[...]

PR #1466, opened 2026-09-08
listen: review of the Feishu/Lark adapter before its live run — refusals in one line, no redelivery storm, Lark-branded errors, docs

wu-changxing:
## Description

Review of #1398 (the 1.8.5 Feishu/Lark listener) before its first live run, stacked on its branch so the review diff reads on its own. Method: give the tool wrong credentials on purpose, read the SDK's failure paths instead of assuming them, and write a failing test for each finding before fixing it.

**Verified against lark-oapi 1.7.3 as installed**, which resolves the PR's two "least confident" items: `ws.Client` accepts `domain` and `log_level`, and exposes `on_reconnecting` / `on_reconnected` as plain attributes (client.py lines 156–161). `EventMessage` has `thread_id`, `root_id`, `chat_type`, `message_type`, `mentions`.

### Findings fixed

| # | Finding | Fix | Test |
|---|---|---|---|
| 1 | `co lark listen` with an invalid app id printed a 40-line Typer traceback ending in `ClientException: 1000040346: app_id is invalid`. | `handle_listen` catches, prints the platform's sentence and the log path, exits 1; lock released. | `test_listen_reports_a_refused_connection_in_one_line_and_exits_1` |
| 2 | After `bot/v3/info` had already said "refused the credentials", `run()` dialled the WebSocket anyway; the SDK's auto-reconnect then wrote `connect failed` every two minutes forever, hiding the refusal. | A `RuntimeError` from `bot_info` (Feishu answered, said no) stops before the dial. Any other exception (no answer at all) still proceeds, since the long connection may get through. | `test_refused_credentials_stop_before_the_connection_is_opened` |
| 3 | The SDK wraps the event handler (`ws/client.py:_handle_data_frame`) and answers HTTP 500 when it raises; Feishu treats 500 as "not delivered, send again" for hours. A payload `to_message` cannot parse would be retried all day. | The handler catches its own parse failure, logs `event not understood (...); skipped`, returns normally. `deliver()` is deliberately left to raise: on a full disk the redelivery is what saves the message. | `test_an_event_the_adapter_cannot_read_is_logged_once_and_the_next_one_still_arrives` |
| 4 | Every error said "Feishu", including for `co lark`; a Lark user is sent to the wrong console. | `Feishu.brand` threaded through `_tenant_token`, `_data`, the rate-limit message. | covered by existing send/error tests (Feishu default unchanged) |
| 5 | `ensure_listener` opened the log for the child and never closed the parent's copy; a long `serve` leaks one handle per restart. | `with logfile.open("ab")` around `Popen`. | existing listener-start tests |
| 6 | Top-level `co --help` listed `feishu` but not `lark`; a comment said eight verbs for nine commands. | help line + comment. | `test_cli_help.py` |

### Second pass (commit `ae218145`): ten more, from an adversarial multi-agent review of the same diff

Each has a test that failed first on the unpatched code.

| # | Finding | Fix |
|---|---|---|
| 7 | **`deliver()` dropped Feishu's redelivery as a duplicate** whenever the log line had been written but the queue file had not: the exact crash the log-first order exists to survive, and the case the blog claimed was "recoverable". | An id in the log is a duplicate only while a file for it is in `new/` or `cur/` or an ok reply is in the outbox; otherwise the queue file is recreated (`re-queued` log line), the log untouched. |
| 8 | **Pid-file lock, two ways broken.** `O_EXCL` created the file empty before the pid was written, so a concurrent starter read `""`, unlinked the winner's lock and both listened (every message twice). After a reboot a reused pid was a phantom nothing cleared. | Reuses `transport.acquire_singleton_lock` (flock, kernel-released on death). The pid in the file is for `check` only. `_alive` and the Windows ctypes branch are gone. |
| 9 | **`serve` consumed a message on any outcome**, so a 529 from the model or a refused reply became an unanswered question; and a mistyped command claimed the message then tracebacked. | Only a sent reply or explicit silence (exit 0, empty stdout) finishes a message; the rest stays taken and returns in an hour. `shutil.which` before the first receive, exit 2. |
| 10 | **Junk in `new/` crashed every consumer.** `.DS_Store`, a zero-byte file, `{"id":…}` without a chat: `JSONDecodeError`/`KeyError` out of `receive`, `serve`, `ls`; `done()` could not remove it; the sweep bounced it back hourly. | Only `<ms>-<id>` names are ours; a queue file that does not parse is moved to `bad/` with a log line. `lookup()` skips a log line with no `chat`. |
| 11 | **My own finding 2 was too broad**: every `RuntimeError` from `bot_info` was treated as a refusal, so a 502 page or a rate limit on that one call killed the listener before the dial. | `CredentialsRefused`, raised only where Feishu rejects the app_id/secret pair; everything else logs and dials. |
| 12 | **`ensure_listener` spawned a child that loaded `~/.co/keys.env` itself**; a `CO_FEISHU_HOME` there or a parent `--env-file` sent it to a different directory than the one waiting for it. | Child gets `CO_<PROVIDER>_HOME` pinned to the parent's root and `--env-file` forwarded. |
| 13 | **Cached tenant token never invalidated** on a 99991663/5 reply; a rotated secret left `serve` failing every reply for up to two hours. | `_call()` drops the cache and retries once on the invalid-token codes. |
| 14 | **`@_user_1` is a prefix of `@_user_10`**: a message tagging ten people lost the tenth (`@Alice @Alice0 hi`). | Substitute longest key first. |
| 15 | **`reply --again` with the same text reused the same reply uuid**; Feishu swallowed it while the CLI printed an id and recorded `ok`. | `send(..., fresh=True)` uses a fresh key; retries still reuse. |
| 16 | **SDK missing gave exit 1 "see the log"** from `receive`/`serve`, contradicting the documented exit-3 shape; the pip command was only inside the log file. | `listen` asks `listen_requirements()` before the lock and exits 3 with the pip line; a listener that died at once has its last log lines printed on stderr, unwrapped. `send`/`reply` still do not need the SDK. |

### Reviewed and left alone, on purpose

- `mentioned` falls back to "any mention counts" when the bot's open_id is unknown. Documented; only reachable when bot-info got no answer, now narrowed to real transport failures.
- `_claim` catches only `FileNotFoundError` on rename. Correct on POSIX; on Windows a lost race can also be `PermissionError`, but the Windows path is already marked untested in #1398 and belongs with its live run.
- No `[feishu]` optional-dependency extra in pyproject for `lark-oapi`. Worth doing, but it touches the lockfile check and belongs in its own small PR.

### Docs

- `docs/cli/feishu.md`: exit-code table, `--raw`, `--no-start`, `serve` in the verb list, what a refusal looks like now, a **Lark** section (domain, `LARK_*`, `~/.co/lark/`), and a **Keeping it running** section with a launchd plist, since that is how listeners run on our machines; systemd mentioned for Linux.
- Verified the CLI loads `~/.co/keys.env` itself (the global `--env` option exists to override it), so the unit carries no secrets.

## Related Issue

#1462 (1.8.5 plan), #1310, #1389. Review of #1398; tracker #1411.

- **Suggested labels:** feature, platform
- **Proposed target version:** 1.8.5, with #1398
- **Estimated release window:** merges into #1398's branch; ships when #1398 does
- **Milestone:** 1.8.5 — Feishu/Lark listen on the shared mailbox
- **Why this release:** it changes what the first live user sees when the tool fails, and removes a retry storm the live run would otherwise have found the hard way
- **Forward-port tracking issue (stable patches only):** #1411

## How this was written

- [ ] I wrote this myself
- [ ] AI-assisted — I reviewed every line and can explain why each change is there
- [x] Mostly AI-generated

**Which model?** Claude Fable 5.1 in Claude Code.

- **Least confident:** that Feishu redelivers on a 500 from the long connection exactly as it does for webhooks. The SDK code is unambiguous about answering 500; the redelivery schedule is from Feishu's event docs and the blog post in #1398, not observed.
- **Not verified:** no live tenant. Findings 1 and 2 were reproduced against `open.larksuite.com` with fake credentials before and after the fix (before: traceback; after: `Feishu refused the credentials: 10003 invalid param`, `details: <log>`, exit 1, no lock left, no reconnect attempts). Finding 3 is from reading the SDK and is pinned by a fake-SDK test.
- **If wrong, what breaks first:** if some transient Feishu-side failure of `bot/v3/info` returns a JSON error code rather than a transport error, finding 2 would make `listen` exit instead of retry; `receive` and `serve` restart it a minute later, so the cost is one minute of not listening and a clear log line.

## Type of Change
- [x] Documentation update

## Dev Blog (required)

> docs/blog/2026-09-08-a-500-means-send-it-again.md

## Changes Made
- `connectonion/listen/feishu.py`: two-clause exception handling around `bot_info` in `run()`; guarded `on_message`; `brand` in messages.
- `connectonion/cli/commands/listen_commands.py`: `handle_listen` catches and reports.
- `connectonion/listen/mailbox.py`: close the parent's log handle after `Popen`.
- `connectonion/cli/main.py`: `lark` help line; comment.
- `tests/unit/test_listen_commands.py`, `tests/unit/test_listen_feishu.py`: three regression tests plus a small fake-SDK helper.
- `docs/cli/feishu.md`, `docs/blog/2026-09-08-a-500-means-send-it-again.md`.

## Testing

Each new test was run on the unpatched code first and failed for the defect's own reason (`RuntimeError` escaping `handle_listen`; `AttributeError: ... 'message_id'` escaping the handler; `DID NOT RAISE` before the dial).

```
$ python -m pytest tests/unit/test_listen_mailbox.py tests/unit/test_listen_feishu.py tests/unit/test_listen_commands.py tests/e2e/cli/test_cli_help.py -q
91 passed        (74 in #1398 + 3 first pass + 14 second pass)
```

Full `tests/unit tests/e2e/cli` on the final tree (network tests deselected, xdist -n 6): **8496 passed, 21 skipped, 0 failed**.

Live, with fake credentials against `open.larksuite.com`, after both passes: `Lark refused the credentials: 10003 invalid param`, `details: <log>`, exit 1, no lock file left, no reconnect attempts.

## Checklist
- [ ] Any dependent changes have been merged and published — merges into #1398, which is still draft pending its live run

https://claude.ai/code/session_01Aa5AJBfKe8P97Wd3ektcVx


wu-changxing:
Full local run on the final tree (`ea513a86`), Python 3.14, xdist `-n 6`, network-marked tests deselected: `tests/unit tests/e2e/cli` → **8482 passed, 21 skipped, 0 failed** (78 s). That is the 8479 from #1398's merged base plus the three regression tests here.

wu-changxing:
Second pass pushed as `ae218145` + `d16f1225`: ten more findings from an adversarial multi-agent review of the same diff, each with a test that failed first on the unpatched code (17 red → 91 green in the listen files). The serious ones: `deliver()` dropped Feishu's redelivery as a duplicate in exactly the crash the log-first order exists to survive; the pid-file lock could yield two listeners (empty-file race) or a phantom after reboot, now an flock via `transport.acquire_singleton_lock`; `serve` consumed a message on any failure; junk in `new/` crashed every consumer; and my own first-pass refusal catch was too broad (`CredentialsRefused` now). Full table in the description. Final tree: **8496 passed, 21 skipped, 0 failed**; fake-credential live path re-verified (`Lark refused the credentials … exit 1`, no lock left).

wu-changxing:
Subsumed. The sixteen review findings from this pass are on main via #1472 and then #1491 (merge f74518120b) — the flock singleton lock, dedup checking new/cur/outbox rather than only the log, serve keeping a failed message for the sweep, and junk in new/ going to bad/. All of it shipped in 1.8.5b1.

Its base branch has been merged and the package has since been renamed to connectonion/inbox/, so there is nothing here to rebase onto.

Issue #210, opened 2026-07-16
Refactor the test suite: eliminate false positives, config drift, and leaked resources


wu-changxing:
## Summary

The test suite is green, but the audit found several cases where green does not mean trustworthy: lifecycle behavior is mocked by reimplementing it inside the test, some tests contain no assertions against production code, and background threads raise unhandled exceptions after pytest has already reported success.

Baseline on commit `1c770f9c585fdff42fd7679c9a3e6fde5e7c5df6`:

```text
2550 collected; 2411 selected; 139 deselected
2404 passed, 7 skipped, 6 warnings in 122.30s
```

After that successful summary, three `BrowserDaemon.serve` threads raised unhandled `ConnectionAbortedError` exceptions from `_srv.accept()`. The latest GitHub Actions run is also green on Python 3.10–3.13, so this is a signal-quality and maintainability issue, not a currently failing-CI issue.

This is a follow-up to #83. That issue fixed 37 failures and later rewrote the old skipped relay/WebSocket tests. The historical 38 skips are no longer the current problem; the root causes called out there—implementation-coupled mocks, loose assertions, and test/architecture drift—still need systematic cleanup.

### P0 — false positives and leaked resources

- `tests/unit/test_events.py:21-60` replaces `Agent._execute_and_record_tools` for the whole module. The fake executor then manually invokes `before_each_tool`, `on_error`, `after_each_tool`, and `after_tools`. If the real Agent → ToolExecutor path stops emitting an event or emits it in the wrong order, these tests can still pass.
- `tests/unit/test_tui_components.py:242-247` and `tests/e2e/test_host_ws.py:342-347` are `pass`-only tests. `tests/unit/test_tui_components.py:249-258` only asserts against a dictionary declared inside the test.
- `tests/unit/test_tui_components.py:264-318` turns import failures into skips. In the baseline run, `SmartInput` and `Pick` were skipped this way, which can hide a packaging or dependency regression.
- `tests/e2e/cli/test_browser_daemon.py:435-463,501-516,760+` starts daemon threads that are not consistently stopped and joined. Pytest exits successfully while three thread exceptions are printed afterward.

### P1 — configuration and command drift

- `pytest.ini` is the active config and excludes `real_api` + `network`; `pyproject.toml` contains a second pytest config that excludes `real_api` + `slow`, sets a different asyncio mode, and defines a different marker set. Pytest does not merge these files.
[...]

PR #1483, opened 2026-09-11
Read-only commands run unattended, and an e2e runs the job instead of the table (#1481)

### LINKED ISSUE #1481: tool_approval denies by default: only 11 commands auto-approve, read-only commands like head/grep are refused unattended
## Summary

`tool_approval` denies by default. Only **11** commands auto-approve, all of them
test/build tools. Everything else — including read-only commands like `head`,
`grep`, `wc`, `cat`, `ls` — falls through to `ask`, and in an unattended run
`ask` means **denied**.

This is not an AI judgement. There is no LLM anywhere in
`useful_plugins/tool_approval/`; it is static list matching.

## The lists, 1.8.4

```
FOCUSED (auto-allow): cargo eslint go jest make mypy pyright pytest ruff tsc vitest
DESTRUCTIVE (deny)  : del erase format rm rmdir shred truncate
EXTERNAL (ask)      : curl mail rsync scp sendmail ssh wget
SENSITIVE (deny)    : aws az env gcloud keychain printenv security
```

Fallthrough, `policy.py:197`:

```python
return decision("command", "ask", "command is outside the focused verification allowlist",
                "call", requires_human=True)
```

So "not known to be dangerous" resolves to "ask a human", not "allow".
`head`, `grep`, `wc`, `cat`, `tail`, `sed`, `ls` are in **no list at all**.

## It compounds: the chain rule

`check_bash_chain_permitted()` splits on `|`, `&&`, `;` and requires **every**
segment to be permitted. Measured on this machine with the shipped
`Bash(co *)` grant:

| command | permitted |
|---|---|
| `CO_WHO=x co browser -t t get_text` | ✅ |
| `CO_WHO=x co browser -t t get_text \| head -40` | ❌ |
| `CO_WHO=x co browser -t t get_text \| grep -i foo` | ❌ |
| `co browser -t t run_page_script a.js && echo ok` | ❌ |

One unlisted read-only segment poisons an otherwise allowed command.

## Production impact, 2026-09-11

An unattended LinkedIn round (7×/day) died mid-phase. The engage phase ran 16
clean iterations, then hit:

```
Tool 'bash' denied by connectonion.auto: command is outside the focused
verification allowlist; no approval channel is available
```

The phase never recovered: it stopped at 28/300 iterations, posted **0**
comments (normal range 3–10), and never wrote its report block, so three
downstream health checks that read that block also failed. The denied step was
an inspect-the-result command — i.e. a pipe into `head`/`grep`.

Same mechanism bit us on the 1.7.0 upgrade, where `co email send` was refused
unattended and a daily digest silently stopped sending.

## Ask: invert the default

Default **allow**; require approval only for what is actually dangerous.

Rationale: an agent is trusted to run commands, and a policy that stops it on
`ls` is one the operator routes around — which is worse for safety than a
policy that stops it only on `rm`.

Concretely:

1. Keep DESTRUCTIVE / SENSITIVE / EXTERNAL as deny-or-ask. Those are the real
   categories.
2. Make the fallthrough `allow` rather than `ask`.
3. If a full inversion is too large for one release, the minimum fix is a
   read-only allow list — `head tail cat grep rg wc ls sed awk sort uniq cut
   basename dirname jq` — none of which writes, networks, or touches
   credentials.
4. Either way, the chain rule needs the same treatment, or piping a permitted
   command into `head` stays broken.

## Note on where this is felt

`ask` is a sensible default when a human is present. It is equivalent to `deny`
when nobody is. The policy currently cannot tell those two situations apart,
and unattended automation is where it does the most damage and is least
visible.

wu-changxing:
## Description

Fixes #1481. `tool_approval` refused every shell command outside eleven test/build tools when nobody was present, so an unattended `co browser ... | head -40` killed a production LinkedIn round at iteration sixteen. This makes read-only commands run unattended, lets a read-only pipe segment ride along beside a granted command, and adds the end-to-end test that models an unattended job — which found three more bugs on the way.

Refs #210 (job-shaped tests), #1462 (rides in 1.8.5).

## What changed

**Policy (`useful_plugins/tool_approval/policy.py`)**

| command | 1.8.4 | now |
|---|---|---|
| `head -40 f`, `grep -i x f`, `wc -l f`, `ls`, `sed -n 1p f`, `cd src && ls` | ask → denied unattended | allow (`read`) |
| `cat /etc/hosts`, `head ~/.ssh/id_rsa`, `cd ..` | ask | ask (outside the workspace, like the read tools) |
| `sed -i s/a/b/ f` | ask | ask (rewrites the file) |
| `echo x > notes.txt`, `cat << 'EOF' > src/main.rs …` | ask / unparseable | allow (`workspace_edit`, the write tool's rule) |
| `echo x > ../out`, `> .co/host.yaml` | ask | deny (outside / control file) |
| `cargo test 2>&1 \| tail -20` | ask (tail) | allow (`2>&1` is not a write) |
| `co browser … get_text \| head -40` with the shipped `Bash(co *)` grant, unattended | denied | allow (`configured_command`) |
| `co browser status && co email send …` | denied | denied |
| `bash << EOF … EOF`, `python3 -c "…"`, `ping 8.8.8.8` | ask → denied | unchanged |

The remote-EXEC whitelist in `host.yaml` (whose own comment forbids `cat`/`head`) is a different gate and is not widened.

**Bugs the e2e found in 1.8.4 (`approval.py`, `policy.py`)**

- A `quiet=True` agent has no console; six call sites reached `agent.logger.console.log_permission_granted(...)` directly, so every auto-approved call in a quiet agent raised `AttributeError` after the policy said yes. One guarded helper now.
- `record_approval_policy` matched the trace entry by `id`; the executor stores the call id as `tool_id`. The decision never reached the trace in a real agent.

**Tests**

- `tests/unit/test_auto_approve_policy.py`: 40 new cases across read-only, outside-workspace, `sed -i`, redirects, heredocs, control files, granted-command chains. 19 of the first 29 failed on the 1.8.4 policy before the fix.
- `tests/unit/test_tool_approval.py`: a quiet Agent with a real MockLLM run gets a tool auto-approved and the decision lands on the trace.
- `tests/e2e/test_an_unattended_agent_builds_a_rust_cli.py`: the job. Scripted model, real everything else, real `cargo`. Asserts which rule carried each of seven steps; a second test removes the two operator grants and asserts exactly those two steps are refused. Skips with a reason where cargo is absent; ubuntu runners have it.
- `tests/e2e/real_api/test_real_unattended_rust_cli.py`: same job, real model chooses the commands. Three runs surfaced the heredoc finding, the `cargo new` → `write` refuses to overwrite → needs `edit` finding, and two correct refusals (`ping`, `python3 -c`). Third run passed in 22 s. Opt-in.

- **Suggested labels:** bug, tests
- **Proposed target version:** 1.8.5
- **Estimated release window:** next alpha (1.8.5a1, with #1472)
- **Milestone:** maintainer assigns
- **Why this release:** production automation is dying on it today; the policy widening is bounded (read-only names, workspace paths, no redirect outside the workspace) and the two 1.8.4 bugs affect every quiet agent. Rollback is a revert.

## How this was written

- [x] Mostly AI-generated — Claude Fable 5.1 via Claude Code, on the maintainer's instruction to fix #1481, design an e2e that would have exposed it, and merge.

**Least confident about:** the read-only list itself. It is a judgement about which commands cannot write or execute their input; `awk` can call `system()` and `sed` has `e` on GNU, neither of which this checks. Both are bounded by the workspace-path and redirect rules, but a determined prompt injection has more room than before. Say so if you want `awk`/`sed` out.

**Not verified:** the Windows branch (`bash` tool refuses Windows anyway); the e2e on macOS CI (the macOS job runs only the browser e2e).

**If this is wrong, what breaks first:** `test_the_two_grants_are_the_only_two_the_job_needs` — it pins exactly which steps need a grant, so a policy change in either direction fails there with the step named.

## Dev Blog

> docs/blog/2026-09-11-sixty-green-tests-and-a-job-that-could-not-run.md

## Testing

```
$ make test
8711 passed, 21 skipped, 6 warnings in 39.87s

$ pytest -m real_api tests/e2e/real_api/test_real_unattended_rust_cli.py   # paid, one run
1 passed in 22.29s
```

Fail-first: with `policy.py`/`approval.py` from `origin/main` swapped in, both scripted e2e tests fail (the first on the quiet-console crash at `mkdir`).

## Scope

- `connectonion/useful_plugins/tool_approval/policy.py` — read-only list, path/redirect/heredoc rules, per-segment grants, trace id match
- `connectonion/useful_plugins/tool_approval/approval.py` — the guarded console helper
- `docs/useful_plugins/tool_approval.md`, `docs/features/permissions.md` — the new rules
- tests as above

https://claude.ai/code/session_01VLqqyEND1RpEtSPTe86NsK

### DIFF
--- a/connectonion/useful_plugins/tool_approval/approval.py
+++ b/connectonion/useful_plugins/tool_approval/approval.py
@@ -303,6 +303,21 @@ def _get_batch_remaining(agent: 'Agent', current_tool_id: str) -> list:
     return []
 
 
+def _log_permission_granted(agent: 'Agent', tool_name: str, tool_args: dict, source: str, reason: str) -> None:
+    """Show a grant on the console when there is one.
+
+    A `quiet=True` agent has a logger whose `console` is None, and that is the
+    shape every unattended run takes. Six call sites reached
+    `agent.logger.console.log_permission_granted(...)` directly, so in a quiet
+    agent every auto-approved tool call died with AttributeError — the
+    approval said yes and the tool still failed. Found by the Rust e2e, not by
+    the sixty unit tests, none of which built a real quiet Agent.
+    """
+    console = getattr(getattr(agent, 'logger', None), 'console', None)
+    if console is not None:
+        console.log_permission_granted(tool_name, tool_args, source, reason)
+
+
 def _log(agent: 'Agent', message: str, style: str = None) -> None:
     """Log message via agent's logger if available.
 
@@ -423,7 +438,7 @@ def check_approval(agent: 'Agent') -> None:
         )
         if direct_tool == tool_name and tool_name in {'codex', 'claude_code'}:
             if getattr(getattr(agent, 'logger', None), 'console', None):
-                agent.logger.console.log_permission_granted(
+                _log_permission_granted(agent, 
                     tool_name,
                     tool_args,
                     'host',
@@ -456,7 +471,7 @@ def check_approval(agent: 'Agent') -> None:
                 )
             if verdict == 'allow':
                 if hasattr(agent, 'logger') and agent.logger and hasattr(agent.logger, 'console'):
-                    agent.logger.console.log_permission_granted(
+                    _log_permission_granted(agent, 
                         tool_name, tool_args, 'policy', policy.get('reason', 'auto-approved')
                     )
                 return
@@ -490,7 +505,7 @@ def check_approval(agent: 'Agent') -> None:
                 permitted, reason, source = check_bash_chain_permitted(tool_args['command'], permissions)
                 if permitted:
                     if getattr(getattr(agent, 'logger', None), 'console', None):
-                        agent.logger.console.log_permission_granted('bash', tool_args, source, reason)
+                        _log_permission_granted(agent, 'bash', tool_args, source, reason)
                     return
 
             # Check each permission in the dict
@@ -519,7 +534,7 @@ def check_approval(agent: 'Agent') -> None:
                     reason = perm.get('reason', 'unknown')
                     source = perm.get('source', 'config')
                     if getattr(getattr(agent, 'logger', None), 'console', None):
-                        agent.logger.console.log_permission_granted(tool_name, tool_args, source, reason)
+                        _log_permission_granted(agent, tool_name, tool_args, source, reason)
                     return
 
     # =================================================================
@@ -533,7 +548,7 @@ def check_approval(agent: 'Agent') -> None:
         tool_name = pending['name'] if pending else 'unknown'
         tool_args = pending.get('arguments', {}) if pending else {}
         if getattr(getattr(agent, 'logger', None), 'console', None):
-            agent.logger.console.log_permission_granted(tool_name, tool_args, 'mode', 'full_access mode')
+            _log_permission_granted(agent, tool_name, tool_args, 'mode', 'full_access mode')
         return
 
     # reject_hard was set by a previous tool in this batch — reject remaining
@@ -559,7 +574,7 @@ def check_approval(agent: 'Agent') -> None:
     if mode == AUTO:
         if tool_name in FILE_EDIT_TOOLS:
             if getattr(getattr(agent, 'logger', None), 'console', None):
-                agent.logger.console.log_permission_granted(
+                _log_permission_granted(agent, 
                     tool_name, tool_args, 'mode', AUTO
                 )
             return
--- a/connectonion/useful_plugins/tool_approval/policy.py
+++ b/connectonion/useful_plugins/tool_approval/policy.py
@@ -248,26 +248,96 @@ def _classify_single_command(command: str, root: Path | None = None) -> dict:
     return decision("command", "ask", "command is outside the focused verification and read-only allowlists", "call", requires_human=True)
 
 
-_OUTPUT_REDIRECT = re.compile(r">(?!&\d)")
+def _redirect_targets(command: str) -> list[str]:
+    """The files a command's output redirects write to.
+
+    `2>&1` duplicates a descriptor and writes nothing; `> >(cmd)` is process
+    substitution, and the inner command is classified on its own by
+    _extract_subcommands. Everything else — `>`, `>>`, `2> file` — is a file
+    being written, which bashlex keeps out of the word list, so this is the
+    one place a write hiding behind a read-only command is seen.
+    """
+    import bashlex
+
+    targets: list[str] = []
+
+    def visit(node):
+        if node.kind == "redirect" and node.type in (">", ">>") and hasattr(node.output, "word"):
+            if not node.output.word.startswith(">("):
+                targets.append(node.output.word)
+        for attr in ("parts", "list"):
+            for child in getattr(node, attr, None) or []:
+                visit(child)
+        for attr in ("command", "output", "input"):
+            child = getattr(node, attr, None)
+            if child is not None and hasattr(child, "kind"):
+                visit(child)
+
+    for node in bashlex.parse(command):
+        visit(node)
+    return targets
 
 
-def _has_output_redirect(command: str) -> bool:
-    """A `>` that points at a file means something is being written, so
-    nothing about the command is read-only — even `head -1 f > out`. `2>&1`
-    and `>&2` only duplicate a descriptor and write nothing new. bashlex keeps
-    redirects out of the word list, so this is the one place they are seen."""
-    return bool(_OUTPUT_REDIRECT.search(command))
+def _redirect_verdict(targets: list[str], root: Path | None) -> dict | None:
+    """A redirect is a file write, and is held to the write tool's rules.
+
+    Inside the workspace it is a reversible edit and allowed — `echo x > f`
+    is what a model reaches for instead of the write tool. A control file is
+    denied, a target outside the workspace is denied, and a target that
+    depends on the environment (`$HOME/...`) cannot be resolved and asks.
+    """
+    from .approval import _is_control_file
+
+    for target in targets:
+        if "$" in target or "`" in target:
+            return decision("command", "ask", "redirect target depends on the environment", "call", requires_human=True)
+        if root is None:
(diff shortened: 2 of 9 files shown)


wu-changxing:
Verification of the fix, beyond the suite.

**The issue's own table, shipped permissions, `io=None`, expected verdict printed beside actual** — 20 cases, all correct:

| command | verdict | rule |
|---|---|---|
| `CO_WHO=x co browser -t t get_text` | allow | configured_command |
| `CO_WHO=x co browser -t t get_text \| head -40` | allow | configured_command |
| `CO_WHO=x co browser -t t get_text \| grep -i foo` | allow | configured_command |
| `co browser -t t run_page_script a.js && echo ok` | allow | configured_command |
| `head -40 notes.txt` / `grep -rn foo . \| wc -l` | allow | read |
| `cd greeter && cargo test 2>&1 \| tail -20` | allow | verification |
| `echo hi > notes.txt` | allow | workspace_edit |
| `cat /etc/hosts` | deny | read_outside_workspace |
| `head ~/.ssh/id_rsa` / `cat .co/keys/agent.key` | deny | credentials |
| `cat .env` | deny | credentials |
| `sed -i s/a/b/ notes.txt` | deny | workspace_edit |
| `echo x > ../outside.txt` | deny | write_outside_workspace |
| `echo 'Bash(*)' > .co/host.yaml` | deny | authorization_control |
| `rm -rf greeter` | deny | deletion |
| `co browser status && co email send …` | deny | command |
| `co deploy` / `curl …` / `bash << EOF … EOF` | deny | publication / external_network / command |

**Through the real `co ai`, unattended, in a temp project** — not the test harness:

```
$ co ai "Run exactly this one shell command and show me its output verbatim: cat notes.txt | head -3"
[co]   ⚡ bash: cat notes.txt | head -3   policy read-only command chain (2 commands)
alpha
beta
gamma
```

```
$ co ai "Show me the first 2 lines of server.pem, then the first 2 lines of notes.txt"
[co] ✗ Error: Tool 'read_file' denied by connectonion.auto: credential access is never auto-approved
[co]   ⚡ read_file(path="notes....")   policy read-only workspace operation
alpha
beta
```

Two holes came out of that verification, both now fixed with fail-first tests, both pushed here:

1. **`head ~/.ssh/id_rsa` was allowed.** The outside-workspace rule was meant to catch it and normally does, because `~` is not the project — but the workspace is sometimes the home directory, keys get committed, and `.co/keys/agent.key` is always under the project root. The credential check only looked for `.env`/`secret`/`credential` among the words. Key material is now denied by path component, filename and suffix, wherever it sits.
2. **`cat server.pem` was denied while `read_file("server.pem")` was allowed.** That is how the second CLI run above started: asked for the shell form, the model reached for the tool form, and the policy said yes. It declined to print the contents on its own initiative, which is the only reason nothing leaked. The read/write/edit/delete tools now refuse key material by the same rule.

Offline suite: 8741 passed, 21 skipped.

wu-changxing:
Closed the hole this PR's own "least confident about" named.

> `awk` can call `system()` and `sed` has `e` on GNU, neither of which this checks.

Ran it:

```
awk 'BEGIN{system("rm -rf /")}'   →  allow
```

That was the one part of this change that made the system less safe than 1.8.4. Keeping the two on a read-only list while excluding their execution constructs means writing a parser for two languages in a security path and being right about it against prompt injection, so both come off the list: **nothing that takes a program text is read-only.** They ask, as they did before 1.8.5, including `sed -n 1,10p` and `awk '{print $1}'` — whose job `head`, `tail`, `cut`, `jq` and `read_file(limit=, offset=)` already do. Nothing else on the list takes a program.

Verified through the real `co ai`, unattended:

```
$ co ai "Run exactly this one shell command with bash and show its output: awk 'BEGIN{system(\"echo pwned\")}'"
[co] ✗ Tool 'bash' denied by connectonion.auto: command is outside the focused verification and read-only allowlists
```

Offline suite: 8741 passed.

wu-changxing:
Adversarial sweep before publishing, and two pre-existing holes it found.

Twenty-six bypass attempts against the shipped permissions with no approval channel. **Twenty-five refused:**

`cd /etc && cat passwd` · `cd ..` · `co browser status | curl -d @-` · `co browser status && wget --post-file=-` · `cat notes.txt | ssh box …` · `grep -f /etc/passwd` · `jq --slurpfile s /etc/passwd` · `diff notes.txt /etc/passwd` · `tee out.txt` · `cp`/`mv`/`ln -s`/`install`/`dd` to `/tmp` · `sh -c 'rm -rf /'` · `env sh -c` · `nohup rm -rf /` · `eval rm -rf /` · `perl -e` · `python3 -c` · `node -e` · `./target/debug/greeter` without its grant · `head -1 f && rm -rf greeter` · `head -1 f; curl …`

**The twenty-sixth was `make install`** — and 1.8.4 allowed it too, along with bare `make` and `make -C /etc all`, as "focused test, lint, or build command". `cargo` was already limited to test/check/clippy/build, `go` to test, the package runners to test/lint/build/check/typecheck; `make` was the one left open. `make` now needs a named verification target with no `-C`/`-f` redirecting it. Pushed here with tests both ways.

To be precise about what that does and does not fix: focused verification is an execution surface by design — anything that may write a test and run it may run what the test runs. The narrowing removes the commands in that category that were never verification.

**The sweep also found `cat $(cat which_file.txt)` allowed.** The outside-workspace check looks at arguments that look like paths; a substitution does not, and it reads whatever the file names. Two neighbours had been refused and I had read that as the rule holding — `head $(echo /etc/shadow)` because bashlex splits the word and the tail starts with a slash, `cat ${SECRET_PATH}` because it contains "secret". Luck, twice. A file argument containing `$(`, a backtick or `$name` now asks; a bare `$` is end-of-line, so `grep 'foo$' f` is untouched.

Running tally for this PR: the reported bug, five defects found while fixing it, two pre-existing holes in the code it sits next to. One was found by a test; the other seven by running the thing. All eight have a test that was red before the fix.

Offline suite: 8760 passed, 21 skipped.

PR #1480, opened 2026-09-11
[1.8.5] The queue is an inbox, and something finally consumes it

wu-changxing:
Closes #1478.

- **Proposed target version:** 1.8.5, first published as `1.8.5a1`
- **Estimated release window:** next alpha — as soon as #1462 gate 1 passes live. This PR does not change receipt, so it neither fixes nor blocks that gate.

Stacked on #1472 (`fix/185-integration`). Merge that first; this branch is based on it.

## Why now

#1462 freezes the directory and verb contract at 1.8.5. Everything here changes names and adds consumers, so it has to land before `1.8.5a1` or it cannot land at all.

## What changed

**The queue is an inbox, under one root.** `mailbox` already meant Gmail in this repository (`useful_tools/gmail_mailbox.py`, `MailboxError`, `co email`), so the queue had borrowed a word that was taken and every reader had to work out which one a line meant. It is now `inbox`, which this repository already uses for messages received in `co email inbox` and `co sms inbox`.

| before | after |
|---|---|
| `connectonion/listen/`, `Mailbox` | `connectonion/inbox/`, `Inbox` |
| `~/.co/feishu/`, `~/.co/lark/` | `~/.co/inbox/<provider>/` |
| `inbox.jsonl`, `outbox.jsonl` | `received.jsonl`, `sent.jsonl` |
| `CO_FEISHU_HOME`, `CO_LARK_HOME` | `CO_INBOX_HOME` |

One root so a consumer watches `inbox/*/new/` instead of a list somebody keeps in sync, and `ls ~/.co/inbox` answers which channels a machine listens to. One variable because moving one provider and leaving the others only ever produced a half-configured machine. The verbs do not change: `listen` is the action, `inbox` is the place.

**`Inbox.serve(handler)` — one lane per conversation.** Ordered inside a lane, parallel across lanes; Kafka's partition-by-key and Erlang's process-per-session. A pool was the obvious alternative and it answers two messages from one person at once, in the wrong order. The handler owns the outcome and that is the whole contract — returning means finished however it was finished, raising means it waits in `cur/` for the sweep — which is why one loop drives a subprocess, an Agent or a Host without knowing which it has.

**A lease instead of a guess.** The stale window was a guess at how long an answer takes, documented as a guess in DD-063. A handler slower than the guess had its message taken away mid-sentence and answered twice. `renew()` touches the file while a handler runs, so the window measures silence rather than work. This is SQS's `ChangeMessageVisibility`, which we had copied the timeout from without copying the renewal.

[...]

PR #1484, opened 2026-09-11
[1.8.5] co auth feishu: scan once, and the app exists

wu-changxing:
- **Proposed target version:** 1.8.5, first published as `1.8.5a1`
- **Estimated release window:** next alpha, with #1480 and #1472

Stacked on #1480 (`feat/1.8.5-inbox`), which is stacked on #1472. Merge in that order.

## The problem

Connecting a Feishu bot took eleven steps, and step four decided whether anyone finished: create a self-built app in the console, add three scopes whose names have to be exactly right, choose long connection, subscribe to one event, publish, then copy an id and a password into `~/.co/keys.env`.

The eleven steps are not the problem. Nothing tells you when one of them is wrong. A missing scope is not an error at setup time — it is an error an hour later, in a log, when somebody @-mentions the bot and nothing happens.

## What this adds

```bash
pip install lark-oapi
co auth feishu
```

A QR code in the terminal and a link beside it. Scan, approve, and the application exists — in your own tenant, owned by you — with `FEISHU_APP_ID` and `FEISHU_APP_SECRET` written to the selected env file. No console, no copy, no paste. The flow begins on Feishu and moves to Lark by itself if that is where your tenant lives, so there is nothing to choose up front either; whichever it turns out to be is what gets written.

## Why we thought this was impossible

Worth recording, because the reasoning was wrong in an ordinary way.

`application/v6` returns an application's info and no `app_secret` — correct, an API that hands out credentials for an app you name is a vulnerability — and it needs a `tenant_access_token`, which you cannot get without the secret. We verified that exhaustively. Feishu's own CLI does the setup in one step anyway, and its binary contains `/open-apis/application/v6/larksuite_cli_app/probe`, an endpoint named after the CLI itself. That read as the private back door that explained the discrepancy.

The CLI is MIT licensed. `larksuite_cli_app/probe` is a best-effort telemetry ping fired *after* credentials are saved, whose result the code explicitly discards. The real flow is forty lines above it:

```
POST /oauth/v1/app/registration   action=begin  archetype=PersonalAgent  auth_method=client_secret
  → device_code, user_code, verification_uri
                                  (person scans and approves)
POST /oauth/v1/app/registration   action=poll  device_code=…
  → client_id, client_secret
```

The begin request carries no client id, key or signature — nothing that says which program is asking. It is OAuth 2.0's device authorization grant pointed at app creation instead of login. It is documented on open.feishu.cn under 集成智能体到飞书, and wrapped by all three official SDKs, including the Python one this project already depends on.

So the implementation is `lark_oapi.register_app` plus a QR renderer. `qrcode>=8.2` was already a dependency; `lark-oapi` stays optional and a missing one exits 3 with the pip command, the same shape `co feishu listen` uses.

## `co env set` refuses these two names

`FEISHU_APP_ID`, `FEISHU_APP_SECRET`, `LARK_APP_ID`, `LARK_APP_SECRET` are now refused with `Next: co auth feishu`, the way the Google and Microsoft records are.

The reason is different and the message says so. A Google record field is refused so one field cannot describe a different account than the rest. An app secret is refused because a hand-typed one came from somewhere the command cannot check — and there is now a command that can. They are not OAuth records (two names, not five), so they live in their own small table rather than in `PROVIDER_FIELDS`.

## What one scan does not do

- **It cannot list applications you already own** — no API does — so it always creates a new one, and choosing between applications happens on Feishu's own confirmation page. That is the right place for that decision.
- **The preset carries more permissions than a message bot needs**, and the SDK's `addons` parameter only adds; there is no way to narrow it. In a company tenant an administrator may have to approve them. **Not tested, and not claimed.**
- **The manual path is still documented**, one collapsed `<details>` section down, for anyone with an application already.
- One unverified detail from the local investigation: an app created this way opens the long connection and subscribes to `im.message.receive_v1` (`[source] feishu-websocket: connected`), but `/open-apis/bot/v3/info` answered `20008` on it, where a console-made bot answers normally. Whether it can be invited to a group and @-mentioned is a ten-minute live test that has not been run.

## Evidence

```
tests/unit              9011 passed, 15 skipped   (24s)
tests/e2e                596 passed,  7 skipped   (55s)
```

`tests/unit/test_feishu_auth.py`, 12 cases written failing first: the link is printed for a terminal with no camera beside it; a QR is drawn; the app is named so it is findable later; credentials land in the selected file; a Lark tenant writes the `LARK_*` names; **the secret is never printed**; a missing SDK exits 3 with the pip command; a denied or expired scan is one line and exit 1, not a traceback; **a failed scan writes nothing**; a response missing either half is refused; and `co env set` refuses both names.

CLI surface, against a built CLI in a temp HOME:

| check | result |
|---|---|
| `co auth --help` lists feishu and lark | yes |
| `co env set FEISHU_APP_SECRET` | exit 2, names `co auth feishu` |
| `co env set LARK_APP_ID` | exit 2, names `co auth lark` |
| `co env set OPENAI_API_KEY` still saves | exit 0 |
| Google record still refused | exit 2, names `co auth google` |

Two failures on this machine are not from this branch: `test_co_command_works` fails identically on a clean `origin/main` checkout, and `test_native_egress_preflight_real_driver` needs real network.

## Checklist
- [x] I proposed a target version and release window above
- [x] Any dependent changes have been merged and published — no: stacked on #1480, then #1472
- [ ] Maintenance-branch forward-port tracker — N/A, mainline work
- [x] Evidence the linked issue asks for: test counts, the CLI table, and what is not claimed

### DIFF
--- a/connectonion/cli/commands/env_commands.py
+++ b/connectonion/cli/commands/env_commands.py
@@ -39,8 +39,18 @@ console = Console()
 # line the next read refuses, and `co env` is the command that fixes those.
 _NAME = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 
-_PROVIDER_AUTH = {"GOOGLE": "co auth google", "MICROSOFT": "co auth microsoft"}
-_PROVIDER_LABEL = {"GOOGLE": "Google", "MICROSOFT": "Microsoft"}
+_PROVIDER_AUTH = {"GOOGLE": "co auth google", "MICROSOFT": "co auth microsoft",
+                  "FEISHU": "co auth feishu", "LARK": "co auth lark"}
+_PROVIDER_LABEL = {"GOOGLE": "Google", "MICROSOFT": "Microsoft",
+                   "FEISHU": "Feishu", "LARK": "Lark"}
+
+# An application's credentials are not an OAuth record — two names, not five —
+# so they are listed here rather than in PROVIDER_FIELDS. They are refused for
+# the same reason: one command writes them, and that command can also create
+# the application they belong to.
+_APP_CREDENTIALS = {f"{prefix}_{field}": prefix
+                    for prefix in ("FEISHU", "LARK")
+                    for field in ("APP_ID", "APP_SECRET")}
 
 
 def _provider_of(key: str) -> str | None:
@@ -189,6 +199,12 @@ def handle_env_set(key: str, value: str) -> None:
         _fail("AGENT_CONFIG_PATH chooses which global directory is read, so a file inside it "
               "cannot set it. Export it in your shell instead:\n"
               "  export AGENT_CONFIG_PATH=/path/to/.co\nNext: co env", 2)
+    app_provider = _APP_CREDENTIALS.get(key)
+    if app_provider is not None:
+        auth = _PROVIDER_AUTH[app_provider]
+        _fail(f"{key} is written by {auth}, which also creates the application it belongs to. "
+              f"Feishu has no API that hands out an app secret, so a hand-typed one came from "
+              f"somewhere this command cannot check. Next: {auth}", 2)
     provider = _provider_of(key)
     if provider is not None:
         auth = _PROVIDER_AUTH[provider]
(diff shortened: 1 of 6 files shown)


wu-changxing:
Subsumed by #1491, merged to main as f74518120b. Every commit on this branch is in main; GitHub did not close it automatically because its base was the stacked feat/1.8.5-inbox rather than main.

PR #1489, opened 2026-09-11
Release 1.8.5a2: a refusal you can act on

wu-changxing:
## Description

Publishes **1.8.5a2**, an opt-in preview. Version bump only — the change it carries merged as #1487.

1.8.5a1 fixed what the permission policy *decides*. This carries the fix for what it *says* when it decides no.

## Why a2 rather than folding into a1

a1 is already published, and a refusal that cannot be acted on is exactly what breaks during the exercise a preview exists for. An agent handed the old one-sentence refusal has one move available — try again — and trying again cannot work, because the decision is a lookup against a static list. The incident behind #1481 reached 28 of 300 iterations that way: one refusal and 272 retries of it.

## What ships

Everything in [1.8.5a1](../blob/main/docs/releases/1.8.5a1.md), plus:

- Every policy refusal carries a `<system-reminder>` with why, the exact permission line and the two places it can go, and a re-think prompt saying what to try instead — per effect class.
- The suggested grant is verified against the real grant path before printing, in the order the live call asks. Two of the first suggestions were untrue; a refusal no pattern can lift now says so.
- blog-gate's story check no longer reads a provider 503 as a failing post.

- **Suggested labels:** tests
- **Proposed target version:** 1.8.5a2 (this PR is the release)
- **Estimated release window:** on merge
- **Milestone:** maintainer assigns
- **Why this release:** the preview is published for exercise, and this is what makes a refusal during that exercise actionable instead of a silent stop. No change to what is allowed or denied. Rollback is a revert plus a yank.

## How this was written

- [x] Mostly AI-generated — Claude Opus 5 via Claude Code, on the owner's standing instruction to publish this line.

**Least confident about:** the reminder's cost. Roughly 25 lines per refusal in the message stream; a run hitting several refusals pays for all of them. Judged worth it against a retry loop, but it is a judgement.

**Not verified:** whether a smaller model honours "do not retry". Verified on `gemini-3.8-flash` — one attempt, no retry, told the operator why.

**If this is wrong, what breaks first:** `test_the_grant_the_refusal_suggests_actually_allows_the_call` — sixteen refused calls whose suggested grant must actually lift them.

## Dev Blog

Shipped with #1487: `docs/blog/2026-09-11-the-remedy-was-a-lie.md`. This PR is the version bump, so it carries no new post — apply `no-blog` if the gate objects.

## Testing

```
$ make test
8844 passed, 21 skipped, 6 warnings in 40.19s
```

Version gates pass (`test_the_version_agrees_with_itself`, `test_every_release_has_an_entry`); the docs-site channel check skips, as noted for a1.

## Follow-up still open

`docs-site/lib/version.ts` needs `PREVIEW_VERSION = '1.8.5a2'`. That repo is not checked out beside this one here.

https://claude.ai/code/session_01VLqqyEND1RpEtSPTe86NsK

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -58,7 +58,38 @@ See [1.8.4 notes](docs/releases/1.8.4.md) for migration and acceptance limits.
 The planned 1.8.4b1 was not published separately; its reviewed changes are included
 in 1.8.4. Publication is performed and verified by the immutable-tag workflow.
 
-## Release candidate: 1.8.5a1 (preview, published)
+## Release candidate: 1.8.5a2 (preview, published)
+
+Every refusal the policy makes now says why, how to allow it next time, and
+to re-think rather than retry. 1.8.5a1 fixed what the policy decides; this
+fixes what it says when it decides no. An agent handed "command is outside the
+focused verification allowlist" has one move available — try again — and
+trying again cannot work, because the decision is a static lookup. That is how
+the incident behind #1481 reached 28 of 300 iterations: one refusal and 272
+retries of it.
+
+The suggested grant is verified against the real grant path before it is
+printed, in the order the live call asks — control-file gate first, then the
+grant check. Two of the first suggestions were untrue: `Bash(cat *)` cannot
+allow `cat /etc/shadow`, because a wildcard is honoured only for the effect
+its own text names, and no pattern at all can allow `echo x > ../outside.txt`,
+because redirects are stripped from the text patterns match against. Those now
+say there is no grant rather than naming a line that would not work — a wrong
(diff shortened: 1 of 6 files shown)


wu-changxing:
Superseded. 1.8.5a2 was published from #1490 at 05:27 UTC on 11 September — PyPI has the artifact and GitHub has the v1.8.5a2 pre-release. The change this PR was cutting a version for (#1487, every refusal says why and how to allow it) is on main as 12b5205e and is in that published a2, so there is nothing left to release here.

Two sessions were preparing the same version bump at the same time; that is the thing to avoid rather than either PR.

PR #1515, opened 2026-09-12
feat(inbox): record which consumer handled each message

### LINKED ISSUE #1503: [1.8.5] The inbox record does not say which consumer handled a message
## Outcome

The inbox's own record says **which consumer** handled each message. Today it does not.

```
done.jsonl   {"id": "om_9f8e", "at": "2026-09-12T02:31:07Z"}
sent.jsonl   {"at": …, "chat": "oc_a1b2", "reply_to": "om_9f8e", "ok": true, "id": "om_new"}
             ^ nothing here says who
```

Verified against `inbox/store.py` on main: `done()` writes `{"id", "at"}` and `record_sent()` writes seven fields, none of which identify the consumer.

## Why it matters now rather than later

Several consumers of one directory is not an edge case, it is the design. DD-063's whole argument is that the tool writes files and consumers arrive by themselves, and three ship today: `co ai --listen`, the Host lifespan, and `co <provider> consume -- <command>` with Claude Code, Codex or a script behind it. Taking a message is `rename(2)`, so they cannot collide — that part works and is tested.

What they can do is leave an indistinguishable trail. During development the normal setup is a Host answering in one terminal while a person drives `co feishu receive` in another to inspect something. When a group gets an answer nobody expected, or does not get one, the question is *which of them had it* — and the files cannot say.

The same gap makes a real failure unreadable. A message that came back after its lease expired, was picked up by a different consumer and answered twice looks in `sent.jsonl` like two replies with no explanation of why. With the consumer recorded it reads as what it is.

## Shape

One field on the durable records, and one line in the output.

- `done.jsonl` and `sent.jsonl` gain `by`: a short string naming the consumer — `host`, `co-ai`, `receive`, or `consume:<command basename>`.
- `Inbox.serve(handler, by=…)` takes it, so a consumer names itself once rather than at every call site. `receive` and `reply` invoked straight from a shell name themselves from the verb.
- `co <provider> ls` and `co <provider> log` show it where they already show a message's state.
- It is descriptive, not authorization: nothing decides anything on the strength of it, and a consumer that lies only mislabels its own work. Who may command an agent is #1479.

## Definition of done

- [ ] `done.jsonl` and `sent.jsonl` carry `by` for every writer in the tree — the Host lifespan, `co ai --listen`, `consume`, and a bare `receive`/`reply`/`done` from a shell
- [ ] A record written by an older version with no `by` still reads (absent, not crashing)
- [ ] Two consumers on one inbox, one message each: the records distinguish them, and a test asserts it
- [ ] A message handed out twice after a lease expiry shows both consumers
- [ ] `co feishu log` shows who handled the last few
- [ ] Nothing branches on the value

## Not in scope

Anything that treats `by` as identity or permission. It is a label on work that happened, the way a commit's author is.

wangzhengzhuo05:
## What

`done.jsonl` and `sent.jsonl` now carry a descriptive `by` field naming the consumer that handled or sent a message, so several consumers sharing one inbox directory leave a distinguishable trail instead of an identical one.

Values: `host`, `co-ai`, `consume:<command basename>`, and `send` / `reply` / `done` for the bare shell verbs.

## Why

Several consumers of one inbox is the design (`co ai --listen`, the Host lifespan, and `co <provider> consume -- <command>`). They cannot collide on a message — taking one is `rename(2)` — but the records they left said nothing about who did what, so "which of them had it?" was unanswerable.

## How

- `connectonion/inbox/store.py`: `done()` and `record_sent()` take an optional `by` and include it when supplied; `recent_records()` reads the last few lines of a JSONL file for the display.
- `connectonion/inbox/consumer.py`: `serve()` takes `by` and threads it to the loop, which stamps both completion paths (the success path and the give-up-after-max-attempts path).
- `connectonion/cli/commands/listen_commands.py`: shell verbs name themselves (`send`, `reply`, `done`), `consume` derives `consume:<basename>` once, and `co <provider> log` shows who handled the last few.
- `connectonion/cli/co_ai/listen.py` and `connectonion/network/host/inbox.py`: the `co-ai` and `host` consumers name themselves at their `serve()` and `record_sent()` call sites.

The field is optional, so records written by an older version still read (absent, not crashing). It is descriptive only: nothing branches on its value, and a consumer that lies only mislabels its own work.

Touched files were left clean under the repo's ruff config (import sorting; one unused `typing.Optional` import in `network/host/inbox.py`).

## Tests

`tests/unit/test_inbox_store.py` gains two tests using the existing helpers:

- two consumers on one inbox, one message each — the `done.jsonl` and `sent.jsonl` records carry the right `by`, and a record written without `by` still reads;
- a message handed out twice after a lease expiry shows both consumers in `completed`.

`python -m pytest tests/unit/test_inbox_store.py tests/unit/test_listen_commands.py -q` -> **58 passed**. `ruff check` clean on every changed file.

Mutation check: dropping the two `by` writes in `store.py` makes both new tests fail; restoring them makes them pass. The tests are load-bearing.

Fixes #1503

---
Implemented with AI assistance (OpenCode + muse-spark), then reviewed and verified locally (targeted tests, lint, mutation check) before submission.

**Proposed target version:** next minor / next alpha
**Estimated release window:** future roadmap / unknown date

### DIFF
--- a/connectonion/cli/co_ai/listen.py
+++ b/connectonion/cli/co_ai/listen.py
@@ -17,10 +17,10 @@ than corrupting it.
 """
 
 import threading
-from pathlib import Path
-from typing import Callable, List, Optional, Sequence
+from typing import Callable, Optional, Sequence
 
-from ...inbox import Inbox, provider as _provider
+from ...inbox import Inbox
+from ...inbox import provider as _provider
 from ...inbox.settings import Channel
 
 
@@ -67,9 +67,11 @@ def _handler(channel: Channel, inbox: Inbox, provider, agent, sessions: dict, gu
         try:
             sent = provider.send(message.chat, reply, reply_to=message.id)
         except Exception as exc:
-            inbox.record_sent(chat=message.chat, text=reply, reply_to=message.id, error=str(exc))
+            inbox.record_sent(chat=message.chat, text=reply, reply_to=message.id,
+                              error=str(exc), by="co-ai")
             raise RuntimeError(f"reply failed: {exc}") from exc
-        inbox.record_sent(chat=message.chat, text=reply, reply_to=message.id, provider_id=sent)
+        inbox.record_sent(chat=message.chat, text=reply, reply_to=message.id,
+                          provider_id=sent, by="co-ai")
 
     return answer
 
@@ -95,7 +97,7 @@ def listen(channels: Sequence[Channel], agent_factory: Callable, *,
         loops.append(threading.Thread(
             target=inbox.serve, args=(handler,),
             kwargs={"workers": workers, "idle_seconds": idle_seconds,
-                    "should_stop": stop, "once": once},
+                    "should_stop": stop, "once": once, "by": "co-ai"},
             name=f"listen-{channel.provider}", daemon=True))
 
     for loop in loops:
--- a/connectonion/cli/commands/listen_commands.py
+++ b/connectonion/cli/commands/listen_commands.py
@@ -8,7 +8,6 @@ LLM-Note:
   Errors: a missing credential prints the item and the next action and exits 3 | a provider refusal prints its own words and exits 1 | nothing is printed on the success path of listen (Rule of Silence); the log has it
 """
 
-import json
 import os
 import shutil
 import subprocess
@@ -66,7 +65,7 @@ def _listener_or_exit(inbox: Inbox) -> None:
 
 def handle_done(name: str, message_id: str) -> None:
     """Forget a taken message without replying, so it does not come back."""
-    Inbox(name).done(message_id)
+    Inbox(name).done(message_id, by="done")
 
 
 def handle_listen(name: str, raw: bool = False) -> None:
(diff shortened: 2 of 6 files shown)


wu-changxing:
Integrated as #1556 — cherry-picked onto stable 1.8.5 with your commit and authorship intact.

This PR could not merge itself: it is a fork branch, so I cannot push a rebase to it, and it had gone `CONFLICTING` against work merged into 1.8.5 today. The conflict was one hunk in the import block of `listen_commands.py` — you removed `json` (your change took away its last use) and I added `re` a few hours ago for the missing-scope matcher in `co <provider> check`. Both kept.

9,493 unit tests pass on the integrated branch.

Thank you — this closes #1503, which I had written into 1.8.5 as a known limit earlier today without realising the fix was already waiting. Closing here in favour of #1556; the commit that lands is yours.

wangzhengzhuo05:
Thanks @wu-changxing — closing here in favour of #1556, which you already cherry-picked with the commit and authorship intact.

Appreciate you tracking down the import-block conflict (the `json` removal on my side vs the `re` addition for the missing-scope matcher) and keeping both — that resolution is exactly right, and I would not have been able to push a rebase to a fork branch either.

Thanks also for closing #1503 as completed rather than leaving it as a known limit.

PR #1522, opened 2026-09-13
[1.8.6] Mail listings take a date window, and Outlook answers in JSON

### LINKED ISSUE #1521: [1.8.6] Gmail / Outlook date-window listings and machine-readable output
## Current release decision — 2026-09-15

Target **1.8.6**, tracked by #1555; this supersedes the historical 1.8.5 target. Continue #1522 and complete Gmail date-window + JSON/pagination composition while preserving the existing envelope and accurate completeness. Independent of Wiki, which is 1.8.7.

---

## The mail CLIs cannot express the query an agent needs

`co outlook inbox` takes `-n <count>` and `--unread`. That is the whole surface:

```
Options:
  --last    -n  <int>   How many emails to show [default: 10]
  --unread  -u          Only unread emails
```

Two things are missing, and both block an agent driving the CLI rather than a
person reading it.

**No date range.** "Everything between two dates" is the natural unit for any
sweep — a census of who you correspond with, a backfill, a weekly digest. With
only `-n`, the caller asks for 5000 and discards most of the answer. On a real
mailbox that is 1851 messages fetched to read a 150-day window.

**No machine-readable output.** The command prints a Rich table sized for a
terminal. A caller that wants sender, date and subject has to scrape it, and
the scrape breaks whenever the table changes. Every other agent-facing surface
we ship (`co ai --json`, `co browser`, `co call`) answers in JSON.

The same holds for `co gmail`.

### What the library has, and where

`list_between(start, end, max_results)` and `my_addresses()` exist on both
`Outlook` and `Gmail` — but **only on the unmerged Wiki branch**, added in
80d7a926 for the Wiki sources. On `main` neither class has them.

(An earlier draft of this issue said they were already on `main` and that this
was only a wiring job. That was read off the Wiki worktree without checking
which branch it was.)

So the work is two steps: port those two methods to `main` verbatim from that
branch — verbatim so the eventual Wiki merge has identical text and nothing to
reconcile — then wire the flags. They were never Wiki-specific: anything that
reads mail by time window needs them, and locking them inside a 6,785-line
feature branch is the wrong place for them.

### Proposed

```
co outlook inbox --since 30d --json
co outlook inbox --since 2026-06-01 --until 2026-09-01 --json
co outlook sent  --since 7d --json
co gmail   inbox --since 30d --json          # same flags, same shapes
```

- `--since` accepts `Nd` / `Nw` / `Nm` or an ISO date; `--until` defaults to now.
- `--json` prints one array of objects with the fields the library already
  returns (`id`, `from`, `to`, `date`, `subject`), nothing reformatted.
- `-n` keeps working and keeps its meaning: a cap, not a window.
- Without `--json` the table is unchanged, so nobody's habits break.

### Why now

The Wiki's first stage is a census over both mailboxes, and the design is that
it runs as a Skill driving the CLI rather than as another Python entry point —
the Skill is the interface, the library is the engine. A Skill cannot drive a
CLI that has no date range and no JSON, which is why the prototype went around
the CLI and called the Python classes directly. That prototype should be thrown
away, and it can be as soon as these flags exist.

Target: 1.8.5.

wu-changxing:
## Current release decision — 2026-09-15

Target **1.8.6** under #1555, superseding 1.8.5 below. Finish Gmail date-window + JSON/pagination composition; the current refusal remains a documented gap, not completed acceptance. Wiki is independently scheduled for 1.8.7.

**Proposed target version:** 1.8.6a1
**Estimated release window:** next preview, within days

**Rebased onto stable 1.8.5** (`a57ff6f7`) on 2026-09-16. It had been sitting on 1.8.5b4, twenty commits back, which is why its four `test` jobs were red — measured after the rebase: 9,503 unit tests pass locally with only the four known cross-test-pollution failures. A Design Journal post is added for the blog gate.

---
---

Closes #1521. Target 1.8.5.

## What changes

```bash
co outlook inbox --since 30d --json
co outlook inbox --since 2026-06-01 --until 2026-09-01
co gmail   inbox --since 2w
```

`--since` takes `Nd` / `Nw` / `Nm` / `Ny` or a date; `--until` defaults to now.
`-n` keeps its meaning as a cap, and without the new flags both commands behave
exactly as before.

## Why the flags could not just be wired up

The issue originally said the library already had `list_between` and this was a
wiring job. That was read off the unmerged Wiki branch — on `main` neither
provider had it. So this PR brings `list_between` and `my_addresses` over for
both Outlook and Gmail, **ported verbatim from 80d7a926**, so that when the Wiki
branch merges the two sides are identical text and there is nothing to
reconcile.

They were never Wiki-specific. Anything that reads mail by time window needs
them, and leaving them inside a 6,785-line feature branch was the wrong place.

## Outlook gets `--json`; Gmail deliberately does not

`co gmail inbox --json` already exists and answers with a versioned envelope —
`schema_version`, `provider`, `account`, `status`, `data`, `error`, plus
`--cursor` paging. Outlook had nothing. Adding a second, differently-shaped
`--json` to the same CLI would be worse than the gap, so Outlook gets a plain
array of the provider's own fields and Gmail's envelope is untouched.

Where the two cannot meet yet, the command refuses rather than misleads:

```
$ co gmail inbox --since 30d --json
--since/--until do not work with --json yet; run without --json, or see issue #1521
```

A dropped window would have returned the last ten messages and looked like it
worked.

## Verified against real mailboxes

```
$ co outlook inbox --since 2d --json -n 3
[ { "id": "AQMkAD...", "from": "express@airbnb.com", "to": ["xietianle@outlook.com"],
    "date": "2026-09-11T01:03:12Z", "subject": "RE: Reservation for QVB/...", ... } ]

$ co outlook inbox --since 上周
Cannot read '上周' as a date. Use 30d, 2w, 6m, 1y, or 2026-06-01.      (exit 2)
```

## A bug the tests found

`--since 21d` issued a fourth request for a window **0.0004 seconds wide** — the
window's start and `now` are computed a few hundred microseconds apart, so the
third boundary lands just short of the end. A round trip that can only come back
empty. The walk now stops when less than a second remains.

## Tests

- `tests/unit/test_mail_window.py` — 9: window parsing, the week-at-a-time walk
  being contiguous and forward, the cap stopping it early, a backwards window
  refused, and JSON carrying the listing fields but not message bodies
- `tests/e2e/cli/test_cli_outlook.py` — new file, 2: flags forwarded, and the
  no-window path unchanged
- `tests/e2e/cli/test_cli_gmail.py` — +2: window forwarded, and `--since --json`
  refused with a pointer to the issue

Full offline suite: 9961 passed. The 4 failures and 2 errors that remain are
identical to `origin/main`'s own baseline measured today (shell interrupt,
session-history decoy, two admin routes, and the native-egress tests that need
a real gateway and Chromium).


wu-changxing:
One more commit: `co outlook search` help now says that `from:<address>`, `to:<address>` and `participants:<address>` narrow by who. Graph's `$search` already accepted them — `from:zhouodywork@gmail.com` returned the five mails it should — but the help said only "matches subject and body", so nobody looking for one person's mail had a way to know. No code change.

Issue #1497, opened 2026-09-12
[2.x] A secret belongs in the OS keychain, not in a plaintext keys.env


wu-changxing:
## Outcome

A secret that `co` needs lives in the operating system's keychain, and `~/.co/keys.env` stops being a plaintext file full of credentials.

Today every secret this product touches — model keys, the OpenOnion token, Google and Microsoft refresh tokens, Feishu and Lark app secrets, the agent's own signing key path — sits in one `0600` text file. That is a deliberate, defensible choice: it is inspectable, portable, and identical on every platform, and `co env` exists because being able to read and repair it matters. It is also one `cat` away from being in a log, a screen share, or a model's context.

## What this is not

Not a change to how `co env` works for the person using it, and not a second place to look. `co env` should keep answering "what does this machine hold and where did it come from", whether the value is in a file or in the keychain.

## Why now, and why 2.x

Two things happened that make this concrete rather than theoretical.

**A real tool on this machine already does it, and we met the consequences.** `lark-cli` keeps its app secret in the macOS keychain — `~/.lark-cli/config.json` stores only `{"source": "keychain", "id": "appsecret:cli_…"}` — with `keychain-downgrade` as a documented escape hatch for sandboxes that cannot reach it. When we wanted to import that credential, the honest answer was that we should not: reading another tool's keychain entry and writing it into our plaintext file is a downgrade wearing the word "import". We solved that case differently (#TBD, authorize the existing application by scanning). The general problem stands.

**`co env` made the exposure legible.** Running `co env show --reveal` against a real `~/.co/keys.env` prints a tax file number, an invite code and several email addresses alongside the API keys, because a credentials file accumulates whatever anyone ever needed. The 1.8.4 review replaced name-based masking with blanket redaction for exactly this reason. Masking the display is the right patch; it is not the fix.

It is a 2.x item because it changes where credentials live, which is not something to do in a patch release, and because it has to work on three platforms before it can be the default anywhere.

## Shape

- A value can be stored either in the selected env file or in the keychain, and the file records which — the way `lark-cli` records `{"source": "keychain", "id": …}` rather than inventing a lookup.
- `co env set --secret NAME` puts it in the keychain; `co env set NAME` keeps today's behaviour. Existing files keep working untouched.
- `co env` shows the name, the source, and never the value unless `--reveal`, which is already true.
- `co auth google|microsoft|feishu` write their records to the keychain once it is available, since those are the values nobody types by hand.
[...]

Issue #17, opened 2025-10-04
Implement core Debugger infrastructure for interactive debugging


wu-changxing:
## Description

Implement the core Debugger infrastructure that enables interactive debugging at `@xray` breakpoints.

## Related Discussion

#16 - Design Discussion: Interactive Auto-Debug System for AI Agents

## Acceptance Criteria

- [ ] Create `Debugger` class in new `connectonion/debugger.py` file
- [ ] Implement `auto_debug(agent, step=False, interactive=True)` function
- [ ] Add `agent.debugger` attribute to Agent class
- [ ] Integrate debugger checks in `tool_executor.py` after xray table display
- [ ] Implement basic REPL menu with commands: continue, inspect, quit
- [ ] Add `should_pause()` method that checks for `@xray` decorator
- [ ] Write unit tests for Debugger class

### New File: `connectonion/debugger.py`

```python
class Debugger:
    def __init__(self, agent, step_mode=False, interactive=True):
        self.agent = agent
        self.step_mode = step_mode
        self.interactive = interactive
        self.continue_mode = False

    def should_pause(self, xray_enabled):
        """Determine if we should pause at this tool."""
        if not self.interactive:
            return False
        if self.continue_mode:
            return False
        if self.step_mode:
            return True
        return xray_enabled

    def pause_at_tool(self, tool_name, tool_args, result, agent):
        """Show interactive menu and handle user commands."""
        # Show menu, get user input, execute command
        pass

def auto_debug(agent, step=False, interactive=True):
    """Enable interactive debugging for agent."""
    agent.debugger = Debugger(agent, step_mode=step, interactive=interactive)
    return agent
```

### Modify `tool_executor.py` (after line 170)

```python
# Interactive debugging
if agent.debugger and agent.debugger.should_pause(xray_enabled):
    modified = agent.debugger.pause_at_tool(
        tool_name=tool_name,
        tool_args=tool_args,
        result=result,
        agent=agent
    )
    if modified is not None:
        result = modified
```

### Modify `agent.py` `__init__`

```python
# Add debugger attribute (default None)
self.debugger = None
```

## Testing

```python
def test_debugger_creation():
    agent = Agent(name="test", tools=[])
    debugger = auto_debug(agent)
    assert agent.debugger is not None
    assert agent.debugger.interactive == True
    assert agent.debugger.step_mode == False

def test_debugger_should_pause():
    agent = Agent(name="test", tools=[])
    auto_debug(agent)

[...]

PR #1539, opened 2026-09-14
Break the Lark auth deadlock: an honest link, a door for a console credential, and a secret that is not plain text (#1537)

### LINKED ISSUE #1537: co auth lark used the server's /page/launcher URL, which cannot complete the flow; lark-cli builds /page/cli instead
`co auth lark` cannot create an application on a data-residency tenant. The
launcher page discards the `user_code` and renders "Link expired" while the
device code is still alive on the server.

## Reproduced, with the contradiction visible

A fresh code, loaded exactly once, in a browser already logged into the tenant:

```
21:21:02  user_code:        4LLH-3FGX
21:21:02  uri_complete:     https://open.larksuite.com/page/launcher?user_code=4LLH-3FGX
21:21:02  expires_in 3600s — polling
```

Navigate to that exact URL:

```
go_to    https://open.larksuite.com/page/launcher?user_code=4LLH-3FGX
current  https://open.larksuite.com/page/launcher          ← user_code is gone
text     Link expired
```

And at the same second, the poller against
`accounts.larksuite.com/oauth/v1/app/registration` (`action=poll`) answers:

```
21:20:04 still pending (code is alive)
```

So the code has not expired. The page never reads it — the query string is
stripped on load, and "Link expired" is what that page shows when it has no
code.

Four codes were burned this way before the cause was visible, each one reported
by the owner as "expired" within a minute or two of being issued.

## The tenant is data-resident

The owner's Lark tenant serves from `kjp2hfo6uy94.jp.larksuite.com` (JP). The
device code is issued by the global `accounts.larksuite.com`, and the launcher
it hands back is on the global `open.larksuite.com`. The region-local
equivalent, `open.jp.larksuite.com/page/launcher`, answers `404 Not Found`.

One more data point that fits: in a browser **not** logged in, the same URL
redirects to `accounts.larksuite.com/accounts/page/login` with the `user_code`
preserved inside `redirect_uri`. It is only after a session exists that the code
is dropped — consistent with the launcher resolving the session's tenant and
then failing to carry the code into that tenant's context.

## Why this matters beyond one scan

`co auth lark` is the documented way to obtain Feishu/Lark credentials —
`co env set FEISHU_APP_SECRET` refuses and names this command. On a
data-residency tenant there is currently no working path through our own CLI:
the person is told to scan, scans, and is shown "Link expired" with a live code.

It also blocks #1462: the reconnect acceptance run needs a dedicated
application, and the runbook's recommended way to get one is this command.

## What to work out

1. Does `lark_oapi.register_app` accept a region-specific accounts domain, and
   does that tenant issue codes its own launcher can read? `domain` and
   `lark_domain` are already parameters; `co auth lark` sets `domain` to
   `https://accounts.larksuite.com` as of #1534.
2. If the SDK cannot reach a data-resident tenant, `co auth lark` has to say so
   rather than print a link that will say "Link expired". The failure is
   currently indistinguishable from the person being slow.
3. `--app-id` reuse goes through the same launcher, so it is likely to fail the
   same way. Worth confirming before it is offered as the workaround.

## Not a fix, but worth recording

`lark-cli` on the same machine holds working credentials for this tenant and
keeps its secret in the macOS keychain. Reading it out into a plaintext env
file is the downgrade #1497 exists to avoid, so it is not the answer here.

**Proposed target version:** 1.8.6
**Estimated release window:** with the 1.8.6 channel work

wu-changxing:
Closes #1537. Unblocks #1462.

Three commits that together break one deadlock:

1. `co auth lark` **could not** create an application on a data-residency tenant, and said nothing true about why.
2. `co env set LARK_APP_SECRET` refused the hand-typed value, and pointed back at `co auth lark`.
3. Opening that door means people now paste an app secret into a plain-text file — so `--secret` encrypts it.

---

# 1. The link does not last ten minutes, and "expired" does not mean expired

`co auth lark` printed a link; the page said "Link expired" within a minute; the
code was alive the whole time. Four codes were spent before the word on the
screen was doubted — because the command had said the link lasted ten minutes.

```python
expire_in = begin_res.get("expires_in", 600)   # 600 is the SDK's FALLBACK
```

The server returns **3600**. I quoted a default as a measurement, and an hour
went into investigating a timeout that never existed.

## What is actually happening

Measured on the owner's JP data-residency tenant, in a browser logged into it,
with a code issued seconds earlier and opened exactly once:

```
go_to    https://open.larksuite.com/page/launcher?user_code=4LLH-3FGX
current  https://open.larksuite.com/page/launcher        ← user_code is gone
text     Link expired
```

…while the poller against the registration endpoint answered, in the same
second:

```
21:20:04 still pending (code is alive)
```

The page never read the code. `user_code` is stripped on load, and "Link
expired" is what that page renders with no code. In a browser **not** logged in,
the same URL redirects to login with `user_code` preserved inside `redirect_uri`
— it is only once a tenant session exists that it is dropped.

## Why the fix is a sentence and not a domain

The satisfying fix would be a region-local accounts host, and
`lark_oapi.register_app` already takes `domain`. There is no such host:

| host | result |
|---|---|
| `accounts.larksuite.com` | ok — launcher on `open.larksuite.com` |
| `accounts.jp.larksuite.com` | 404, `text/html` |
| `accounts.sg.larksuite.com` | 404, `text/html` |
| `<tenant>.jp.larksuite.com` | 404, `text/html` |

And it cannot be pre-checked: a server-side GET of the launcher returns 200 with
the code still in the URL, because the verdict is rendered client-side after the
page resolves the browser's session.

The TTL now comes from the server every time, and when the field is absent the
command says *"The platform did not say how long this link is valid"* rather
than reaching for 600 again. The printed warning names `--app-id` as something
that **also** fails — I ran it first: same endpoint, same launcher, same "Link
expired". A remedy that cannot work sends the reader in a circle with the
confidence of an instruction.

---

# 2. A door for the one app credential a person can legitimately type

The refusal was right in general — an app secret should come from `co auth`,
because a hand-typed one has no checkable source. But with `co auth lark` unable
to serve this tenant, the two rules pointed at each other and there was no way
out of the loop.

```
co env set LARK_APP_SECRET <value> --from-console
```

The flag is narrow on purpose: it is accepted **only** on `FEISHU_`/`LARK_` app
credentials, and passing it on any other key is an error rather than a no-op —
so it cannot quietly become the way people set everything.

---

# 3. …and that value does not have to sit in plain text

## What lark-cli does, and why we cannot copy it

The reference implementation, `lark-cli`, generates an AES-256 master key, keeps
it in the OS keychain, and writes AES-GCM ciphertext beside it
(`internal/keychain/keychain_darwin.go`, MIT). It has to invent a master key
because it has nothing else to start from.

We cannot copy it. An agent under launchd, over ssh, or on a `co deploy --to`
box has **no login keychain at all** — the same wall the browser licence gate hit
over SSH (openonion/browser#137). A store that only opens when a human is sitting
at the machine is not a store an agent can use.

## We do not need a master key

`derive.py` already turns a secret into a tree of keys, so the encryption key is
*derived at a path* and thrown away. There is nothing to store, and therefore
nothing to put in a keychain. `slip13_path` already carries a rotation index
whose entire purpose is this, so `co env rotate` is "derive at index+1,
re-encrypt" — no key-wrapping scheme to design.

```
co env set LARK_APP_SECRET <value> --from-console --secret
co env get LARK_APP_SECRET
co env rotate LARK_APP_SECRET
```

`co env get` consults the encrypted store **last** — after the process
environment and the file — so adding it cannot change what any existing setup
resolves to.

## The root is agent.key, not recovery.txt

The first version of this rooted the tree at `.co/keys/recovery.txt`. That was
wrong, and it is the part of this PR most worth reading.

The phrase file is **optional**. `co keys` prints "missing" for it without
complaint, and writing the twelve words on paper and deleting the file is the
correct thing to do with a recovery phrase. Rooting there would have meant that
doing the right thing with your backup silently made every stored secret
unopenable — with no error pointing anywhere near the cause — and that a routine
`co env get` opened the most sensitive file on the machine on every lookup.

`.co/keys/agent.key` is 32 bytes, exactly what a SLIP-0010 master key takes, and
`address.load()` already opens it on every command.

**Nothing is given up by moving the root down one level**, because `agent.key` is
itself `derive_path(bip39_seed, slip13_path(ACCOUNT_URI))`. The twelve words
still reach every secret — through the agent key rather than around it:

```
twelve words ──BIP-39──▶ seed ──SLIP-0013──▶ agent.key ──SLIP-0013──▶ secret key
```

So a new machine still decrypts from the words alone. It just no longer *reads*
them to do ordinary work.

## What it does not claim

`agent.key` is a `0600` file next to the ciphertext. Anyone who can read it can
decrypt the whole store, and the module says so in its own docstring rather than
implying otherwise — but that person can already sign as the agent, so this is
not what they would attack. What it buys is narrower and real: an app secret is
no longer sitting in plaintext in a file people `cat` in a screen share, paste
into an issue, and read aloud on a call.

---

# Tests

| file | tests | red-first |
|---|---|---|
| `test_auth_says_what_expired_means.py` | 11 | all verified red against `main` |
| `test_env_set_accepts_a_console_credential.py` | 10 | all verified red against `main` |
| `test_secret_store.py` | 18 | the 4 that assert the new root verified red against the phrase-rooted version |

The four red-first secret-store tests are the ones that matter: `recovery.txt`
never being needed, the twelve words still reaching the secret, a different agent
key failing loudly rather than returning wrong bytes, and a truncated key being
refused instead of silently deriving *something*.

One existing test asserted `"--app-id" not in out` to mean "nothing was offered".
The new warning mentions `--app-id` in order to say it does not help, so a bare
substring can no longer tell an offer from a caution; that assertion was
tightened to the offer itself.

**Verified on real files, not only in tests:** `co env set --secret` → the
plaintext is absent from the file on disk → `co env get` returns it → `co env
rotate` moves it to index 1 → `co env get` still returns it. Then `recovery.txt`
was moved aside and `co env get` still returned the value, and the file was
restored byte-identical (sha256 unchanged).

Full unit suite: **9429 passed**, with only the four known pre-existing
cross-test-pollution failures, all of which pass in isolation and are untouched
by this branch.

**Proposed target version:** 1.8.5b9
**Estimated release window:** today, 2026-09-15


wu-changxing:
## CodeQL: one high-severity alert, examined and kept open

[Alert #17](https://github.com/openonion/connectonion/security/code-scanning/17) — `py/clear-text-logging-sensitive-data` on `env_commands.py`, the `print(get_secret(...))` in `co env get`.

**The data flow is real.** `co env get LARK_APP_SECRET` decrypts the value and writes it to stdout. CodeQL is not mis-reading the code.

**It is the command's contract, not an accident.** `co env get` is the substitution form — `export LARK_APP_SECRET=$(co env get LARK_APP_SECRET)` — so a redacted answer here would be a *wrong* answer. `co env show` is the browsing command, and that one redacts by default and warns about shared logs.

**Suppressing it would move the number, not the exposure.** The same function has three other `print()` calls, a few lines above, that reveal an OAuth token (`record.values[key]`), an inherited process value, and a file value such as `OPENAI_API_KEY` — identically, in clear text. They are unflagged only because `read_env_file` and `resolve_provider_credentials` are not modelled as taint sources. Rewriting the one flagged line to break the flow would leave every equivalent path untouched and the alert count at zero.

**What encryption was for.** #1497 is about a secret sitting at rest in a file people `cat` in a screen share, paste into an issue, and read aloud on a call. That is what the store removes. Retrieval by explicit name is inherently plaintext; there is no version of "get" that does not reveal.

The reasoning is now written at the call site (a90ee589) so the next reader meets it there rather than rediscovering it.

I attempted to dismiss the alert as *won't fix* with this rationale and was blocked by a permission guard on security-alert state — correctly, I think. **@wu-changxing it needs your hand**, or it can stay open as a documented, accepted flow.

---

Separately, [alert #16](https://github.com/openonion/connectonion/security/code-scanning/16) (`py/incomplete-url-substring-sanitization`, high) is already on `main` from the 9/14 PR: `assert "open.larksuite.com" in out` in `test_auth_lark_begins_on_lark.py:81`. Here `out` is captured stdout, not a URL being sanitized for a security decision — a genuine false positive, and not part of this branch.

Issue #1555, opened 2026-09-15
[1.8.6] Release plan: WhatsApp groups, Ollama, email CLI and calendar invitations


wu-changxing:
## Schedule update — 2026-09-18

Only what comes *after* this release changed; 1.8.6's own scope below is
untouched.

- The adapter plan #1463 moves from "approximately 1.8.8, tentative" to **1.9.1**
  — after the sender allowlist (#1479) rather than before it.
- #1499 moves from 1.8.9 to **1.9.0**.
- **1.8.8 and 1.8.9 are deliberately unassigned.** Wiki (1.8.7) is open-ended,
  so anything scheduled behind it would be a date nobody could keep.

Order: 1.8.6 → 1.8.7 Wiki → 1.8.8/1.8.9 unassigned → 1.9.0 → 1.9.1 adapters.
Recorded in `docs/roadmap.md`.

Still open from the Delivery list below: *"Reconcile existing release
labels/milestones with this decision."* The milestone this issue sits in is still
named `1.8.6 — WhatsApp, Discord and Telegram mailbox adapters`, which describes
a plan that has now been superseded twice. This connector cannot rename, create
or reassign milestones.

---

## Owner decision — 2026-09-15

1.8.6 is a bounded release for WhatsApp group access, Ollama, email CLI
improvements, and calendar invitation correctness. Wiki is scheduled for 1.8.7.
The previous multi-channel adapter plan (#1463) moves to approximately 1.8.8
(tentative — **superseded, now 1.9.1**). Calendly is excluded for now.

## Scope and acceptance

| Workstream | Existing work | Required outcome |
| --- | --- | --- |
| WhatsApp existing-group access | #1543 | Verify real device linking, inbound group messages, sender/group identity, mention-only triggering, replies and reconnect behavior. Keep the linked-device optional-extra scope and documented limitations explicit. |
| Ollama / custom local endpoints | #103, #1538 | Complete review and installed-package acceptance for local text, structured output and a tool-capable model round trip without cloud keys or silent fallback. Document model-specific limitations; Wiki and automatic note collection are not dependencies. |
| Gmail / Outlook CLI | #1521, #1522 | Complete useful date-window queries and machine-readable output. Gmail date filtering must compose with the existing JSON/pagination contract; preserve caps, account context and accurate completeness reporting. |
[...]

PR #1559, opened 2026-09-17
WhatsApp reads every message on protobuf 7, check stops calling an unscanned QR linked, and a live e2e suite

wu-changxing:
## Description

The first real-account run of **1.8.6a1** (17 September, a dedicated number linked as a companion device on the owner's Mac) found two defects that the unit suite could not, and this PR fixes both and adds the live suite that would have caught them.

### 1. Every inbound message was dropped on protobuf 7 (release blocker)

`_context_info` read `FieldDescriptor.label` on the first field of every message, outside its `try`. protobuf 7 removed `label` (6.x added `is_repeated`), and the extra's requirement `protobuf>=6.33.5` resolves to 7.x on a fresh install. Live log:

```
2026-09-17T02:08:04Z event not understood (AttributeError: 'google._upb._message.FieldDescriptor' object has no attribute 'label'); skipped
```

No message reached the inbox. The unit fake for the descriptor carried a `label` attribute, so the suite stayed green. Fixed by reading `is_repeated`, falling back to `label` only where `is_repeated` does not exist.

Verified live: with protobuf pinned to 6.33.6 as a stopgap on the test machine, the same released code received a direct message (`chat=…@lid`, `mentioned=true`), and `co whatsapp reply` delivered to the owner's phone (id `3EB0B0ECCD1AB285150783`, confirmed received on the handset).

### 2. `check` reported an unscanned QR as a linked device

`listen` creates `session.db` when it shows the first QR, before any phone scans it. The owner's first scan failed on the phone with a connection error and left exactly that file; `check` then said the account was ready. A link is the row whatsmeow writes to `whatsmeow_device` when the phone confirms, and `check` now requires it: `The WhatsApp session at … was never linked: a QR code was shown but no phone confirmed it.`

### 3. `tests/e2e/real_api/test_real_whatsapp.py`

Each number is scanned **once, ever**: a linked session is a file, and the suite takes its path from the environment and drives the real CLI as subprocesses (source checkout or installed wheel via `WHATSAPP_E2E_CO`). Setup and run commands are in the module docstring.

| Needs | Test |
|---|---|
| nothing | `check` refuses a session whose QR was never confirmed (connects to WhatsApp far enough to be shown a QR) |
| bot session | connects with no QR; restart reuses the link; `check` passes; a send to our own chat goes out and is not delivered back; Ctrl-C stops the listener (**strict xfail**, see below) |
| + driver session | DM arrives once and the reply reaches the sender; a message sent while the listener was down arrives exactly once |
| + group JID | only a message naming the number is `mentioned` |

## Found, not fixed here

- **Ctrl-C does not stop `co whatsapp listen`.** 40 s after SIGINT it was still running: the main thread is inside `client.connect()` in Go, so Python never runs the handler. SIGTERM stops it in ~2 s. Recorded as a strict xfail so the fix flips it. The docs say "Ctrl-C stops".
- **`reply` is not a quoted reply.** `WhatsApp.send()` accepts `reply_to` and does not use it, so the answer lands as a plain message.
- **`co whatsapp consume` does not gate on `mentioned`.** `mention_only` lives in `Channel`, used by `co ai` and the Host inbox; plain `consume` runs the command for every group message. The docs' "Groups: when the bot answers" section reads as if it applied.
- The phone's "connection error" on the first scan did not reproduce on a recorded second attempt (paired in 8 s); cause unknown.

## Related Issue

Part of #1555 (WhatsApp existing-group access: real device linking, inbound, replies). Does not close it: group mention and reconnect are still owed live.

- **Suggested labels:** bug, platform, tests
- **Proposed target version:** 1.8.6 (next preview; #1 blocks any WhatsApp claim)
- **Estimated release window:** before 1.8.6a2
- **Milestone:** 1.8.6
- **Why this release:** without #1 the headline 1.8.6a1 feature receives nothing on a default install.

## How this was written

- [ ] I wrote this myself
- [ ] Mostly AI-generated

- **Least confident:** the driver and group e2e tests have not been run yet (no second linked number); `@lid` chats mean the DM test asserts the chat's server, not a phone number.
- **Not verified:** the full live suite end to end. Verified live: the protobuf failure and its stopgap, DM receive and reply on the owner's handset, the unpaired-`check` e2e test (failed on the released code at `assert 0 == 3`, passes with the fix).

## Type of Change
- [x] Test coverage

## Dev Blog (required)

Not included; the live acceptance record for #1555 will carry the narrative once the group run is done. Maintainer's call on `no-blog`.

## Testing

```
tests/unit/test_inbox_whatsapp.py                 27 passed   (both new protobuf tests failed first with the live AttributeError)
tests/unit tests/e2e/cli (offline, -n 6)          10094 passed, 18 skipped, 0 failed
test_real_whatsapp.py, no sessions configured     8 skipped; unpaired-check test runs live and passes
```

https://claude.ai/code/session_01U2EcJHGjPwvXZqNhweg2mV


wu-changxing:
Live observations after opening, same run (listener pid started 12:09:50 AEST, `co whatsapp consume` with a test consumer that answers groups only when `mentioned`):

- **Automatic reconnect works.** WhatsApp closed the socket and the same process re-authenticated 0.7 s later, no restart and no re-pair:
  ```
  12:32:15.357 [whatsmeow.Client.Socket ERROR] - Error reading from websocket: failed to get reader: failed to read frame header: EOF
  12:32:16.084 [whatsmeow.Client INFO] - Successfully authenticated
  ```
  The inbox `log` records only `connected as 61410724095` at 02:32:16Z; the disconnect is not written there (Feishu logs `reconnecting`), so the inbox log alone cannot tell a reconnect from a restart.
- **Bot added to a human-created group.** The first group event (`120363410170505910@g.us`) arrived with a redelivered duplicate that was dropped (`duplicate … dropped`), and the consumer stayed silent as intended (`mentioned=false`). Its `text` was empty; whether that was the group-creation notice or a non-text message is not yet known. Unlike Feishu (`[image]`), a non-text WhatsApp message reaches consumers as an empty string.

Group plain-text vs @-mention and the third participant are still to run.

PR #1576, opened 2026-09-17
check also refuses a session WhatsApp logged out, not only one never scanned

wu-changxing:
## Description

Follow-up to #1559, from the same real-account run of 1.8.6a1. #1559 made `check` require a row in `whatsmeow_device` instead of trusting that `session.db` exists, and described the failing state as "a QR that was never confirmed". The run then produced the second way to reach that state.

At 13:49 AEST the listener printed:

```
13:49:21.415 [whatsmeow.Client INFO] - Got device removed stream error, sending LoggedOut event and deleting session
13:49:21.564 [whatsmeow.Client.Socket ERROR] - Error reading from websocket: failed to get reader: failed to read frame header: EOF
```

whatsmeow deleted the device row and **kept the 850 KB session file**, and the installed 1.8.6a1 `check` printed `✓ whatsapp reachable … 0 unread` and exited 0 for an account with no linked device. The device-row check from #1559 already refuses this; its message named only the unscanned-QR cause, so it now names both:

> The WhatsApp session at … has no linked device: either its QR code was never confirmed on a phone, or the device was removed (from the phone's Linked devices, or by WhatsApp). Next: co whatsapp listen …

Verified read-only against the real logged-out session file.

## Found, not fixed here

- **A logout is invisible in the inbox log.** Its last line stayed `connected as …` from 02:32Z; the "device removed" and the listener's exit appear only on the terminal. Anyone reading `log` (or `check` before this PR) sees a healthy account. `listen` should record the LoggedOut event and exit non-zero with the relink command.
- Why the device was removed (the phone's Linked devices, or WhatsApp acting on the number) is not yet known; asked the owner.

## Related Issue

Part of #1555 (WhatsApp real-device linking and reconnect behaviour). Follow-up to #1559.

- **Suggested labels:** bug, platform
- **Proposed target version:** 1.8.6
- **Estimated release window:** with the next 1.8.6 preview
- **Milestone:** 1.8.6
- **Why this release:** `check` is the command the docs send people to after linking; it must not call a removed device ready.

## How this was written

- [ ] I wrote this myself
- [ ] Mostly AI-generated

- **Least confident:** nothing in the change; the open question is operational (why WhatsApp removed the device).
- **Not verified:** a relink after this logout.

## Type of Change

## Dev Blog (required)

Covered by the Design Journal entry added to #1559; maintainer's call on `no-blog`.

## Testing

```
tests/unit/test_inbox_whatsapp.py                 47 passed (new: a session WhatsApp logged out is not a linked device)
tests/unit tests/e2e/cli (offline, -n 6)          10132 passed, 18 skipped, 0 failed
```

https://claude.ai/code/session_01U2EcJHGjPwvXZqNhweg2mV


wu-changxing:
Already on main in another form: `WhatsApp.linked()` now says "No linked device in <session> — either the QR was never scanned, or this device was logged out from the phone", with the reason in a comment (whatsmeow clears device, identity_keys and sessions together on logout, so the two states leave byte-identical evidence), and `test_a_session_whose_qr_was_never_scanned_is_not_a_linked_device` asserts both stories. A logged-out session is the same zero-device state that test builds, so nothing here is left to port. Closing during the 1.8.8 feature sweep.

PR #1598, opened 2026-09-18
feat(browser): co browser requests / request — see what a page actually sent

### LINKED ISSUE #1597: [1.8.8] co browser: see what a page actually sent and received, so a skill can analyse it
## Outcome

`co browser` can answer: **that action I just performed — what requests did it send, and what came back?**

This is a base tool, not a product feature. Its consumer is a skill: today
`browser-workflow-skill-builder` analyses a page through the DOM
(`save_page_context` → HTML/CSS → selectors), and has no way to see the network
layer at all. This adds the missing half of the same picture.

## What exists today, and why it is not this

- **`save_page_context`** saves HTML, CSS and interactive elements. DOM only.
- **The egress gateway** (DD-056, #991/#1297/#1036) *is* a loopback proxy every
  Chromium connection passes through — but its job is to fail-closed **authorize**
  an outbound connection, not to record what went over it. Different purpose, and
  that chain is still blocked on two-machine acceptance.
- **Nothing captures traffic.** `grep` over the repo finds no `page.on("response")`,
  no `route()`, no CDP Network domain, no HAR. Zero hits.

## Why not a MITM proxy

The original framing was a man-in-the-middle proxy. For this purpose the browser's
own instrumentation is strictly better and much cheaper:

| | CDP / Playwright events | MITM proxy |
|---|---|---|
| Setup | none | CA certificate installed into the profile |
| TLS | already inside it | must terminate and re-originate |
| Bodies | decrypted, decompressed, as the page saw them | same bytes, more plumbing |
| Attribution | knows the page, frame and initiator | IP and port only |

A proxy earns its place when the traffic is **not** from our browser (a native app,
another process). That is a different issue. This one stays in the browser.

## Surface

```bash
co browser requests                          # what this page has sent
co browser requests --url-contains=/api/     # filtered
co browser requests --method=POST --type=xhr
co browser requests -n 20 --json             # for a skill to read
co browser requests --clear                  # isolate the next action
co browser request 7                         # one request, in full
co browser request 7 --raw                   # with header and cookie values
```

`requests` prints newest last, one per line: index, method, status, type, size,
duration, URL. `request <n>` prints request headers, request body, response
headers and response body.

## What is captured

Metadata for every request — that is cheap and the list is the index.

Bodies only for `xhr` / `fetch` / `document` responses under a size cap, and only
for textual content types. Images, fonts, media and video segments are listed but
their bodies are never read: they are the bulk of the traffic and none of it is
what anybody is looking for. A body over the cap is written to a file and the
record names the path.

Bodies must be read while the response is alive — Playwright cannot return one
after a navigation — so this is eager, which is exactly why the type filter above
matters.

## Header values are redacted by default

This is the one deliberate design decision and it is worth arguing about.

The consumer of this data is a skill, and a skill feeds an LLM. `Cookie`,
`Authorization`, `Set-Cookie` and friends are precisely the values that must not
reach a prompt by accident — the same reason the inbox gates its payloads behind
`--raw` (DD-063).

But redacting them entirely would defeat the purpose: knowing that an endpoint
requires an `X-Sign` header, and what shape it has, is most of the answer.

So the default shows the **name and the shape**, not the value:

```
x-sign: <32 hex>
cookie: <4 pairs, 312 bytes>
authorization: Bearer <172 chars>
```

`--raw` prints the values. A skill can learn the contract without the secret
entering the context window; a human reversing a signature runs `--raw` once.

## Not in scope

- Generating a CLI or a script from what was captured. That is the larger idea
  this serves and it needs this first.
- Replaying a captured request. Reproducing a signed request means reimplementing
  its signing from the page's JS, not resending bytes — a different job.
- Capturing traffic from anything that is not this browser.
- WebSocket frames. Worth having, probably the obvious follow-up, but the
  request/response case is the whole of what a skill needs to start.

## Notes for whoever builds on it

Two things that bite people who reverse a request correctly and still get rejected,
recorded here so the skill that eventually generates callers knows to expect them:

- **TLS fingerprinting.** A server may check JA3/JA4 and the HTTP/2 SETTINGS frame
  order. A reimplemented request can be byte-identical at the HTTP layer and still
  be refused because the handshake does not look like Chrome. `curl_cffi` /
  curl-impersonate exist for this, or the call goes back through the browser.
- **Some secrets are not on the client.** A server-issued token has no algorithm to
  reimplement. Acquire it with the browser, then use it from wherever.

Neither is a reason not to build this. Both argue that anything generated from this
data should be able to fall back to the browser per call rather than being all or
nothing.

## Acceptance

- [ ] An `xhr` POST made by a real page appears in `co browser requests` with its
      status, size and duration.
- [ ] `co browser request <n>` shows the request body the page sent and the
      response body it received.
- [ ] Header values are shaped, not printed, without `--raw`; printed with it.
- [ ] Image, font and media responses are listed and their bodies are never read.
- [ ] `--clear` then one action shows only that action's requests.
- [ ] A body over the cap lands in a file and the record names it.
- [ ] `--json` output is what a skill reads; it round-trips.
- [ ] Capture costs nothing measurable on a page that makes no XHR.

wu-changxing:
## Description

`co browser` could read a page's DOM and could not see a single byte it sent. The skill that analyses pages (`browser-workflow-skill-builder`) therefore works only from HTML and selectors — half the picture. This adds the other half as a **base tool for a skill to read**, not a product feature.

```bash
co browser requests                            # the index
co browser requests --url-contains=/api/ --method=POST
co browser requests --as-json=true             # for a skill
co browser requests --clear=true               # then one action, to isolate it
co browser request 7                           # one request, both directions
co browser request 7 --raw=true
```

Against a real page:

```
1  GET   200  image     —     28ms  http://127.0.0.1:8731/logo.png
2  GET   200  document  435B  14ms  http://127.0.0.1:8731/
3  POST  200  fetch     247B  19ms  http://127.0.0.1:8731/api/echo
#  method status kind   size  took  url
```

```
#3  POST http://127.0.0.1:8731/api/echo
--- request headers ---
  authorization: Bearer <48 chars>
  content-type: application/json
  origin: http://127.0.0.1:8731
  user-agent: Mozilla/5.0 (X11; Linux x86_64) … HeadlessChrome/149.0.7827.55 …
  x-sign: <32 hex>
--- request body ---
{"q":"hello","ts":1789769729255}
--- response headers ---
  set-cookie: <2 pairs, 28 bytes>
--- response body ---
{"echoed": {"q": "hello", …}, "rows": [0, 1, 2, …]}
```

**Not a MITM proxy.** The browser's own instrumentation needs no certificate, is already inside TLS, sees bodies decompressed as the page saw them, and knows which frame made the call. A proxy earns its place for traffic that is *not* ours — a different tool, not this one. (The existing egress gateway, DD-056, *is* a proxy every connection passes through, but its job is to fail-closed authorize an outbound connection, not to record it.)

## Related Issue

Closes #1597.

- **Suggested labels:** enhancement, browser
- **Proposed target version:** 1.8.8
- **Estimated release window:** next stable
- **Milestone:** [maintainer assigns the confirmed release milestone]
- **Why this release:** 1.8.8 was left unassigned two days ago on the principle that *work gets a number when somebody is about to do it*; somebody now is. Additive — two new read-only verbs, one new module, and one line in `_ensure_page`. Nothing existing changes behaviour. Rollback is reverting one commit. Merge timing is the maintainer's call: landing on `main` now means it physically rides whichever preview is cut next.

## How this was written

- [ ] I wrote this myself
- [ ] Mostly AI-generated

The `_looks_opaque` heuristic that decides whether an unknown header's value gets shaped. A name list alone is not enough — every site invents its own signature header (`x-bogus`, `x-gorgon`, `x-sign`), and printing the one header an operator most wants shaped would be the worst possible default. So a long unbroken value in a header whose *name* matches `sign|token|secret|auth|sess|key|nonce|bogus|gorgon` is shaped too. That regex is a guess calibrated on the names I know. It will miss some and, less likely, shape something harmless. `--raw` is one flag away in both directions, and no correctness depends on it.

Second: the ring buffer is 500 records per tab with no configuration. A long-polling page evicts its own earlier requests within minutes. I think that is the right default (the newest are what anybody wants) but it is a number I chose rather than measured.

**What did you not verify — which paths did you never actually run?**

I never ran this against a site with real anti-bot machinery, obfuscated signing, or HTTP/2 — the acceptance was a local HTTP/1.1 server I wrote, driven by the real headless Chromium. So the capture path is proven; how legible the output is on a page that is actively hostile is not.

I did not test a WebSocket-heavy page. Frames are out of scope and simply do not appear, which means a page whose real traffic is one WebSocket will look empty and the log will not say why. That is a known gap, not a surprise, but it is the most likely way somebody's first impression of this is "it doesn't work".

Not tested on Windows or macOS. The only platform-sensitive part is the body file path, which uses `Path.home()` like `save_page_context` beside it.

A body that should have been captured is not, because the content-type did not match `_TEXTUAL` — an API returning JSON under an unusual media type shows a size and an empty body. Cheap to diagnose (`request <n>` shows the content-type that was rejected) and cheap to fix.

The expensive failure would be capture slowing down or destabilising ordinary browsing. That is why bodies are read only for three resource types with textual content, why `attach` is a capability check that leaves an odd page alone rather than raising, and why every handler stores its reason on the record instead of propagating — a handler that raises inside Playwright is swallowed and the record is lost silently, which is the one outcome worse than no log.

## Type of Change

## Dev Blog (required)

Blog file in this PR:

> `docs/blog/2026-09-18-the-fake-agreed-with-me.md`

## Changes Made

- **`connectonion/useful_tools/browser_tools/_async_network.py`** (new, 370 lines) — the log: capture policy, header shaping, filtering, rendering. Sits beside `_async_element_finder`, `_async_scroll`, `_async_humanize` by the same convention.
- **`connectonion/useful_tools/browser_tools/_async_browser.py`** — four touches: the import, `self._network` in `__init__`, `attach()` in `_ensure_page`, `forget()` in `_release_tab`, plus the two verbs. `_install_sync_methods()` mirrors public async-core methods onto the facade automatically, so `co browser help` and the daemon pick them up with no registration.
- **`docs/co-browser.md`** — a "Seeing what the page sent" section under Command Reference.
- **`tests/unit/test_browser_network_log.py`** (new) — 29 tests.
- **`tests/unit/test_sync_browser_facade.py`** — two names added to `ASYNC_ONLY`. That list's own comment sets the bar: *"a NEW verb the oracle predates"*, which these are — the legacy sync oracle sees no traffic at all.

## Testing

```
$ python -m pytest tests/unit/test_browser_network_log.py -q -p no:randomly
29 passed in 0.90s

$ python -m pytest tests/unit -q -p no:randomly -k "(browser or doc or facade) and not server_commands"
1 failed, 1326 passed, 5 skipped, 8462 deselected in 29.12s
```

The one failure is `test_remote_browser_private_runtime.py::test_private_target_is_stable_short_and_separate_from_local`, confirmed failing on a stashed clean tree before any of this existed.

### The offline suite was not enough, again

Worth reading, because it is the second time in three PRs.

29 tests passed. Then the first real page printed a duration of `-1789769685953ms`. Playwright's `startTime` is an absolute epoch; every other field in `ResourceTiming`, `responseEnd` included, is an offset *from* it. `responseEnd` was already the duration.

The test that should have caught it:

```python
self.timing = {"startTime": 0, "responseEnd": 143}
assert record["duration_ms"] == 143
```

`startTime: 0` is not a value a page returns — it is 1 January 1970. I picked it because zero made my arithmetic come out right, and then asserted the arithmetic. The fake and the code were written in the same hour from the same misreading and agreed with each other perfectly.

Fixed in code, and the fake now carries an observed epoch (`1789769685953.0`) so it would fail the old implementation. The blog post is about this.

### Real-browser acceptance

Headless Chromium 149 against a local page that posts a signed `fetch` and loads a PNG:

- image listed, body never read (`—` in the size column) ✅
- document and fetch bodies captured ✅
- durations real after the fix (28ms / 14ms / 19ms) ✅
- `x-sign` → `<32 hex>`, `authorization` → `Bearer <48 chars>`, `set-cookie` → `<2 pairs, 28 bytes>`; `SUPERSECRETVALUE` absent ✅
- `--raw=true` prints all three values ✅
- `--clear=true` drops 3 and the next action's requests start at index 4 — never reused ✅

One caught regression: `attach` originally called `page.on(...)` unguarded, which broke 21 existing tests that pass fake page objects. It is now a capability check (`callable(getattr(page, "on", None))`) rather than a caught `AttributeError`, and re-verified against the real browser afterwards.

## Scope

Six files, one concern.

| File | Why |
|---|---|
| `_async_network.py` | The log. All new logic is here. |
| `_async_browser.py` | Where a page is created and released, and where verbs live. |
| `docs/co-browser.md` | The verb reference. |
| `tests/unit/test_browser_network_log.py` | The tests. |
| `tests/unit/test_sync_browser_facade.py` | Its allowlist gates new async-only verbs by design. |
| `docs/blog/…` | blog-gate. |

Deliberately not touched: `useful_skills/browser-workflow-skill-builder/SKILL.md`. Teaching the skill to *use* this is the next step and wants its own change, once somebody has driven it against a real site.

## Example Usage

```bash
co browser go_to https://app.example.com/orders
co browser requests --clear=true
co browser click_element_by_selector "#next-page"
co browser requests --kind=fetch
#  →  12  POST  200  fetch  4.1K  180ms  https://app.example.com/api/orders/search

co browser request 12
#  →  the search payload it posted, and the JSON that came back
```

## Checklist

- [x] This PR ships its dev-blog post under `docs/blog/`
- [x] If this targets a maintenance-branch patch … — N/A, mainline
- [x] Every piece of evidence #1597's acceptance list requires is above

## Additional Notes

**What this does not do, and why the next thing needs it.** It does not generate a CLI from what it captured, and it does not replay anything. Reproducing a signed request means reimplementing the signing from the page's JS — reading the algorithm, not resending bytes — and that needs this first, to see what the algorithm produced.

Two things recorded in #1597 for whoever builds that, because they bite people who reverse a request *correctly* and still get refused: a server may fingerprint the TLS handshake (JA3/JA4, HTTP/2 SETTINGS order), so a byte-identical request from Python is still not Chrome; and a server-issued token has no algorithm to reimplement at all. Both argue that anything generated from this data should fall back to the browser per call rather than being all-or-nothing.

https://claude.ai/code/session_01WaAFWfbdhFGiYo64Hvn1gn

---
_Generated by [Claude Code](https://claude.ai/code/session_01WaAFWfbdhFGiYo64Hvn1gn)_


wu-changxing:
CI is green on `05c4b357` — all 22 jobs, including `test (3.10)` through `(3.13)`, `lockfile`, and all twelve `remote-browser-egress` matrix jobs.

One update to a limitation I stated in the description. I wrote *"Not tested on Windows or macOS"*, and that was true of my own run. It is now partly untrue in this PR's favour: `macos-e2e` and `windows-e2e` both run a **real `co browser` end-to-end** — daemon spawn, Chrome, navigation, and on Windows also three concurrent tabs and a daemon restart — and all of that now happens with the capture listeners attached to every page, because `attach()` sits in `_ensure_page`. Those passed.

So what is proven on all three platforms is that capture does not disturb the browser. What is still only proven on Linux is that it *captures* — the assertions about shaped headers, skipped image bodies and the body file under `~/.co/browser_network/` came from my local run against a real page. The file path is the one genuinely platform-sensitive line, and it uses `Path.home()` exactly like `save_page_context` beside it.

Worth recording because it is the reverse of the usual direction: the honest caveat got smaller after CI ran, not bigger.

---

wu-changxing:
Superseded by #1663, which carries this PR's commit unchanged (rebased onto current main, 174 commits ahead) and builds on it: the verbs move under `co browser network` to match vercel-labs/agent-browser, HAR recording and `cookies` are added, and flags drop the `=true` form. Leaving this open for the owner to close.

wu-changxing:
Superseded by #1663 (merged): `co browser -t TAB network requests | request | clear | har start | har stop` and `cookies` on the tab names, which covers this PR's `requests` / `request` and adds HAR recording. The same patch also sat on release/1.8.8b2 (#1640, closed); nothing of it is left behind.

PR #1608, opened 2026-09-19
docs: 1.8.6 is the stable channel

wu-changxing:
`docs/releases.md` still described **1.8.5** as the current release and **1.8.6a1** as the current preview. The preview text it advertised said WhatsApp "has not been accepted against a real account yet" — which stopped being true today, when that acceptance was exactly the gate 1.8.6 stable closed.

## Changes

- **Current release** → 1.8.6, with what it actually is: twelve verbs, the mention rule (an `@`, *or a reply quoting something the bot said* — the case with no `@` in it anywhere), `edit` and `delete`, Markdown arriving as formatting, local models that fail loudly rather than quietly becoming a paid model.
- **Current preview** → none. 1.8.6a9 was the last one and it became 1.8.6.
- Links the [release-candidate acceptance record](docs/acceptance/1.8.6/release-candidate-2026-09-19.md) rather than summarising it, and says the thing worth knowing about it: it was run against the package installed from PyPI, not a checkout.

The collapsed history block keeps its own label — its content is 1.8.5b11, so calling it "the preview line that became 1.8.6" would have been wrong.

- **Proposed target version:** 1.8.6
- **Estimated release window:** on merge
- **Milestone:** 1.8.6

Docs-only; no code or version metadata touched.


wu-changxing:
Superseded: stable moved on to 1.8.7 and the channel docs were updated with it. Closing rather than merging a page that would say 1.8.6 is current.

